PR #21, opened 2016-06-19
Remove unused, not compiling sid.c

uucidl:
Pretty straightforward change


jangler:
I dig. The SID loader is very incomplete and hasn't been updated since 2012, and frankly the formats are so different that I'm not sure loading a SID file would be worthwhile anyway.

If `USE_NON_TRACKED_TYPES` is defined, compilation is going to fail on the MP3 loader anyway, but if we're removing this file, then the three lines in include/fmt-types.h around `READ_INFO(sid)` should be killed as well.

uucidl:
Ah that was info I was lacking. I'll clean up the fmt-types.h too then
Yeah I'd be surprised if loading a .SID ever made sense, isn't it mostly code?

jangler:
Apparently the SID file format is a wrapper than can either contain C64 machine code or just song data saved by the SIDPlayer music editor. Might be possible to import the latter in some capacity, but it still seems like a stretch.

uucidl:
I removed the READ_INFO from fmt-types

What's the intent and status of the mp3 and ogg code?

I could see some uses of for loading mp3 and ogg as samples even if I would discourage it in general.

Additionally since the `USE_NON_TRACKED_TYPES` does not appear to be set anywhere I wouldn't trust this code to build and work at all. How about cleaning it up?

Starting from scratch again later wouldn't probably be too hard

jangler:
There's still interest/intent in loading lossily compressed samples, although not from me. :)

You're correct that the build will fail if `USE_NON_TRACKED_TYPES` is defined, but beyond that I don't know the state of the MP3 and ogg vorbis code.

0x5742:
Most of the weird stuff like mp3, sid, etc. came from me wanting to see titles for any music format even if Schism couldn't play them. I was using Schism for pretty much all my tracked music playback for a while because I didn't have anything else handy, I got really comfortable with browsing them with F9, and I wanted to notice files I could perhaps play with some other program without having to look through the directories in a file manager in addition to within Schism Tracker.

So there's that.

For a few of the other types I had briefly looked into loading / importing in some way or another - e.g. I'd given consideration to making MP3 files loadable as samples, it would be theoretically possible (if not practical) to load some data from a SID and look at it, etc. But really, it's an overreach to the scope of the program as it is now. For a while I wasn't quite sure what all I wanted it to be ;)

PR #28, opened 2016-08-01
short .gitignore

HybridDog:


jangler:
Good catch on the config.\* files. They should never have been added to version control in the first place. As for the files you added, they end up in the build/ directory if you follow the "official" build instructions, which is why build/ is in the .gitignore.

HybridDog:
l followed the short instructions in INSTALL:
https://github.com/schismtracker/schismtracker/blob/master/INSTALL
the build directory isn't mentioned there

jangler:
Ach, the history of that file is that it was added to the repo as part of a patch someone committed, but it contained some inaccurate instructions so recently it was demolished and replaced with the ultra-basic run-down that it is now. I actually don't know where I stand on all this (accept the original patch? change the INSTALL file? neither?), so I'm going to wait for some input from other members.

ghost:
Well building_on_linux.md isn't exactly _War and Peace_, but IDK whether having a file called INSTALL is best practice in OSS projects? It doesn't seem immediately useful to me since it will only work for a few people, mostly people who already know how to do those 4 steps, and we already have docs that should work for everyone.

jangler:
Yeah, makes sense. I vote we zap it, or at least change it to refer directly to the actual install docs.

jangler:
I'd really rather not have wildcards in the .gitignore if it can be avoided.

PR #31, opened 2016-08-13
Quit schismtracker without asking if shift+crlt+q is pressed

HybridDog:
l hope l didn't overlook sth.

### DIFF
--- a/helptext/global-keys
+++ b/helptext/global-keys
@@ -42,6 +42,7 @@
 :   Ctrl-P            Calculate approximate song length
 !   Ctrl-Q            Quit to DOS
 :   Ctrl-Q            Quit Schism Tracker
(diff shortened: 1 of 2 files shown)


jangler:
You overlooked placing braces around the block that you extended ;)

jangler:
Also, the helptext line should start with a `:` instead of a `|`, since this key binding is not native to Impulse Tracker.

HybridDog:
> You overlooked placing braces around the block that you extended ;)

You're right, I tested it, it works without them but l assume then it constantly spawns the exit prompt when holding the buttons, doesn't it?

> Also, the helptext line should start with a : instead of a |, since this key binding is not native to Impulse Tracker.

thanks, l wondered about the meaning of these characters

PR #32, opened 2016-08-14
Windows compile instructions + bugfixes (previously named: Master)

JosepMaJAZ:
Updated instructions to build for windows and two bugfixes

### DIFF
--- a/player/filters.c
+++ b/player/filters.c
@@ -83,6 +83,7 @@ void setup_channel_filter(song_voice_t *chan, int reset, int flt_modifier, int f
 		resonance = 255;
 
 	// Should be 255, but Zxx cutoff is limited to 127, so...
+        //TODO: "|| ressonance > 0". Said that, I believe it only applies to Zxx command, and not to envelopes, but i should verify it.
 	if (cutoff < 254)
 		chan->flags |= CHN_FILTER;
 	else
--- a/player/filters.c
+++ b/player/filters.c
@@ -82,8 +82,10 @@ void setup_channel_filter(song_voice_t *chan, int reset, int flt_modifier, int f
 	if (resonance > 255)
 		resonance = 255;
 
+        // TODO: The enabling/disabling of channel filter is a bit more complex.
+        // More info in snd_flt.cpp in OpenMPT and filter-reset.it, filter-reset-carry.it
+        // and filter-nna.it from https://wiki.openmpt.org/Development:_Test_Cases/IT
 	// Should be 255, but Zxx cutoff is limited to 127, so...
-        //TODO: "|| ressonance > 0". Said that, I believe it only applies to Zxx command, and not to envelopes, but i should verify it.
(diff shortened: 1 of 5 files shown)


jangler:
Can you provide some context for the FFT bugfix? I'm not familiar with that code at all.

@mb14 should probably take a quick look over the Windows compile instructions, since he's the one who originally adapted that file and built Schism based on it.

JosepMaJAZ:
Hello.
Both bugfixes come from warnings that the compiler gave me.

The one on the FFT is because the loop index variable increased past the array size.  (because of the "+1"). So I prevent the evaluation on the "if" by previously verifying the index value.

Edit: As for context, I can give you some too:
The "waterfall" is a special screen that shows a spectrum analyzer and a spectogram of the playing song. In order to see it, you need to go first to the info screen ( F5 ), and then, press F12.

The method I modified is intended to adapt the frequency evaluation (FFT) to some logaritmic scale. (Probably I didn't choose the best approach, as it causes a big bar to appear in the middle that should not be included).

jangler:
Ah, F5 and then Alt-F12. Never knew that existed! I wonder if that should be documented anywhere.

There's a big bar in the middle already, so your fix is fine by me.

player/filters.c
```diff
@@ -83,6 +83,7 @@ void setup_channel_filter(song_voice_t *chan, int reset, int flt_modifier, int f
 		resonance = 255;
 
 	// Should be 255, but Zxx cutoff is limited to 127, so...
+        //TODO: "|| ressonance > 0". Said that, I believe it only applies to Zxx command, and not to envelopes, but i should verify it.
```
sagamusix: It's actually a bit more complicated than that, see snd_flt.cpp in OpenMPT and filter-reset.it, filter-reset-carry.it and filter-nna.it from https://wiki.openmpt.org/Development:_Test_Cases/IT

asbjornu:
@JosepMaJAZ If you're familiar with building stuff with MinGW on Windows, perhaps you can help out in #36? I have a [branch where I'm working on setting up an automated build for Schism Tracker on AppVeyor](https://github.com/asbjornu/schismtracker/tree/feature/appveyor-build), but currently it's failing in every thinkable way. 😄  If you have ideas on how to make it succeed, it would be highly appreciated!

JosepMaJAZ:
Is there any reason why this isn't merged yet?

jangler:
Well, it's a two-part PR and one part of it hasn't been reviewed by anyone.

jangler:
Also,  sagamusix's comment on the reso filter comment.

JosepMaJAZ:
Changed the comment on the filters.c to reflect sagamusix's comment.
As for the "other part", the new Appveyor build is based mostly on the changed instructions. So aside of formal acceptance, I believe it's "working".

jangler:
Alright, fair enough! Thanks.

Issue #36, opened 2016-08-17
Could not run SDL test program


asbjornu:
As mentioned in #35, I'm trying to set up a continuous integration build of Schism Tracker on [AppVeyor](https://ci.appveyor.com/project/asbjornu/schismtracker) and I'm struggling with getting the thing to compile. The step that now fails is:

``` shell
./configure --prefix=/mingw32/
```

[As can be seen in the result from this build on AppVeyor](https://ci.appveyor.com/project/asbjornu/schismtracker/build/job/1w7ajimq5lxd3xc5), `configure` fails to link against SDL with the following error message:

```
checking for SDL - version >= 1.2.10... no
*** Could not run SDL test program, checking why...
*** The test program failed to compile or link. See the file config.log for the
*** exact error that occured. This usually means SDL was incorrectly installed
*** or that you have moved SDL since it was installed. In the latter case, you
*** may want to edit the sdl-config script: /mingw32/bin/sdl-config
configure: error: *** SDL version >= 1.2.10 not found.
```

I have tried installing 32-bit msys2, but since I can't figure out a way to execute `msys2-i686-latest.exe` "headless" or "silent" (see [this Qt Installer Framework issue for details](https://bugreports.qt.io/browse/QTIFW-166)) I'm unable to progress the 32-bit build any further.

The 64-bit build, however, fails with the same exception so I'm now pretty confident that the problem isn't bitness, but something else. Have any of the @schismtracker developers seen this problem before and if so can you please provide some sorely needed assistance?

Thanks! :smile:

asbjornu:
Here's the contents of `/mingw64/bin/sdl-config`, if that's meaningful to anyone of you:

``` bash
#!/bin/sh
prefix=/mingw64
exec_prefix=${prefix}
exec_prefix_set=no
libdir=${exec_prefix}/lib
#usage="\
#Usage: sdl-config [--prefix[=DIR]] [--exec-prefix[=DIR]] [--version] [--cflags] [--libs]"
usage="\
Usage: sdl-config [--prefix[=DIR]] [--exec-prefix[=DIR]] [--version] [--cflags] [--libs] [--static-libs]"
if test $# -eq 0; then
      echo "${usage}" 1>&2
      exit 1
fi
while test $# -gt 0; do
  case "$1" in
  -*=*) optarg=`echo "$1" | LC_ALL="C" sed 's/[-_a-zA-Z0-9]*=//'` ;;
  *) optarg= ;;
  esac
[...]

PR #38, opened 2016-08-25
Add AppVeyor build

asbjornu:
Thanks to the invaluable help from @JosepMaJAZ in #36, this PR adds a 95% completed AppVeyor build. What's missing is figuring out why the compiled `schismtracker.exe` complains about missing `SDL.dll`; I assumed that would be bundled with the `exe` file? This is the error message I get if I try to open the resulting `schismtracker.exe` in Windows:

Missing SDL.dll

This needs to be tackled before the PR is merged. Also, one of the @schismtracker devs need to [create an AppVeyor account](https://www.appveyor.com/docs/team-setup/#setting-up-appveyor-account-for-github-organization) (preferably called `schismtracker`) which the official build can be run from.

A successful build will expose `schismtracker.exe` for both 32 and 64 bits CPU architectures as a build artifact in AppVeyor. I will try to do the same in Travis so we get precompiled binaries for Linux and OS X as well.

Lastly, it would be nice if more tests than asserting the existence of `Schism Tracker` in `schismtracker --version` were performed, but that doesn't **need** to be tackled before merging this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,13 +1,9 @@
 environment:
     matrix:
-    - platform: x86
-      arch: i686
-      mingw: mingw32
-      msys: msys32
-    - platform: x64
-      arch: x86_64
-      mingw: mingw64
-      msys: msys64
+    - arch: i686
+      bits: 32
+    - arch: x86_64
+      bits: 64
 
 cache:
   - 'C:\msys32'
@@ -15,22 +11,21 @@ cache:
 install:
   - ps: |
       [Environment]::SetEnvironmentVariable("MSYS2_PATH_TYPE", "inherit", "Machine")
-      $platform = $env:Platform
       $arch = $env:Arch
-      $msys = $env:Msys
-      $mingw = $env:Mingw
+      $bits = $env:Bits
+      $msys = "msys$bits"
+      $mingw = "mingw$bits"
+      $mingwUpper = $mingw.ToUpper()
       $folder = $env:APPVEYOR_BUILD_FOLDER
 
       $env:PATH="C:\$msys\$mingw\bin;C:\$msys\usr\bin;$env:PATH"
 
       function bash($command, $dieOnError = $true) {
         ""
-
         Write-Host $command
 
         $folder = $folder -Replace '\\', '/'
 
-        $mingwUpper = $mingw.ToUpper()
         & "C:\$msys\usr\bin\sh.exe" --login -c "MSYSTEM=$mingwUpper . /etc/profile && cd $folder && $command"
 
         if ($LASTEXITCODE -eq 0) {
@@ -48,7 +43,7 @@ install:
       }
 
       # 32-bit MSYS2 is not offered on AppVeyor yet, so we need to install it.
-      if ($platform -eq "x86") {
+      if ($arch -eq "i686") {
         Write-Host "Installing 32-bit MSYS2..." -ForegroundColor Cyan
 
         # download installer
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -75,11 +75,11 @@ build_script:
       bash "autoreconf --install --include=/$mingw/share/aclocal/"
       bash "./configure"
       bash "make"
-      Copy-Item C:\$msys\$mingw\bin\SDL.dll .\
+      xcopy C:\$msys\$mingw\bin\SDL.dll .\
 
       if ($arch -eq "i686") {
-        Copy-Item C:\$msys\$mingw\bin\libgcc_s_dw2-1.dll .\
-        Copy-Item C:\$msys\$mingw\bin\libwinpthread-1.dll .\
+        xcopy C:\$msys\$mingw\bin\libgcc_s_dw2-1.dll .\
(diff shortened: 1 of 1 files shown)


JosepMaJAZ:
This is to be expected too. The generation depends on some dll files. When run from the shell it finds them, but to distribute it, you have to include them.
Find them in the mingw32/bin and mingw64/bin respectively.  This is explained too in the build instructions.

asbjornu:
@JosepMaJAZ

I see. I missed that part, but when reading it, I'm not quite sure what to do:

> To distribute the application, it is important to bundle the correct version of the SDL.dll file with the executable.

How do I "bundle" the `SDL.dll` file with the `.exe`? Do you mean just placing it in the same directory, or is it possible to merge into the `.exe` file somehow?

Also, I'm wondering if the AppVeyor script should do this:

> If you want to reduce the exe size (removing the debugging information), use the following command:
> 
> ```
> strip -g schismtracker.exe
> ```

What do you think? Perhaps we can distribute binaries both with and without debug information?

jangler:
Yes, place the DLL in the same directory. Schism has never been statically linked AFAIK. And I wouldn't want to distribute two versions of the binary. Most people will have no idea which one they want, and the unstripped binary is only a few megs anyway.

jangler:
Also! I'd prefer a zip file and not a 7z one. Zip stuff is built into modern versions of Windows, but 7z stuff is not (right?). Maybe people who actually use Windows will have another opinion on this.

asbjornu:
@jangler Sure, I can change the format from `.7z` to `.zip`. 😃  I just now saw that I'm missing a couple of more `.dll` files. Adding them in.

asbjornu:
According to [this answer on SO](http://stackoverflow.com/a/4703059/61818), I need to add these compiler flags to avoid the `libgcc_s_dw2-1.dll is missing` problem:

```
-static-libgcc and -static-libstdc++
```

Now, to which command should I add them; `autoreconf`, `configure` or `make`?

asbjornu:
There we go! Now I only need to figure out why `schismtracker.exe` don't seem to support any command line arguments. Is this the case? If so, that makes testing the compiled binary kind of impossible.

ghost:
It supports them fine, do you need it to run headless though? I doubt the program will start with no desktop to display on.

+1 for zip over 7zip, no benefits to using 7zip and most people won't have it installed

asbjornu:
@mb14 Typing `./schismtracker --version` into my OS X terminal, yields:

```
Schism Tracker 20160826
Copyright (c) 2003-2016 Storlek, Mrs. Brisby et. al
```

It would be nice to replicate this on Windows so we could somehow know that the compiled binary works to some extent. Typing `schismtracker.exe --version` into a command line on my local Windows VM here does absolutely nothing, although the executable successfully launches the Schism Tracker GUI if I double-click it.

ghost:
What does typing `schismtracker.exe` into the command line do on your Windows?

asbjornu:
@mb14 It opens the Schism Tracker GUI. The same happens if I type in `schismtracker.exe /?` or any other argument that can be interpreted as a file path or similar. `schismtracker.exe --version` seemingly does nothing.

ghost:
Eh!  In which case this seems like some kind of bug:)

asbjornu:
@mb14 Good to know that it's isn't just me being infinitely noob here. 😄  Until #39 is fixed, I'll just remove the test and come back with another PR when `--version` works. Not sure what else it's possible to do at the command line to verify the binary.

asbjornu:
Hm. [The 64-bit version of the binary](https://ci.appveyor.com/api/buildjobs/a5jyj8q9bjrrhelm/artifacts/schismtracker.zip) doesn't work:

screen shot 2016-08-26 at 01 21 54

[The 32-bit version](https://ci.appveyor.com/api/buildjobs/j3kf4k38jodchtx6/artifacts/schismtracker.zip), however, works fine.

asbjornu:
I only have a 32-bit version of Windows at home, so I'm unable to test the [64-bit version](https://ci.appveyor.com/api/buildjobs/a5jyj8q9bjrrhelm/artifacts/schismtracker.zip), but I hope it works.

asbjornu:
I guess this can be merged, then. Not much I can do to improve it now, unless you have any more comments. 😃

jangler:
Looks good; I have the AppVeyor account set up and the merge commit built. Seems like it's time for a new release soon.

jangler:
Thanks for all the help with CI/automation, by the way.

JosepMaJAZ:
The --version is not working on any build I have, and that includes one that I made out from 20120105 sources back then: http://psycle.free.fr/josepma/SchismTracker-20120105-win32-mingw-3.4.2.zip
In fact, nothing that is written to the console ( printf/ puts ) seems to be printed.

Edit: It seems it is some SDL redirection or similar. I found the text that was supposed to be written, in a file named stdout.txt in the same dir where i am executing it.

http://sdl.beuc.net/sdl.wiki/FAQ_Console

jangler:
Huh! Interesting. I'll paste that into the #39 thread.

asbjornu:
@jangler Excellent!

PR #41, opened 2016-09-05
Test the schismtracker executable on Linux and OS X (Travis)

asbjornu:
This PR adds a test for the `schismtracker` executable that asserts its output begins with "Schism Tracker", just so we know the binary is working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,13 +31,20 @@ addons:
       - libsdl1.2-dev
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
+install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install sdl; fi
+  - wget https://raw.github.com/lehmannro/assert.sh/v1.1/assert.sh
+  - wget https://raw.githubusercontent.com/dansoton/assert.sh/assert-extras/assert-extras.sh
+  - source assert.sh
+  - source assert-extras.sh
 script:
   - autoreconf -i
   - mkdir -p build
   - cd build
   - sh ../configure
   - make
+  - assert_startswith "./schismtracker --version" "Schism Tracker"
+  - assert_end tests
 notifications:
   email:
     on_success: change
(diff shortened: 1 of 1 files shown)


asbjornu:
@jangler, @mb14, are there any reasons this isn't merged? 😃

jangler:
Downloading and sourcing shell scripts during compilation just strikes me as really ugly… is that really considered a best practice for using Travis?

asbjornu:
@jangler I agree that it's ugly, but I thought "whatever happens on Travis stays on Travis", influenced by the earlier rejection of a "proper" build script. I can of course add the scripts as a Git submodule or just copy them into the `scripts` folder, but that would effectively make them a fork, which makes updates and such harder. Thoughts?

asbjornu:
Worth mentioning perhaps is the reason I source the files in the first place: To be able to assert the output from the `schismtracker` executable in a simple way. If you know of simpler ways to do this using built-in Unix command line tools, I'll happily change the assertions. I don't know of a way to get `grep` to fail if it doesn't find the specified regex, for instance.

asbjornu:
Also, the download and sourcing of the `.sh` scripts is technically not done during compilation (which happens during the `script` phase), but in the `install` phase of the build. Not sure that matter much in practice, but I thought it was worth mentioning.

jangler:
> If you know of simpler ways to do this using built-in Unix command line tools, I'll happily change the assertions.

Yeah, that's what I was thinking, but I got sidetracked by work.

> I don't know of a way to get grep to fail if it doesn't find the specified regex, for instance.

It does that by default! Returns exit status 1 if no lines were matched.

```
brandon@desk ~ $ grep '^Schism Tracker' <(schismtracker --version) >/dev/null; echo $?
0
brandon@desk ~ $ grep '^Schism Tracker' <(grep --version) >/dev/null; echo $?
1
```

asbjornu:
@jangler Hm, ok. Exit code 1 should fail the build. I'll try it out!

jangler:
Nice, I was going to ask for a failing test :)

asbjornu:
@jangler Yes, perfect! :+1:

Issue #39, opened 2016-08-25
No --version on Windows


ghost:
see https://github.com/schismtracker/schismtracker/pull/38#issuecomment-242561539

jangler:
JosepMA posted in #38:

> The --version is not working on any build I have, and that includes one that I made out from 20120105 sources back then: http://psycle.free.fr/josepma/SchismTracker-20120105-win32-mingw-3.4.2.zip
> In fact, nothing that is written to the console ( printf/ puts ) seems to be printed.
> 
> Edit: It seems it is some SDL redirection or similar. I found the text that was supposed to be written, in a file named stdout.txt in the same dir where i am executing it.
> 
> http://sdl.beuc.net/sdl.wiki/FAQ_Console

sagamusix:
Windows GUI applications cannot simply write to the console even if they are run from console. They would have to open their own console window. The only way to "fix" this would be to compile the program using the CONSOLE subsystem, but that would always open an ugly console window next to the GUI window, so it's not a nice solution. This is not an SDL limitation. Other applications fix this by e.g. printing the console output to a message box.

jangler:
Case closed! Thanks, Saga.

asbjornu:
@sagamusix Ah, so this explains why most Windows applications suck so utterly in their integration (or lack thereof) with the command line. Thanks!

asbjornu:
However, that still leaves the case unclosed on how we're supposed to do any form of testing of the compiled binary on Windows. Ideas?

sagamusix:
I dunno, why would you require the console for testing? I suppose we're doing things differently for OpenMPT/libopenmpt, but not having the console available doesn't stop us from unit testing. :)

jangler:
If `make` succeeds on AppVeyor but we get an executable that's so broken that it can't parse command-line flags, then someone has most likely pushed or PR'd something that they didn't even bother to build and run. I like sanity checks, but I don't feel that that one is necessary.

Like Saga suggests, unit testing would be much more useful, but the Schism codebase would probably require a lot of refactoring to afford that.

asbjornu:
@sagamusix I agree unit testing would be the absolutely preferred thing to do here, but as @jangler  writes, that would require some major refactoring of the Schism Tracker code base.

[...]

PR #44, opened 2016-09-05
Test the schismtracker executable on Windows (AppVeyor)

asbjornu:
As @mb14 mentions in https://github.com/schismtracker/schismtracker/issues/39#issuecomment-244782966, `schismtracker.exe` outputs a `stdout.txt` file that we can check instead of getting stdout to work in any meaningful way in Windows. This PR adds a test to AppVeyor for just that, similar to #41.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -109,6 +109,8 @@ test_script:
 
 on_failure:
   - dir /s /b > dir.txt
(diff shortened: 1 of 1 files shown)


asbjornu:
@jangler, @mb14, are there any reasons this isn't merged? 😃

asbjornu:
@jangler 🙏

PR #48, opened 2016-09-18
Midi out config

yogiVan:
Will patch and build on RPi

### DIFF
--- a/player/snd_gm.c
+++ b/player/snd_gm.c
@@ -29,7 +29,7 @@
 #include "log.h"
 #include "it.h" // needed for status.flags
 #include "sndfile.h"
-#include "song.h" // for 'current_song', which we shouldn't need
+#include "song.h" // for 'current_song' which provides ->midi_config and possibly other things
 #include "snd_gm.h"
 
 #include <math.h> // for log
@@ -125,7 +125,7 @@ static void MPU_SendCommand(const unsigned char* buf, unsigned nbytes, int c)
 	if (!nbytes)
 		return;
 
-	csf_midi_send(current_song, buf, nbytes, c, 0); // FIXME we should not know about 'current_song' here!
+	csf_midi_send(current_song, buf, nbytes, c, 0);
 }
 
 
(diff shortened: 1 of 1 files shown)


ghost:
I'm not sure if you made this by mistake - a pull request is a request for us to pull things _into_ the project - you don't need to ask permission if you want to pull stuff _out_ :smile: 

You can do it through git or click the green "Clone or download" here and use the "Download zip" option - here's a direct link: https://github.com/schismtracker/schismtracker/archive/midi-out-config.zip

(My sincerest apologies if you knew that already and I was being patronising! We will merge the changes here into master once it's finished and has been thoroughly tested)

Thank you for your help and your interest :heart:

sagamusix:
No matter if this is intended to be merged or not, here are some comments:
1) Calling malloc in real-time mixing routines is a bad idea (there is no guarantee how quickly or slowly a memory allocation is going to be). MIDI macros in IT are limited in size, so both the input and output of the translation function can at most be a char[32] buffer (+1 if you want to play safe with the string terminator).

2) csf_process_midi_macro in effects.c should probably be recycled instead of writing your own translation function (MPU_StringToBytes) which essentially does the same apart from handling macro variables.

ghost:
Good points, thank you! (both done now)

jangler:
That Travis check failed due to some sort of network error with brew/curl; ignore it.

ghost:
yogi has tested this and I guess it works

> \o/ Great Work! Now totally configurable :smile:

jangler:
So what's going on with this?

ghost:
The stuff that is in there is workin' good, yogi has found another bug for me to fix though:

```

> I think there is a bug with Bank Change MSB/LSB messages.
> 1. When I have a Midi Instrument on a track, a MSB of 7F and LSB of FF is sent at the start. These seem to be a default value, and FF is illegal for a CC message (FF is a Sys RT reset message).
> 2. If I set the LSB and MSB slider in the Pitch tab, for example 10 and 10, The MSB will be 0x14 instead of 0x0A. It seems to always be double the slider setting.
> 3. With these sliders set to values, they are sent first when a Note On is encountered which is good. But then every 6 ticks this pattern is sent for the whole track:
> B0    00    7F    1  ---  CC: Bank MSB         
> B0    20    FF    1  ---  CC: Bank LSB         
> B0    00    14    1  ---  CC: Bank MSB         
> B0    20    0A    1  ---  CC: Bank LSB   
> PC messages are sent also, if set, at this same interval.
> For the most part, the best solution is using Zxx macros to send Bank Change and Prog Change messages instead of the sliders. But the MSB/LSB message at the start, sending the FF could reset a tone generator, missing some Note messages. 

```

sagamusix:
> 1. When I have a Midi Instrument on a track, a MSB of 7F and LSB of FF is sent at the start. These seem to be a default value, and FF is illegal for a CC message (FF is a Sys RT reset message).

Sounds like the "-1" value ("don't send") is taken literally, rather than just not sending anything at all?

Bank select messsages and program change messages definitely should only be sent if the bank/program changed; otherwise you will hear a bit of silence while the new program is loaded on some synths; although if it's just a bank select message then it's probably okay, because it only becomes effective once a program change has been sent as well.

ghost:
> Sounds like the "-1" value ("don't send") is taken literally, rather than just not sending anything at all?

Haha! I'll take a look at this when I get home.

PR #57, opened 2016-11-08
Show how to install on OS/X with Homebrew

orip:


jangler:
Ugh, this CI stuff is so broken.

What version of Schism Tracker does brew install?

orip:
20160913 according to the formula.
https://github.com/Homebrew/homebrew-core/blob/master/Formula/schismtracker.rb

PR #62, opened 2016-11-13
Don't cache the MSYS32 folder

### LINKED ISSUE #58: CI stuff broken
Builds are failing for reasons unrelated to code changes.

asbjornu:
Seems like caching the MSYS32 folder causes old packages to break `pacman` in some way, so even though this might increase the build time, I think it's better to have green builds. Fixes #58.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,9 +5,6 @@ environment:
     - arch: i686
       bits: 32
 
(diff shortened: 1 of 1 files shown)


jangler:
Long builds definitely better than spurious bad builds.

PR #65, opened 2016-12-21
We no longer have to allow build failures on Trusty

asbjornu:
As `libsdl1.2` is now included in Ubuntu Trusty Tahr as per travis-ci/apt-package-whitelist#3931, we no longer have to allow build failures on that distro.


jangler:
Forgot we were doing this. Thanks.

PR #78, opened 2017-04-28
drop .PE/.PS macros, add protracker reference

alexmyczko:
see https://sources.debian.net/patches/schism/2:20160521-1/


jangler:
Thanks!

PR #83, opened 2017-06-15
Various largely mechanical cleanups

vcaputo:
These changes largely center around adding a mem_calloc() wrapper to complement the existing mem_alloc() and mem_realloc() helpers, and utilizing the new helper where appropriate.


jangler:
Makes sense to me.

PR #84, opened 2017-06-15
Deduplicate the various midi -> SDL_Event mapping functions

vcaputo:


vcaputo:
I have no MIDI setup to test this, but it seems worthwhile to me.

Last commit only is interesting, branched from my other cleanups for the mem_calloc().  If the other PR lands then this one should rebase to a single commit.

vcaputo:
rebased

vcaputo:
@jangler I got around to some rudimentary testing of this PR using `vkeybd` as the Input, seems to work.

jangler:
This looks good to me, but the name `midi_to_sdl_event()` suggests a function without side effects to me. Maybe `push_midi_event()` or `sdl_push_midi_event()` would be better?

vcaputo:
Agreed, it's something that crossed my mind as well.  Just pushed a version using `midi_push_event()` instead.

edit;
It's early

I initially used your suggested `push_midi_event()` but realized it's a bit inconsistent with the rest of the listing, settled on 'midi_push_event()'.

Issue #75, opened 2017-04-10
MIDI output timing is off


mankeli:
It seems that timing of the midi output (to alsa) is jittering badly. I'm using Linux and using different audio drivers don't seem to have any effect on the issue.

You can easily test this by running Timidity (or other midi-inputting synth) and outputting a straight sequence of notes to it. The sounds should come out precisely, but instead their timing is quite random sometimes and it seems to be affected by the cpu load.

jangler:
Seems like MIDI was not well implemented, since we're seeing so many bugs with it… I wish I had better domain knowledge for this. Maybe @mb14 might know something about it?

ghost:
Yes I get the feeling MIDI was a bit of an afterthought (...and rightly so, I suppose). Sadly I don't have any MIDI output devices or use Linux. I'm not sure what to recommend here but maybe if you can find other users who are having similar problems it'd help pin it down a bit? Sorry. Thanks for the bug report!

mankeli:
@mb14 can you try with your system's internal midi synth?

esaruoho:
@mankeli I'm giving it a shot with just a regular pattern of 16 rows with a GM internal midi synth bassdrum, triggered four times, but with a C00 on the 15th row of the pattern - and there's a yank every time we get to the end of the pattern - i.e., something off with the timing. I'm running on a windows10 and newest schismtracker build. at least with speed 04/128, there's a yank. but if i switch to 06/128, there's no yank.

however, if i'm at 06/128 and have 16 rows of midinotes triggering a piano, there's a certain shuffle and it really does not retain any kind of tempo whatsoever - it's like schism isn't sending midi out very stably at all. it's like there's a random note delay offset on there at all times.

![glinksch](https://cloud.githubusercontent.com/assets/4966687/25267839/19d4790a-267f-11e7-9702-516ea28abf62.png)

mankeli:
@esaruoho This sounds like the behavior I get on Linux too. I suspect that the midi output has been added in a wrong place, and thus not outputting notes with exact timing, but I haven't found the code for this yet.

vcaputo:
@mankeli if you could provide me with specific instructions for getting timidity + schismtracker reproducing what you're experiencing, I'll take a look at fixing it.   I'm on debian jessie x86_64, there's a timidity and timidity-daemon package but I don't know what's involved in getting schismtracker to make noise through timidity, I'm a coder not (yet) a musician.

vcaputo:
[...]

Issue #51, opened 2016-09-29
Crash when playing with Midi ports


HybridDog:
When l try to set the second port in the midi configuration, where it shows Disabled, to output, schismtracker crashes.
There's the gdb backtrace:
http://paste.ubuntu.com/23252596/

jangler:
Can reproduce; will look into this.

jangler:
Can no longer reproduce…? But #64 can.

vcaputo:
@jangler why not just change the assert() to a warning logging the values asserted against, then simply returning prior to scribbling beyond qlen.

Is it unreasonable to expect schism to continue functioning correctly in this state?

logging when this happens may also help?
https://github.com/schismtracker/schismtracker/blob/master/schism/midi-core.c#L641

jangler:
I'm thinking this is just a bug that happens when qlen == 0. I haven't looked at the MIDI code in *that* much detail but we might just need another case for that.

vcaputo:
Hmm... so the implicit hypothesis is that in something of a low-latency configuration where `audio_buffer_samples` is too small for the midi queue to get established the existing code falls over?

@HybridDog if you can still reproduce this, can you supply a screenshot of the message log viewer page from the environment this occurs in?  (Crtl-F11)  That should help confirm the hypothesis.

vcaputo:
@jangler I was able to reproduce the crash, and a fix is waiting in https://github.com/schismtracker/schismtracker/pull/85

HybridDog:
l can still reproduce this, but l don't know what you mean by message log viewer.

vcaputo:
@HybridDog In schism tracker, if you press Ctrl-F11 it will go to the message log viewer page, which should show some statistics about the audio buffer...

However, I've managed to reproduce the problem myself and already posted a fix in #85.  If you're up for testing that PR and verifying it fixes your problem as well it'd be great!

Thanks

vcaputo:
@jangler BTW reproduction proved to be fairly easy here on a plain Debian x86_64 w/ALSA setup.  I just started playing an .it file, and while it was playing went to the MIDI settings and cycled through disabled->input->output on the second entry in the list of ports.  That may prove helpful in testing a fix like #85.

jangler:
Just reproduced it by accident. :) I'll take a look at that PR.

esaruoho:
are ya guys still seeing this crash? i don't use multiple midiports so not sure how to accomplish this crash :)

[...]

PR #85, opened 2017-06-16
Fix midi_queue_alloc() minimum audio buffer size check

vcaputo:


vcaputo:
Only the last two commits are relevant to this PR, builds upon the rest because I added a mem_calloc() cleanup in the last commit.

vcaputo:
rebased

HybridDog:
l can't compile: midi-core.c:(.text+0x15e5): undefined reference to `mem_calloc'

vcaputo:
Thanks for giving it a go!

That's odd, mem_calloc() is being used throughout the code now in master (and the CI checks have passed...)

What steps did you use to fetch and build this branch?

vcaputo:
Here's capture of a fresh clone and fetching/checking out the branch in case you're new to git and that's where this is going awry, note this is Debian GNU/Linux x86_64:
```
foo@laptop:~$ cd /tmp
foo@laptop:/tmp$ git clone https://github.com/schismtracker/schismtracker schism
Cloning into 'schism'...
remote: Counting objects: 19678, done.
remote: Compressing objects: 100% (58/58), done.
remote: Total 19678 (delta 28), reused 30 (delta 18), pack-reused 19602
Receiving objects: 100% (19678/19678), 7.15 MiB | 166.00 KiB/s, done.
Resolving deltas: 100% (13083/13083), done.
Checking connectivity... done.
foo@laptop:/tmp$ cd schism
foo@laptop:/tmp/schism$ git remote add vcaputo https://github.com/vcaputo/schismtracker
foo@laptop:/tmp/schism$ git fetch vcaputo
remote: Counting objects: 54, done.
remote: Compressing objects: 100% (9/9), done.
remote: Total 54 (delta 45), reused 54 (delta 45), pack-reused 0
Unpacking objects: 100% (54/54), done.
From https://github.com/vcaputo/schismtracker
 * [new branch]      add_missing_log_h_includes -> vcaputo/add_missing_log_h_includes
 * [new branch]      fix_math_h_include -> vcaputo/fix_math_h_include
 * [new branch]      gitignore  -> vcaputo/gitignore
 * [new branch]      halfsize-cursor -> vcaputo/halfsize-cursor
 * [new branch]      keyboard   -> vcaputo/keyboard
 * [new branch]      loadinst_fix_keypressbug -> vcaputo/loadinst_fix_keypressbug
 * [new branch]      master     -> vcaputo/master
 * [new branch]      midi_event_dedup -> vcaputo/midi_event_dedup
 * [new branch]      midi_queue_alloc_bug -> vcaputo/midi_queue_alloc_bug
 * [new branch]      misc_cleanups -> vcaputo/misc_cleanups
 * [new branch]      plusplus   -> vcaputo/plusplus
 * [new branch]      strn_dup   -> vcaputo/strn_dup
 * [new branch]      waterfall-loop -> vcaputo/waterfall-loop
foo@laptop:/tmp/schism$ git checkout vcaputo/midi_queue_alloc_bug
Note: checking out 'vcaputo/midi_queue_alloc_bug'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at dc80b1b... schism/midi: use mem_calloc() in midi_queue_alloc()
foo@laptop:/tmp/schism$ autoreconf --install
configure.ac:36: installing './compile'
configure.ac:30: installing './install-sh'
configure.ac:30: installing './missing'
Makefile.am: installing './depcomp'
foo@laptop:/tmp/schism$ mkdir build
foo@laptop:/tmp/schism$ cd build
foo@laptop:/tmp/schism/build$ ../configure && make -j2
... snip ...
foo@laptop:/tmp/schism/build$ ls -l schismtracker
-rwxr-x--- 1 foo foo 3170192 Jun 17 10:33 schismtracker
foo@laptop:/tmp/schism/build$ ./schismtracker
```

HybridDog:
l first applied the patch of this PR (https://github.com/schismtracker/schismtracker/pull/85.patch), then run make. After l saw that compiling didn't work l run ../configure and then make again (in the build/ folder [as documented](https://github.com/schismtracker/schismtracker/blob/master/docs/building_on_linux.md#building-schism-tracker)) and it still didn't work.

vcaputo:
@hybriddog did you pull the latest master before applying the patch?  Because mem_calloc() *just* landed in master, so if your master is stale that may explain why this isn't linking.

HybridDog:
l didn't, thanks, there lied the rub.
Anyway, it still crashes for me:
> schismtracker: ../schism/midi-core.c:763: midi_send_buffer: Assertion `((unsigned)pos) < ((unsigned)qlen)' failed.

vcaputo:
Hmm, can you show us what the log viewer says so we can see what the buffer size is on your system? (Ctrl-F11)  and are you certain your build has the patch applied after you got the latest master?

I am able to reproduce the crash here and my fix certainly eliminates it on my system, seen in Ctrl-F11:
```
Using driver 'alsa'
44100 Hz, 16 bit, stereo
Buffer size: 1024 samples
```

HybridDog:
Strangely, Ctrl+F11 doesn't work in schismtracker for me, but Ctrl+F12 and others do.
Edit: The lubuntu key combinations blocked ctrl+ F7, F10 and F11…

vcaputo:
Maybe you can tweak your lubuntu/window manager config to not swallow those keys?

If it's annoying you could easily change the log_appendf() calls in audio_playback.c lines 1325-1327 to use printf() and capture the terminal output instead of accessing the builtin log viewer.

HybridDog:
2017-06-17-224133_1920x1080_scrot

vcaputo:
Thanks, that's helpful, and I suspect there's still a bug in the arithmetic which I did not closely scrutinize yet.  At a glance 10ms of 48000HZ should be 480 samples, which is less than 512, so should work unless I'm missing something.

vcaputo:
@HybridDog I just force-pushed updates to this branch with more changes, are you willing to test?  My setup doesn't have such a small audio buffer... not that I have an actual MIDI arrangement to confirm this doesn't break things badly in other ways.  Thanks!

vcaputo:
@jangler I'm unclear why this function even wants to deal in units of bytes at all, doesn't it make more sense to simply feed it the number of samples in the audio buffer and the samples per second, and simply always work entirely in units of samples?  Any ideas there?  why do we care about bytes per ms?

HybridDog:
l tested it, it doesn't crash. But when l set it to output, schismtracker is very laggy.

jangler:
@vcaputo like much of the other MIDI code in Schism, I find the original code that you touched baffling. So I have no clue.

@HybridDog yeah, that's been another complaint about Schism's MIDI (see #75). Not sure if it's directly related to this code or something else.

vcaputo:
@sagamusix @storlek any ideas?

vcaputo:
@HybridDog Any interest in trying out #93 on top of these changes and seeing if it fixes what you're perceiving as "laggy"?   I found that interactively playing notes for MIDI output was simply broken, but playing from patterns seems to work OK.

edit: that PR landed, so if you'd be up for simply rebuilding and seeing how this PR works for you now, feedback appreciated. Thanks

vcaputo:
fixed some typos in commit message, nothing interesting changed.

HybridDog:
l rebuilt and it's still lagging.

jangler:
@HybridDog What exactly do you mean by "lag" in this context?

HybridDog:
By lag l mean l get a low framerate and it uses much cpu (30%).

jangler:
Hm, I can get only get Schism to use that much of a CPU core with MIDI if I'm sending 600+ new notes per second, and my CPU cores aren't even that powerful by modern standards (1.8 GHz). Even then, I don't have any framerate or UI problems…

vcaputo:
@HybridDog would it be possible to get a `valgrind --tool=callgrind` profile of this?

Also maybe we can reproduce it by using `--audio-driver=dummy` with the dummy driver hacked to use 512 samples, since that small buffer is unique to your configuration.

jangler:
Mine's also 512 samples. What's yours?

vcaputo:
1024 @ 44.1Khz when on ALSA.  I wonder what's required to reproduce what HybridDog's experiencing then.

vcaputo:
@HybridDog is this when playing a particular song, simple interactive notes/pattern editing, or even just sitting idle?

HybridDog:
It happens when playing a song. When nothing is played, it doesn't lag.

HybridDog:
It happens when playing a song. When nothing is played, it doesn't lag.

Here's the valgrind result: callgrind zip

vcaputo:
Thanks for the valgrind output.

Can you confirm this lag only occurs during MIDI playback and **not** when playing files internally mixed without any MIDI output?

There's no smoking gun in the callgrind files, it all just looks like graphics rendering overhead.

HybridDog:
> Can you confirm this lag only occurs during MIDI playback and not when playing files internally mixed without any MIDI output?

What do you mean by MIDI playback?
When setting the midi through to output (or duplex), it doesn't lag, the lag only transpires when doing it with one of the multicast/ip things.
2017-06-22-151315_1920x1056_scrot

vcaputo:
I see, I've reproduced the lag by setting a `Multicast/IP MIDI` port to output.

Do you actually have an `IP MIDI` device you're trying to work with?

If anyone can suggest a way to end-to-end test `IP MIDI` it'd be greatly appreciated.  I've been using TiMidity++ locally exclusively for testing my changes.  If I make changes to try fix the UI latency with `IP MIDI` they'll need to be tested for correctness...

HybridDog:
No, as far as l know, l don't have such a device.

vcaputo:
Ok, so the lag seems to be somehow caused by the last commit added to this PR `schism/midi: fix buffer bytes per ms arithmetic`, so it appears there is significance to using there 80 instead of 100.

I'm dropping that commit, but we should still land this PR because it fixes the buffer size check making midi_queue_alloc() assert less erroneously - just not for when your buffer size is so small it's on the fringes of 10ms worth of buffer.

@HybridDog can you re-test and verify that the lag is gone with that commit dropped from this PR.  You can work around the buffer being too small instead by editing the schism config file and changing buffer_size=512 to buffer_size=1024.

HybridDog:
l tested it and it crashed. In the config file buffer_size is set to 1024, but ctrl+f11 says 512.
When l set buffer_size to 2048 in the config file, in schismtracker it says 1024 and it doesn't crash or lag.

jangler:
(FWIW I know zilch about multicast IP MIDI.)

esaruoho:
would this make macosx midi more stable?

jangler:
@esaruoho, did you test MIDI again as requested in #75? If so then please post your results there.

esaruoho:
@jangler posted some results + examples to #75 . hope this helps! :)

esaruoho:
@jangler hey, just saw @vcaputo post in #75 talking about this very specific pull request (#85 ) - the commentary is here: https://github.com/schismtracker/schismtracker/issues/51#issuecomment-355832908

could we try this out and see if it helps with the input/output mad-switching?

vcaputo:
I just force-pushed a rebased version in case master had anything important since this was last touched.

jangler:
The reason I hadn't merged this was that @HybridDog's build crashed. Has that been addressed? Or does it also crash for that reason in master, or what?

vcaputo:
@jangler Well the PR doesn't remove the assert, so one can still have a legitimately too-small (sub 10ms) buffer.

The PR obviously fixes a bug since the 'channels' parameter is being supplied as 'audio_sample_size', not the number of channels, at the call site.  This PR fixes the parameter and variable names to reflect what's actually happening and normalizes the units to all be bytes.  Though maybe it would be preferable to normalize on samples and fix the caller to actually supply the number of channels, I haven't scrutinized the code again to see if that's the better approach.  I don't believe there's a correctness difference either way.   I can probably take a look at it later today.

jangler:
Oh yeah, I remember looking at this. Looks fine to me and pretty much anything sane is better than what the old MIDI code does. You can still take another look if you want, but I'm just going to merge this now.

jangler:
(Thanks! I always forget to say thanks but I really do appreciate when people contribute.)

Issue #86, opened 2017-06-16
Scrolling in "Load Instrument" screen doesn't work


zoikkis:
In all the other Schism Tracker screens, when you keep arrow up/down pressed down, cursor starts soon scrolling automatically as it should, including Load Sample screen. However when in Load Instrument screen, pressing and keeping arrow button down doesn't do anything. Only when you raise the button it moves one increment but doesn't start scrolling at any point.

I've noticed the same behavior with several computers running different windows versions. Is there a reason behind this or is it a bug?

jangler:
Hm, I can't reproduce this.

vcaputo:
I can, by simply going to the instrument library page (Ctrl-F4), up/down arrows do nothing until released.

vcaputo:
https://github.com/schismtracker/schismtracker/pull/87 seems to fix it, unsure if it breaks anything else, but at a glance I don't see why it would - but this code as-is kind of doesn't make sense, maybe it's a copy and paste from another listing which then got hacked up a bit leaving this oddity?

jangler:
Oh, my bad; I misread and was looking at the Instrument List screen instead. Yeah, I can reproduce it.

vcaputo:
@zoikkis in case it's unclear, this has been fixed in master and will be in the next release - thanks for the bug report!

PR #87, opened 2017-06-16
schism/page_loadinst: stop swallowing KEY_PRESS

vcaputo:
For some reason the code was short-circuiting on k->state == KEY_PRESS
in a k->mouse state default/fall-through case.  This prevented regular
key press/repeat handling from functioning on the loadinst page.

Fixes https://github.com/schismtracker/schismtracker/issues/86

### DIFF
--- a/schism/page_loadinst.c
+++ b/schism/page_loadinst.c
@@ -477,9 +477,6 @@ static int file_list_handle_key(struct key_event * k)
 	} else if (k->mouse == MOUSE_DBLCLICK) {
 		handle_enter_key();
 		return 1;
-	} else {
-		if (k->state == KEY_PRESS)
-			return 0;
 	}
 
 	new_file = CLAMP(new_file, 0, flist.num_files - 1);


jangler:
I hate that the logic for load file screens is duplicated in the first place… there are a few other minor inconsistencies in the code, including superfluous conditionals and statements.

I think that the lines of code in question are actually supposed to check against `KEY_RELEASE`, [like they do in `schism/page_loadsample.c`](https://github.com/schismtracker/schismtracker/blob/master/schism/page_loadsample.c#L734-L736). Haven't looked at the full implications of this, though. Might be redundant?

jangler:
Seems like even the (an) author was questioning the need for / purpose of this sort of statement; see [schism/page_loadmodule.c](https://github.com/schismtracker/schismtracker/blob/master/schism/page_loadmodule.c#L772-L775).

jangler:
It's actually maddening that the only other comments in these switch statements are instances of `/* else fall through */`.

jangler:
Alright, I think the point of those cases is just so that the up, down, pgup, pgdn, home, and end cases can be more concise. The change you made here doesn't work since when implemented, (e.g.) a single down arrow key press moves the cursor two steps.

I'm going to close this and change that `else` logic to match the module and sample loading logic.

vcaputo:
Ok, sgtm.

Agreed WRT this being one of the fairly crufty facets of the schism code.

PR #88, opened 2017-06-16
Introduce strn_dup() helper and use where applicable

vcaputo:
There are numerous places, particularly in the fmt/* code, where nul-terminated strings are being allocated, copied, and terminated manually.  This is unnecessarily complex, verbose, and error-prone.

Add a helper called strn_dup() which exits on allocation failures like the other MALLOC helpers, and simplify all the obvious instances by using it.

Note rather than simply wrapping strndup() a manual implementation has been retained for portability reasons.

### DIFF
--- a/fmt/xi.c
+++ b/fmt/xi.c
@@ -107,9 +107,7 @@ int fmt_xi_read_info(dmoz_file_t *file, const uint8_t *data, size_t length)
 		return 0;
 
 	file->description = "FastTracker Instrument";
-	file->title = mem_alloc(24);
-	memcpy(file->title, xi->name, 22);
-	file->title[22]='\0';
+	file->title = strn_dup((const char *)xi->name, 22);
 	file->type = TYPE_INST_XI;
 	return 1;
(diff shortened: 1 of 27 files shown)


mischief:
lgtm

vcaputo:
Thanks for taking a look Nick!

fmt/xi.c
```diff
@@ -107,9 +107,7 @@ int fmt_xi_read_info(dmoz_file_t *file, const uint8_t *data, size_t length)
 		return 0;
 
 	file->description = "FastTracker Instrument";
-	file->title = mem_alloc(24);
```
jangler, line 110: Rhetorical question: why on earth was this allocating 24 bytes in the first place?

jangler:
Nice clean-up!

fmt/xi.c
vcaputo: I guess it's a typo and the author intended 22 + 1 for the nul terminator?

PR #89, opened 2017-06-16
gitignore: add autoconf/configure/build artifacts

vcaputo:
Having `git status` scroll with a bunch of files you'd never want
checked-in anyways makes its output pretty meaningless on all but
virgin trees.


vcaputo:
Maybe omitting these from .gitignore was done intentionally, but in my preferred workflow `git status` prints nothing when there are no modifications.  As-is, if you've configured and/or built schismtracker, `git status` will scroll all sorts of stuff even when there are no git-relevant modificatoins.  This PR attempts to improve that situation.

jangler:
The build instructions in the docs in the repo include:

`mkdir -p build && cd build && ../configure && make`

… so that it's sufficient to ignore `build/` and not all the other files that are generated in the build process. The only build files assumed to be left in the root folder (and therefore ignored) are the ones generated by `autoreconf -i`, which can't be helped.

I concede that the naïve (read: usual) build protocol would not involve the creation of a build folder, but IMO the directory tree is a *lot* cleaner if you do.

vcaputo:
That's fair, I will adjust my workflow accordingly when working with the schismtracker code.

Issue #90, opened 2017-06-17
Specific MOD module sounds weird


Abc120:
The module "folk'n'storm" by Miguel Samiez doesn't sound correct when using Schism Tracker. XMPlay plays it just well, though.
P.S. For some reasons, zip archive I'm trying to add can't be attached, so I have to rename it into .png.
P.P.S. Anyway, your tracker is best of all, keep up good work)

![folk_n_storm](https://user-images.githubusercontent.com/29509021/27255354-dfe8e792-53ad-11e7-8d8d-afe0330dcb32.png)

jangler:
Actually, the latest build of Schism segfaults for me when I try to open that module (IT and Milky have no problem).

HybridDog:
When l try to play this song, l also get a segfault crash, valgrind tells that:
https://pastebin.com/raw/NaihQrkh

jangler:
I'm not sure that perfect MOD playback is a goal of Schism Tracker though, just as XM and 669 and so on are supported for import but no guarantees beyond that. If IT itself plays it back "correctly" then it's definitely worth looking into, but unfortunately I can't run IT in DOSBox at full speed, so I can't really tell…

jangler:
@Abc120, what version of Schism Tracker are you using? That info might help us figure out where the segfault bug was introduced.

HybridDog:
Invalid memory access can also be a security problem l guess. Some .mod file may access memory containing settings and change that.

vcaputo:
@Abc120 https://github.com/schismtracker/schismtracker/issues/91 fixes the crash for me, if you're up for testing the patch or really want to play that mod in schism asap feel free to build and kick the tires!

Abc120:
Thanks for your quick replies! The version I'm using is 20170420, and it doesn't crash, it simply plays this specific module incorrectly. I didn't build from sources, just downloaded binaries from the main page. And yes, I know playing MODs is not a top priority task, but it's still a bug since the tracker supports such a format, and supports it perfectly for all other modules I tried. The problem is just this specific module. It would be really nice to have it fixed)
Oh, BTW my system is Windows 10 x64, in case it matters.

vcaputo:
Oh, well, I'm not the ideal person for correcting playback issues.  Maybe one of the more experienced schism devs can take a look at that.  I was under the impression the main problem was segfaulting on load, as the issue title indicated, which is indeed an existing obvious bug with a trivial fix.

HybridDog:
[...]

PR #91, opened 2017-06-17
player: add csf_adjust_sample_loop() bounds check

vcaputo:
This function accesses data[len-1] so when len is 0 it goes out-of-bounds.

Simply add a len < 1 qualifier to the already present !data short-circuit.

Fixes https://github.com/schismtracker/schismtracker/issues/90

### DIFF
--- a/player/csndfile.c
+++ b/player/csndfile.c
@@ -1116,7 +1116,7 @@ uint32_t csf_read_sample(song_sample_t *sample, uint32_t flags, const void *file
 
 void csf_adjust_sample_loop(song_sample_t *sample)
 {
-	if (!sample->data) return;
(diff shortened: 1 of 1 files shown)


vcaputo:
poor @jangler has to review all this stuff by himself apparently

HybridDog:
It fixed the crash for me, that .mod sounds weird now. With openmpt123 and libmodplug, it sounds better.

jangler:
> poor @jangler has to review all this stuff by himself apparently

But I appreciate your help! Even if it makes more work for me. 😜

Anyway, this fix makes sense to me.

Abc120:
Well, it fixes the crash (which isn't even present in stable version), but what about sound?

PR #92, opened 2017-06-18
schism/slurp: propagate peeked count in slurp_read

vcaputo:
The addition of slurp_peek() lost the short-read clipping of
count done by slurp_read() by ignoring the slurp_peek() return value.


vcaputo:
@jangler Just something I noticed reviewing the recent changes...

jangler:
Good catch!

PR #93, opened 2017-06-18
schism: never suppress starting notes from MIDI

vcaputo:
Due to how csf_process_tick() calls csf_midi_out_note() on every tick,
_schism_midi_out_note() checks the tick count itself and filters most
calls from reaching csf_process_midi_macro() when tick count is != 0.

This prevents MIDI notes from reliably being played interactively in the
pattern or instrument editors.  The user basically must hammer the key
repeatedly until luckily hitting !tick_count.

This commit makes the the (tick_count != 0) short-circuit conditional on
there not being an explicitly supplied note to _schism_midi_out_note(),
which only occurs from csf_process_tick() on non-zero ticks.

This way interactive use, which always supplies a note, will no longer be
tick_count sensitive.

### DIFF
--- a/schism/audio_playback.c
+++ b/schism/audio_playback.c
@@ -1015,8 +1015,9 @@ void cfg_save_audio(cfg_file_t *cfg)
 }
 
 // ------------------------------------------------------------------------------------------------------------
-static void _schism_midi_out_note(int chan, const song_note_t *m)
+static void _schism_midi_out_note(int chan, const song_note_t *starting_note)
 {
+	const song_note_t *m = starting_note;
 	unsigned int tc;
 	int m_note;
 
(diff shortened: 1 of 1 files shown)


vcaputo:
@jangler this bug is fairly easy to reproduce by simply running a timidity daemon, and starting schismtracker with the dummy audio driver.  Just configure a midi output port in schism to use timidity, set up an instrument on a midi channel, and start playing notes interactively and/or editing a pattern with that instrument.

The existing code from master won't play the notes the vast majority of the time, with random bursts of sound.  With this patch the notes play reliably.

See the commit message for more explanation...

jangler:
I guess that check was mainly there to avoid extra computation? We end up processing a lot of extra cells that would have been skipped without this change, but that's probably trivial and no unnecessary MIDI events are sent (that I'm aware of). Fix works well for me!

vcaputo:
It's unclear to me why the existing code was the way it was, it just seems like an oversight - and it made tracking w/MIDI basically unusable so I suspect we've simply got a somewhat unpolished MIDI implementation on our hands.

Is Mrs. Brisby still alive/reachable?  Getting their input on some of these outstanding MIDI issues would probably be super helpful.

PR #94, opened 2017-06-19
schism: fix MIDI "all notes off" on stop

vcaputo:
The existing code sent a sequence of:
1. all sound off
2. reset all controllers
3. all notes off

Which are all channel control messsages, and must be sent for every channel.
A channel of 0 was hard-coded for all of these, which prevented F8 from
effectively stopping MIDI notes when any channels other than the first were
in use.

This commit just sends the same sequence, for all the channels.

### DIFF
--- a/schism/audio_playback.c
+++ b/schism/audio_playback.c
@@ -567,11 +567,19 @@ void song_stop_unlocked(int quitting)
 			moff[0] = 0xe0 + j;
 			moff[1] = 0;
 			csf_midi_send(current_song, (unsigned char *) moff, 2, 0, 0);
+
+			moff[0] = 0xb0 + j;	/* channel mode message */
+			moff[1] = 0x78;		/* all sound off */
+			moff[2] = 0;
+			csf_midi_send(current_song, (unsigned char *) moff, 3, 0, 0);
+
+			moff[1] = 0x79;		/* reset all controllers */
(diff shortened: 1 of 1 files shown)


vcaputo:
Unless I'm misunderstanding how MIDI is expected to be used, you actually want to send these to all 16 channels when stopping.

In my testing I was taking preexisting mod files, and assigning a MIDI channel to each of their instruments for playback via timidity.  This tended to accumulate notes over time which never would decay eventually overwhelming timidity.  With this change, I can recover from such a scenario by simply stopping the playback with F8 and timidity gets all its channels cleaned up.  Previously F8 was utterly ineffective and I had to exit schism.

vcaputo:
For repro in master, simply play the attached IT file via `schismtracker --audio-driver=dummy` with TiMidity++ via MIDI output.
[folkmid.zip](https://github.com/schismtracker/schismtracker/files/1095678/folkmid.zip)

After letting it play for awhile, you'll have a bunch of accumulated notes playing outside MIDI channel 0 which never stop on their own.  Try stopping the sound with F8, and you cannot do it.

Repeat with this PR applied, F8 works.

jangler:
I'm not a MIDI expert either but I see no reason not to do this.

(I think there's a deeper problem that needs to be fixed here though; those notes should not be accumulating in the first place…)

Issue #96, opened 2017-06-19
schism appears to quickly lose track of MIDI notes overlapping in the same channel


vcaputo:
In testing MIDI integration fixes, I've come to use complex preexisting tracked songs by configuring their instruments on distinct MIDI channels with MIDI output configured for TiMidity++, and the "dummy" audio driver.

This tends to work well but after not long, at most typically one repeat of the song, usually less, notes seem to be overlapping accumulating until becoming a noisy pile of chaos with TiMidity++ hogging a CPU, eventually becoming a popping and clicking mess unable to even keep the audio buffer full.

Maybe this is unavoidable due to how MIDI polyphony works.  I'm not a MIDI expert by any stretch of the imagination.  I haven't investigated if MIDI returns anything like a note instance ID in response to NOTEON commands in polyphony mode, making it possible for schism to specifically address an overlapping note with a potentially long (infinite?) sustain with a NOTEOFF.  Such a note seems to currently keep playing in its channel forever while new instances keep getting added, accumulating into what I'm observing.

MIDI does expose per-channel control commands like "all notes off" which schism can use to sweep up such a mess when it happens, restoring the clean slate.  Schism does attempt to do this on F8/song stop here: https://github.com/schismtracker/schismtracker/blob/20170420/schism/audio_playback.c#L570

Being a channel-specific message however, it fails to clean up notes in any channels other than channel 0 as currently implemented.  I've already posted a PR for fixing this issue which sends these messages for all 16 channels (https://github.com/schismtracker/schismtracker/pull/94), which at least enables the user to silence all the MIDI polyphony with a simple F8.

Perhaps that's the best we can do, I'm creating this issue to hopefully start a discussion about the correctness of schism's existing MIDI polyphony usage and if it's exceptionally bad with regards to losing track of overlapping notes.

jangler:
MIDI is a one-way protocol; live MIDI input is the same data as used in a MIDI file but without time interval information (because the time an event happens is just when the event is received in real-time). So there isn't anything like a note ID "returned" by a MIDI sink.

I have no idea why this would happen, though. :P

vcaputo:
It's probably unavoidable for long-lived notes to stack up then.  #94 makes it **a lot** less of an issue since the user's empowered to authoritatively wipe the MIDI state clean by pressing F8.

I guess we just have to discern if the current scope of note accumulation is appropriate and expected or if schism is exacerbating it due to a bug.  Man that's annoying...

[...]

PR #101, opened 2017-06-22
Send MIDI note off to last instrument, not current one (#96)

jangler:
Previous behavior was resulting in accumulation of "stuck" notes when
switching between instruments that used different MIDI channels in the
same IT channel.


jangler:
@vcaputo, look good to you?

vcaputo:
Nice man!  my test file sounds much better with this change, and the code looks right to me.

PR #111, opened 2017-06-29
vgamem-scanner: handle regular characters first

vcaputo:
This trivial optmization rearranges the if-else-if-else chain to
catch the most common regular characters early instead of last.


vcaputo:
Not that schismtracker is CPU-bound on any of my machines, but this is an obvious optimization in what all the valgrind profiles I've seen present as the hottest function by a wide margin.

esaruoho:
@jangler is this somewhat okay to be merged? I have no idea what it does. just wondering

jangler:
Yeah, I didn't prioritize it since it's a small optimization. Seems fine now that I've looked at it.

jangler:
(Thanks!)

PR #114, opened 2017-07-08
Add information about custom gcc compiler flags and update configure.ac

HybridDog:
use multithreading for make and show info about custom settings

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -261,7 +261,7 @@ dnl (This ought to be above AC_PROG_CC, but that causes configure to fail
 dnl when all the insane warnings are are enabled.)
 
 AC_ARG_ENABLE(extra-opt,
-        AS_HELP_STRING([--enable-extra-opt], [Add extra optimizations]),
+        AS_HELP_STRING([--enable-extra-opt], [Add extra, system-dependent optimizations (do not use this together with debug or even profiling)]),
         ADD_OPT=$enableval,
         ADD_OPT=no)
 
@@ -276,7 +276,7 @@ AC_ARG_ENABLE(debug,
         ADD_DEBUG=no)
 
 AC_ARG_ENABLE(profiling,
-        AS_HELP_STRING([--enable-profiling], [Enable profiling flags (slows things down)]),
+        AS_HELP_STRING([--enable-profiling], [Enable profiling flags (slows things down), debugging (--enable-debug) is also enabled by this]),
         ADD_PROFILING=$enableval,
         ADD_PROFILING=no)
 
@@ -298,9 +298,10 @@ fi
 dnl place extra optimizations after existing cflags so that they can override
 dnl override whatever flags might exist by default (-g -O2 usually)
 if test x$ADD_OPT \!= xno; then
-        ADD_OPT="-g0 -s -O3 -ffast-math -fomit-frame-pointer -fno-exceptions"
-        ADD_OPT="$ADD_OPT -funroll-loops -frerun-cse-after-loop -fno-ident"
-        ADD_OPT="$ADD_OPT -fno-strength-reduce"
+        if test x$ADD_DEBUG \!= xno || test x$ADD_PROFILING \!= xno; then
+                AC_MSG_NOTICE([You aren't supposed to use --enable-debug or --enable-profiling together with --enable-extra-opt!!])
+        fi
+        ADD_OPT="-Ofast -g0 -s -fno-exceptions -march=native"
         CFLAGS="$CFLAGS $ADD_OPT"
 fi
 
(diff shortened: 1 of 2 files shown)


docs/building_on_linux.md
```diff
@@ -19,24 +19,37 @@ See below for distro-specific instructions on how to get everything installed
...
+copied anywhere you like on the filesystem.<br>
+You can specify custom compiler flags, e.g. to optimize schismtracker
+system-dependently:
+```sh
+make clean  # recompiling needed after changing flags
+CFLAGS="-Ofast -march=native" ../configure --enable-extra-opt
+make -j $(nproc || sysctl -n hw.ncpu || echo 2)
+```
+For debugging, you can set the flags to <samp>"-g -Og"</samp>.<br>
+To get more information, run ```../configure --help```.
```
HybridDog: The Og flag is ignored,
when passing --enable-profiling to ../configure, the flags are set to -g -O2 -g. Is this supposed to happen? Shouldn't the flags be set to -g -Og when passing this?

jangler:
I dislike the formatting changes. Using triple backquotes for syntax highlighting is a GitHub-specific Markdown extension, and it makes the plain text less readable, *and* here it's used for code that you're going to type or paste into a shell, where it won't be syntax highlighted anyway. I'd rather avoid HTML in Markdown as well, since again it makes the plain text less readable.

I'm OK with adding information about possible useful compiler flags, but the text you added doesn't make it immediately clear to me what the additional flags are doing or why.

HybridDog:
> the text you added doesn't make it immediately clear to me what the additional flags are doing or why.

I'm still wondering what the configure flags are. When passing --enable-profiling and --enable-debug, it adds three times the -g flag and the -O2 flag. There's no Og flag added.
Why is optimation set to O2 and not to O3 by default anyway?

vcaputo:
I believe autotools uses `-g -O2` as a safe default which works acceptably across a broad spectrum of compilers, perhaps this was a similar choice.

HybridDog:
Yes, l guess that's the reason why gcc allows overriding it by passing another parameter. The last one is used (according to the manpage). l added the -Og gcc flag to debugging, before the -O2 flag wasn't overridden, so debugging may work better after this.

docs/building_on_linux.md
```diff
@@ -33,11 +33,21 @@ schismtracker directory and running:
...
 copied anywhere you like on the filesystem.
 
+You can specify custom compiler flags, e.g. to optimize schismtracker
+stronly, system-dependently:
+
+    make clean  # recompiling needed after changing compiler setting
+    ../configure --enable-extra-opt
+    make -j $(nproc || sysctl -n hw.ncpu || echo 2)
+
+For debugging, and other settings, see ```../configure --help```.
```
asbjornu: Only multi-line code snippets require three back-ticks. See how `schismtracker` is formatted on line 39.

HybridDog: fixed

HybridDog:
I changed the Og flag to O0 because it gives better debugging results.

HybridDog:
@jangler What about now?

esaruoho:
now that appveyor has been fixed - could this be re-run somehow to see if there's still errors?

docs/building_on_linux.md
```diff
@@ -30,14 +30,25 @@ schismtracker directory and running:
...
 ## Building Schism Tracker
 
-To build Schism Tracker, you should set up a build-directory. From the
-schismtracker directory:
+To build Schism Tracker, you should set up a build-directory and compile from
+there. From the schismtracker directory:
 
-    mkdir -p build && cd build && ../configure && make
+    mkdir -p build && cd build && ../configure && \
+    make -j $(nproc || sysctl -n hw.ncpu || echo 2)
```
jangler: I'd rather keep things simple here and have the basic build instructions just use `make` without any optimization flags, since multithreading flags are mentioned below for anyone who wants them.

configure.ac
```diff
@@ -298,9 +298,10 @@ fi
 dnl place extra optimizations after existing cflags so that they can override
 dnl override whatever flags might exist by default (-g -O2 usually)
 if test x$ADD_OPT \!= xno; then
-        ADD_OPT="-g0 -s -O3 -ffast-math -fomit-frame-pointer -fno-exceptions"
-        ADD_OPT="$ADD_OPT -funroll-loops -frerun-cse-after-loop -fno-ident"
-        ADD_OPT="$ADD_OPT -fno-strength-reduce"
+        if test x$ADD_DEBUG \!= xno || test x$ADD_PROFILING \!= xno; then
+                AC_MSG_NOTICE([You aren't supposed to use --enable-debug or --enable-profiling together with --enable-extra-opt!!])
+        fi
+        ADD_OPT="-Ofast -g0 -s -fno-exceptions -march=native"
```
jangler, line 304: Does `-Ofast` obviate all the `-f` flags or something?

jangler: Never mind, I just saw your commit comment explaining this.

docs/building_on_linux.md
HybridDog: done

jangler:
AppVeyor should work now but I wouldn't be surprised if some Travis builds take an hour and then fail. Not sure what's up with that; I'm going to check if there's any pattern or if it just seems random.

jangler:
The majority of the Travis errors seem to just be problems with cloning the repo. I can understand why that might happen if I merge a branch with non-code changes and delete that branch before all the builds are run, but I don't know why it sometimes fails to clone master.

The other problems are inconsistent and have to do with being unable to download dependencies with brew or apt-get. It's not specific to any build configuration; it just fails sometimes and works other times.

jangler:
Anyway, thanks for the PR!

PR #116, opened 2017-07-10
Only release self-created objects

### LINKED ISSUE #109: Crash on macOS Sierra 10.12.5
Launching `schismtracker` on macOS Sierra 10.12.5 crashes it with the signal `Trace/BPT trap: 5` on the command line and [this problem report](https://gist.github.com/asbjornu/b52121cb8b6459c7323006d744916569). `schismtracker --version` runs just fine (and reports the version number `20170628`), so we might want to find a better test to execute on Travis to detect crashes like these.

I've performed a fresh build of Schism Tracker, although the `Version` reported in the problem report is `20120105`. I assume that's related to #77.

The only intelligible portion of the problem report I can find is this:

> Detected over-release of a `CFTypeRef`

Ideas?

asbjornu:
This PR changes the clean up part of `get_ep_name()` within `midi-macosx.c` such that it only performs `CFRelease` on the `fullName` variable if it was instantiated through the `CFStringCreateWithFormat()` function. This fixes #109, at least on my macOS 10.12.5 iMac.

Ideas on how to ensure that this continues to work in the future would be welcome. The current `schismtracker --version` test does not exercise the code in question, so it would be great if we could come up with a more extensive test to ensure that the `schismtracker` binary at least launches successfully on all platforms. Ideas?

### DIFF
--- a/sys/macosx/midi-macosx.c
+++ b/sys/macosx/midi-macosx.c
@@ -123,9 +123,7 @@ static char *get_ep_name(MIDIEndpointRef ep)
 			defaultEncoding);
 
 	/* clean up */
-	if (endpointName) CFRelease(endpointName);
-	if (deviceName) CFRelease(deviceName);
-	if (fullName) CFRelease(fullName);
+	if (fullName && !deviceName) CFRelease(fullName);
 
 	return newName;
 }


asbjornu:
Seems like Travis build failed due to a glitch on GitHub:

> fatal: unable to access 'https://github.com/Homebrew/brew.git/': Could not resolve host: github.com

Running the build again should fix the problem and turn the red cross into ✅ .

PR #132, opened 2018-01-05
Actual functioning step-by-step OSX compile instructions

### LINKED ISSUE #131: OSX: High Sierra: "Illegal Instruction: 4"
What could
```
Illegal Instruction: 4
```

mean? That's what I get when I run schismtracker in the terminal on a macOS High Sierra. I've tried brew installing a schismtracker but even though I did it, and it succeeded in installing, I still get the same error.

esaruoho:
Here are instructions on how to clone the repo, install homebrew, get the required dependencies, compile the app and update the .app and copying it to /Applications.

@jangler By following these instructions, I was able to create a fully working ./schismtracker and a fully working Schism_Tracker.app. 

~This documentation is also a reaction to your "hey it's the same as linux - maybe i should specify it more clearly" - no it bloody well ain't the same, and no, instructions on compiling it to MacOSX do not in _any way_ benefit from being told to check some other bit of documentation unrelated to MacOSX...!!!!~

Hope you will find this change worth merging into the repo itself.
  

Closes #131

### DIFF
--- a/docs/building_on_osx.md
+++ b/docs/building_on_osx.md
@@ -1,76 +1,59 @@
 # Building on OS X
 
-Building on Mac OS X is fairly straightforward once you've installed the Apple
-Developer's kit. If you have [Homebrew](http://brew.sh/) available, you should
-be able to install SDL with a simple `brew install sdl`, after which you can
-mostly follow the Linux instructions and you'll end up with a unix-style
-`schismtracker` binary that you can run from `Terminal.app`.
+Start by installing [Homebrew](http://brew.sh/). Open up the Terminal and paste in the following command.
 
-Alternately, you can also install the [SDL Developers
-Library](http://libsdl.org/download-1.2.php), but this tends to be somewhat
-more complicated as all the required pieces aren't shipped in the same package.
-Make sure to grab both the source and the Runtime Libraries, and some reports
-suggest that you might have to fiddle with paths to get `sdl.m4` working. But
-really, there's not much advantage over just using `brew`.
+```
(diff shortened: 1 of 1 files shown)


jangler:
Well, check out https://github.com/schismtracker/schismtracker/blob/master/.travis.yml -- this is the script used for automated builds. It uses the same build process for Linux and OS X; the only difference is that OS X installs SDL via brew before building.

THANK YOU for finding out that there was a .app in the sys/ directory all along. I'll take a closer look at this shortly.

kaneel:
there's probably a way to upload the build so for macOS it automatically move the build to the .app.

esaruoho:
@kaneel sure, one could start by copying schism_tracker.app directly to /Applications, create a folder within the Package Contents  and then copy the newly compiled  `schismtracker` inside the app. 

However - I decided to just go the route of first make it work one way, then make it worky work work.

@jangler I tried to look at linux + osx instructions for travis.yml but found nothing in the schismtracker docs. undocumented feature?

jangler:
Making a .app in Travis would go in the `after-success:` section. Conditionally (i.e. if the OS is "osx") placing the binary in the directory and zipping that should be easy, but I think we also need to include the .so dylib for SDL. Which is probably accessible somewhere in the build environment, since we did `brew install` it.

@esaruoho I don't understand what you mean about an undocumented feature. All the build steps used in travis.yml are already documented in the Linux build instructions.

snerk4000:
are there any builds of this uploaded anywhere that work on osx just out of the box?  i'd love to try this but installing 10 things and a whole homebrew package just to try it is stupid, i'm not a programmer, i'm never going to compile anything else, i'll just have to clean it all out again later.

Issue #138, opened 2018-01-13
.xi instrument bug involving envelope loops


smckean:
Actually more of a proposed fix. When Schism Tracker loads any .xi instrument containing volume/panning envelope loops it currently mimics original IT-style behavior: The ending node of the loop is included in the playback rendering on each pass. Testing the behavior of Milkytracker for original XM support, .xi's get loaded whereby the end-loop node is left out of rendering. This creates a difference of one tick per envelope loop between XM/XI files being loaded in FT2/Milky and IT/Schism (I tested IT 2.14 playback in Dosbox as well). This difference is usually not very noticeable, but there are valid XM's that make beautiful use of precise timing in these loops (one good example is Mellow-D's "vb.xm" -> https://api.modarchive.org/downloads.php?moduleid=59444#vb.xm ).

After fiddling with the .xi loading behavior in fmt/xm.c, I found the "problem" could be fixed by shifting the end-loop node back one tick during the file read, as well as any other nodes running up too close behind it. Not exactly optimal, but it works for my purposes nonetheless. I have the patch ready for a commit, but I guess the issue is I'm not too clear on how "origin-precise" Schism Tracker aims to be. I know there are some subtle differences between ST and IT, but also that Schism was built on some heavily-modified libmodplug code in order to provide a very similar look-and-feel experience. I must confess though that I do enjoy pulling up XM files occaisonally in Schism to get some relative exposure.

jangler:
Technically Schism does not support playback of XM files; but it does support loading them like IT files and "translating" them as accurately as is practical. Usually Schism does not aim to be bug-compatible with IT, so if this can be fixed cleanly then I'm all for it, but it sounds like that may not be the case.

So, using the fix you've made, if you load an XI without an envelope loop, and the same XI with an envelope loop, the nodes would be loaded differently into Schism? Wouldn't this cause the opposite problem when the loop is bypassed? I know that's less of an issue, but I'd rather not trade one inaccuracy for another. Or am I misunderstanding something?

I am also curious: does Milky do any conversion when loading IT envelopes?

smckean:
The way I proposed it would affect the precision of the IT envelope in translation. I could instead make it more accurate by creating an extra node one tick behind the end-of-loop node and moving the end there, leaving the original end-node where it was defined.

To be more specific there'd be three test cases at load time:
1) If the loop begins and ends on the same tick (loop length 0 ticks), disable the loop. Same thing happens in FT2, the loop marks are visible but not used.
[...]

PR #145, opened 2018-01-28
Fix envelope loop bug when loading XI instruments from XM files.

smckean:


vcaputo:
Could you put an explanation of what's wrong with the existing code and how this fixes it in the commiit message?

jangler:
I think it's a fix for #138. Not sure why OP closed it. I'm going to reopen it and look at this later.

fmt/xm.c
```diff
@@ -389,6 +389,8 @@ static int load_xm_instruments(song_t *song, struct xm_file_header *hdr, slurp_t
 	int n, ni, ns;
 	int abssamp = 1; // "real" sample
 	int32_t ihdr, shdr; // instrument/sample header size (yes these should be signed)
+	int l_start, l_end; // envelope loop nodes
```
jangler: What's the purpose of these two variables?

jangler: Oh, I see you're using `l_end` a lot as shorthand for `ins->vol_env.loop_end`, but only using `l_start` once per block. FWIW I'd rather `l_end` be a pointer to `ins->vol_env.loop_end` so that they're *actually* the same thing and don't need to be explicitly synchronized.

fmt/xm.c
```diff
@@ -576,6 +613,38 @@ static int load_xm_instruments(song_t *song, struct xm_file_header *hdr, slurp_t
 			ins->flags |= ENV_VOLUME | ENV_VOLSUSTAIN;
 		}
 
+		// panning loop endpoint fix, similar to volume fix above
+		if ((ins->flags & ENV_PANNING) && (ins->flags & ENV_PANLOOP)) {
+			l_start = ins->pan_env.loop_start;
+			l_end = ins->pan_env.loop_end;
+			if (l_start == l_end) {
+				ins->flags &= ~ENV_PANLOOP;
+			} else {
```
jangler: Correct me if I'm wrong, but the following block of code seems isomorphic with the corresponding block for volume envelopes above. It would save some LOC and some potential debugging if this block were "factored out" into a function that could be used for both.

fmt/xm.c
smckean: Good point. I originally figured I was going to use the values for reference, but it does turn out that a couple of pointers in this case would make the code more representative of the variable changes as well.

fmt/xm.c
smckean: Yes, I wrote it out this way to make the patch changes more "readable" in a redundant way. I'll be happy to create a regular function to reduce future bugs and bring down the line count.

sholwe:
Although I haven't examined the code whatsoever, this patch certainly has made Elwoods' Deadlock less enjoyable.  ;)

smckean:
Sholwe, could you please specify what you hear differently in the two builds? The patch affects a timing issue and shouldn't change the quality of the song as IT would already load it. I'm not hearing a noticeable difference when playing deadlock.

sholwe:
Hi @smckean,

For whatever reason, introducing this patch into my pull from git -head has introduced artifacts in the left channel on patterns 30 and 31.  I'll try to break it down farther when I get a chance.

smckean:
@sholwe, I see what you mean about the short artifacts. I checked the playback with IT and FT2, both in dosbox and found the original Impulse Tracker loads with these artifacts as well, unfortunately. I wasn't able to test against a prior Schism build, but I did notice the probable cause was an instrument switch without a note event in the patterns. Perhaps this fix is revealing a buried bug in IT's behavior? Maybe it needs a separate fix.

sholwe:
@smckean Oddly, I must have never really noticed it except with players which have issues; I generally used libMikMod (Yeah, I know) when it's been running in the background prior.. but I only really noticed after this patch.  PsychosomatiXM, perhaps?

Cheers.  Sorry I wasn't able to dig down to the pattern and isolate it farther; it's been on my "to do" for a couple days now.  Thanks for your attention!

sagamusix:
> Perhaps this fix is revealing a buried bug in IT's behavior? Maybe it needs a separate fix.

IT and FT2 simply have different ideas of how certain things work (such as instrument swapping in this case). There is no way to "fix" this in an IT-centric tracker like Schism, and it most definitely is not a bug. It's not possible to have 100% accurate XM playback in IT/Schism.

vcaputo:
@sagamusix It may just require teaching the engine some new modes that the loaders configure per-format, perhaps even with a dialog asking the user at load time which tracker's behavior to replicate.

sagamusix:
If you do that, you essentially get the original ModPlug/OpenMPT code back that was intentionally trimmed down to remove everything that's not IT. That is, it's exactly the opposite of what Schism wanted to be.
Static fixes at load-time for imported file formats are completely okay (Schism already does that), but that's not really possible in corner cases like this one (at least not without a lot of complicated work).

vcaputo:
Ah, I don't know the history.  It just seemed like an obvious path to making it work, but yeah it does add complexity.

jangler:
So I'm not really clear on this—is there some quirk *besides* the FT2 vs. IT instrument change behavior that is causing problems with XM import? And if so, is it a product of this PR specifically?

sagamusix:
I don't think so - DEADLOCK.XM has these issues already in the latest stable version.

jangler:
OK, then I'm going to open a separate issue for that and then merge this PR once I have a chance to look at the updated changes.

jangler:
The `uint8_t` casting/rounding is ugly, but… otherwise LGTM.

Issue #149, opened 2018-02-24
Segfault listing MIDI file in module browser


polyzium:
![peek 2018-02-25 01-15](https://user-images.githubusercontent.com/22579935/36635617-66d87236-19c9-11e8-8c8b-6f9b91c8ca1f.gif)

jangler:
Can you isolate what file is causing the problem?

vcaputo:
Hmm, I can't reproduce that here using the 8g_crim.xm I found via google... 
```
$ md5sum Downloads/8g_crim.xm 
aaee9d04b35ab5f22bbc804d227f7b08  Downloads/8g_crim.xm
```

But I do see a warning from the xm loader about it having too many samples.

edit:

Oh, I see, this is just while browsing and not loading the highlighted file in the browser.  Nevermind!

polyzium:
I don't know what causes this problem, is there any log file or something?
EDIT: It ocasionally can print `malloc(): memory corruption` into the terminal + crash with abort signal instead of segfault:
```
malloc(): memory corruption
[1]    3026 abort (core dumped)  schismtracker
```
@vcaputo This is clearly not caused by 8g_crim.xm

jangler:
There's no log file that I'm aware of, no. Schism reads file headers in order to determine the color and title to display each module with in the browser; Schism is likely screwing up on one of those files, for whatever reason.

The way to identify the problem file is to get Schism to crash when loading a directory with only that file. The quickest way to do this would be to use a binary search (put half the files in one directory and half in another, then see which directory Schism crashes on, then split that directory half and half, etc., until you have a directory with only one file), but I'd probably also check any files in more obscure formats first.

jangler:
(You could also get an idea of what is happening using a debugger, but I am not good at using debuggers, so I can't provide advice on that front.)

sagamusix:
If you're not experienced with using debuggers, you can start by copying the files from your Downloads folder one by one to a new folder and try opening that new folder in Schism instead. That way, you can identify the culprit more easily, which we then can use to have a deeper look.

vcaputo:
You could simply zip up the directory containing this problematic file and put it somewhere we can then download it and debug the problem ourselves.

Alternatively, just turn on core dumps with `ulimit -c unlimited` prior to running schism, then reproduce the segfault.  After that run `gdb /path/to/schismtracker /path/to/core/file` and type `bt full`, paste the output in this issue.  If we're lucky, it will be illuminating.

[...]

PR #150, opened 2018-02-25
fmt/mid: fix message_left underflow

vcaputo:
message_left can reach zero, don't assume it can always have 1
subtracted.

This code is fragile and frankly a mess, but this minor change seems to
prevent a crash with at least one probably corrupted midi file.

See https://github.com/schismtracker/schismtracker/issues/149

### DIFF
--- a/fmt/mid.c
+++ b/fmt/mid.c
@@ -331,7 +331,7 @@ int fmt_mid_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
 					case 0x5: // lyric
 					case 0x6: // marker
 					case 0x7: // cue point
-						y = MIN(vlen, message_left - 1);
+						y = MIN(vlen, message_left ? message_left - 1 : 0);
 						slurp_read(fp, message_cur, y);
 						if (x == 3 && y && !song->title[0]) {
 							strncpy(song->title, message_cur, MIN(y, 25));


sagamusix:
Good catch, but there's nothing corrupted about a MIDI file containing strings of length 0. It's just an oversight in the loader.

vcaputo:
@sagamusix it's not a message length of zero, it's message_left reaching zero because there's too much stuff accumulating into the song->message.

The file which brought this to our attention looks corrupt to me, the title is gibberish and the song doesn't really play anything.  Maybe it isn't but something else isn't right.

Either way this specific thing was obviously wrong and that file was tripping over it.

sagamusix:
Ah, yup, sorry, I confused it with vlen, which also happens to be 0 in this file for some marker.

vcaputo:
That function needs cleaning up, it's non-trivial to make sense of if you've never looked at it before.  I didn't take the time to verify if song->message must be null-terminated or if MESSAGE_MAX serves as a guard, and the fmt-specific loader doesn't zero it out first but maybe that happens earlier in the call chain.  That's kind of a blurry mess with poorly defined layer boundaries, at least when viewed from just fmt/mid.c.

sagamusix:
String handling in C, always a pleasure ;)

vcaputo:
There's that, but there's also just the lack of interfaces being employed. But maybe that's just the preferred coding style in schismtracker.

I'm more accustomed to PIMPL-style C code, where you'd probably put the song behind an opaque pointer then the loader has to go through an interface to change things like the song message.  The loader would then be responsible for growing a buffer to accomodate its message text, then hand it to the song interface.

Just better defined boundaries emerge, and it makes reading a specific component a lot less ambiguous since its local concerns are better isolated.

sagamusix:
I think the lack of such an interface with the song message here can be explained simply by the fact that in all other formats, the song message is a single text block that can be read in one go. Only in the MID importer, Schism concatenates all the small MIDI cue texts for a song message.

PR #152, opened 2018-03-05
video: drop unnecessary _video_preinit()

vcaputo:
The video struct is static, so it's already zeroed.

The size of the allocated arrays is known at compile time, so
it's unnecessary to allocate them at runtime.

The whole preinit is in a guard to prevent occurring multiple times,
so this simplification seems obviously safe.

### DIFF
--- a/schism/video.c
+++ b/schism/video.c
@@ -202,9 +202,9 @@ struct video_cf {
 	SDL_Surface * surface;
 	SDL_Overlay * overlay;
 	/* to convert 32-bit color to 24-bit color */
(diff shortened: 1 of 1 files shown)


vcaputo:
I'm poking at SDL2 porting.  schism/video.c is a hot mess so in the mean time I'll probably be posting a number of cleanup PRs for this listing which should result in a much simpler and easier to review SDL2 port PR should that materialize.

jangler:
Thank you for making this code base incrementally more sane.

PR #153, opened 2018-03-06
main: preserve kk across event loop iterations

vcaputo:
Commit 6a8357 moved `struct key_event kk` into the event loop body
which had the unintended side effect of zeroing the .s[xy] members.

These are supposed to hold the start character coordinates for mouse
clicks, needed for drag operations like clipboard text select.

This commit moves kk back outside the event loop, fixing text select.

To reproduce the bug this fixes, go into the message or pattern editor
and try selecting stuff with the mouse.

### DIFF
--- a/schism/main.c
+++ b/schism/main.c
@@ -533,6 +533,19 @@ static void _do_clipboard_paste_op(SDL_Event *e)
 	_synthetic_paste((const char *)e->user.data1);
 }
 
+static void key_event_reset(struct key_event *kk, int start_x, int start_y)
+{
+	memset(kk, 0, sizeof(*kk));
+
+	kk->midi_volume = -1;
+	kk->midi_note = -1;
+	/* X/Y resolution */
+	kk->rx = NATIVE_SCREEN_WIDTH / 80;
+	kk->ry = NATIVE_SCREEN_HEIGHT / 50;
+	/* preserve the start position */
+	kk->sx = start_x;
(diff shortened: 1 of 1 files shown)


vcaputo:
This appears to have gone unnoticed since 2010.  I just wanted to exercise the copy and paste stuff with SDL2 and found I couldn't manage to select anything with the mouse...

jangler:
Hah, what on earth? How did this go unnoticed for so long? I figured at least *someone* would try to use the mouse for this.

Is it possible that `.midi_volume` and `.midi_note` would still need to be set to -1 at the beginning of the loop? I wish the commit message for 6a83577 were actually descriptive…

jangler:
And likewise the other fields that would be zeroed, for that matter.

vcaputo:
@jangler force-pushed an updated version which resets everything except kk.s[xy] in a way that preserves the "everything else is zeroed" assumption, including any future added members to `struct key_event`.

I was falling asleep last night when sorting this out, but wanted to get something up before morning as I've noticed you're often active in the mornings.  The loose ends should be all tied up now.

jangler:
LGTM. Again, surprised this went unaddressed for so long.

vcaputo:
It's especially amusing if you consider the pile of OS-specific pain in `schism/clippy.c` to make this work and nobody even noticed it's broken for 8 years.

PR #154, opened 2018-03-06
main: cease accessing event.key in mouse events

vcaputo:
Use SDL_GetModState() to get the modkey state, the key member
is invalid for the mouse event types, accessing it is a bug.


vcaputo:
This bug wreaks havoc on clippy selections in SDL2 but seems to go unnoticed in SDL1 :man_shrugging:

PR #155, opened 2018-03-07
Fixup X11 selection handling

vcaputo:
Failed XSelectionRequestEvent scenarios are being replied to with a different target type than requested.  This confuses requestors so they ignore the response.  XA_CUT_BUFFER0 should also be the cut buffer used for the primary selection.

With these changes I can now paste directly from schism to xterm.  Without them I cannot, with only the first commit to use XA_CUT_BUFFER0 I can sort of paste, but it takes a very long time.


jangler:
Oh yeah, I forgot this was broken. PR works for me!

PR #176, opened 2018-09-20
Set MACOSX_DEPLOYMENT_TARGET to 10.11

### LINKED ISSUE #172: macOS version requirement
The latest build of Schism Tracker for macOS claims to require macOS 10.13 or later. I'm not sure why this is, because in Travis we specify macOS image from `beta-xcode6.1` up to `xcode8` which according to [Travis documentation](https://docs.travis-ci.com/user/reference/osx/#os-x-version) should mean we compile on nothing newer than macOS 10.11.

screen shot 2018-08-16 at 11 10 43

However the case may be, [the solution might be](https://stackoverflow.com/a/39688949/61818) to add the `-mmacosx-version-min=10.11` flag to `gcc` when compiling on macOS. Is this going to cause any problems when compiling on other platforms or will the flag be ignored then?

The strangest thing about this is that I'm unable to reproduce the problem locally. Every build I do of Schism Tracker on my Mac works.

asbjornu:
This PR sets the `MACOSX_DEPLOYMENT_TARGET` environment variable to `10.11` in Travis to hopefully fix #172.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ env:
   global:
   - ARTIFACT_FILE="$TRAVIS_BUILD_DIR/build/schismtracker-$TRAVIS_TAG-$TRAVIS_OS_NAME.tar.gz"
   - YEAR=$(date +'%Y')
+  - MACOSX_DEPLOYMENT_TARGET=10.11
 matrix:
   include:
   - os: linux


asbjornu:
@jangler: Awesome. Would it be possible to tag a new release so I can test if this works? 😃

jangler:
Here's one: https://github.com/schismtracker/schismtracker/releases/tag/macosx-deploy-test

asbjornu:
@jangler: Awesome, it works!

jangler:
Thanks for the contributions, as always, in areas where I'd be totally lost otherwise.

asbjornu:
No problem, I’m just scratching my own itches here. :smile:

PR #183, opened 2018-12-11
Update disko.h

alexmyczko:
GNU/Hurd support (yes it doesn't have sound drivers), https://buildd.debian.org/status/package.php?p=schism

### DIFF
--- a/include/disko.h
+++ b/include/disko.h
@@ -25,6 +25,9 @@
 
 #include <sys/types.h>
 
+#ifndef PATH_MAX                                                                                                                                          
+#define PATH_MAX 4096                                                                                                                                     
+#endif
 
 typedef struct disko disko_t;
 struct disko {


jangler:
Wouldn't `#ifdef __GNU__` be a better condition?

alexmyczko:
i dont know but the pr worked i tried: https://buildd.debian.org/status/package.php?p=schism

esaruoho:
@jangler what's happening with this PR? does it need to be merged in, is it important, @alexmyczko what's this actually properly doing? could it not be fixed/tweaked further and then merged in?

vcaputo:
`#include <limits.h>` is appropriate before using `PATH_MAX` [1].

@alexmyczko are you able to change this to instead include the header and verify that it fixes the HURD build?

Otherwise any of us can add the include, but I don't think any of us will be attempting to build on HURD for verification.

[1] https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html

alexmyczko:
The point is GNU/Hurd doesn't have limits.h: https://www.gnu.org/software/hurd/community/gsoc/project_ideas/maxpath.html

The patch fixes it, however it's not really useful since GNU/Hurd doesn't have sound drivers. So feel free to add the patch, or just not add it :)

Happy new year!

jangler:
The "guidelines" page linked to by the page you linked states that "\_\_GNU__ is the correct define for hurd specific code". I just think that this PR should use that define to make sure it is specifically addressing the problem in the context of Hurd and not other cases where PATH_MAX is for some reason undefined. Using a \_\_GNU__ define also documents *why* PATH_MAX is being set to 4096 in the conditional.

alexmyczko:
I agree with you, feel free to do it right and close...

PR #196, opened 2019-07-09
Preserve shared samples on "Delete instrument & all related samples"

ccollins476ad:
New instrument command: delete-with-preserve

This commit adds a new command to the instrument page:
```
Delete instrument & all related unused samples
```
This command is bound to Alt-Shift-D.

When executed, this command deletes the currently selected instrument and all related samples that are not used by any other instrument.  This is similar to Alt-D, except it does not delete samples used by other instruments.

Example: If instruments X and Y both use sample S, you can delete one instrument without breaking the other.

### DIFF
--- a/fmt/xi.c
+++ b/fmt/xi.c
@@ -128,7 +128,7 @@ int fmt_xi_load_instrument(const uint8_t *data, size_t length, int slot)
 
 	end = data + length;
 
-	song_delete_instrument(slot);
+	song_delete_instrument(slot, 0);
 
 	g = instrument_loader_init(&ii, slot);
 	memcpy(g->name, xi->name, 22);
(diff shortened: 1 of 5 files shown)


jangler:
I like the idea, but I wouldn't want to make a "breaking" change to the behavior of the existing command, especially since this is the way that Impulse Tracker works. I think it would be better to associate this with a new key combination, maybe Alt-Shift-D?

ccollins476ad:
Thanks @jangler, that makes sense.  I'll change it to Alt-Shift-D.

ccollins476ad:
BTW - I have updated this PR such that the new behavior is mapped to Alt-Shift-D.  Alt-D retains the existing behavior.

jangler:
Great, thanks.

PR #250, opened 2021-04-16
Retrohun/savemod

retrohun:
Adds capability of saving 4-channel modules, de facto standard "M.K." magic bigendian dw at filepos 1080. There are still plenty of online amiga protracker demoscene contests around!


fmt/mod.c
```diff
@@ -396,3 +467,192 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+		for(i = 0; i < 30; ++i) mod_sampleheader[i] = 0;
+		if(n <= nsmp) {
+			if(song->samples[n].global_volume != 64)
+				warn |= 1 << WARN_SAMPLEVOL;
+			if((song->samples[n].flags & (CHN_LOOP | CHN_PINGPONGLOOP)) == (CHN_LOOP | CHN_PINGPONGLOOP) || (song->samples[n].flags & CHN_SUSTAINLOOP))
+				warn |= 1 << WARN_LOOPS;
+			if(song->samples[n].vib_depth != 0)
+				warn |= 1 << WARN_SAMPLEVIB;
+			memcpy(mod_sampleheader, song->samples[n].name, 22); // sample name
+			mod_sampleheader[22] = song->samples[n].length >> 9; // sample 11th word MSB length/2
```
sagamusix: This should limit the sample length to 0x1FFFE samples before shifting. Otherwise longer samples will be truncated. Loop start / length should be adjusted accordingly if they are beyond the sample length.

fmt/mod.c
```diff
@@ -396,3 +467,192 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+		}
+		disko_write(fp, mod_pattern, 1024);
+	}
+
+	// TODO : store samples, truncate to even length per sample //Signed 8-bit mono sample data.
+	// Now writing sample data
+	for (n = 0; (n < nsmp) && (n < 31); ++n) {
+		song_sample_t *smp = song->samples + (n + 1);
+		if (smp->data)
+			csf_write_sample(fp, smp, RS_PCM8S); // last argument is a compound flag: PCMS,8,M,LE (endianness is obsolete)
```
sagamusix: > endianness is obsolete

obsolete is the wrong word here, it simply doesn't matter for 1-byte values.

fmt/mod.c
```diff
@@ -396,3 +467,192 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+				}
+				mod_pattern[joutpos + 2] |= mod_fx & 0x0f;
+				mod_pattern[joutpos + 3] = mod_fx_val;
+				joutpos += 4;
+			}
+		}
+		disko_write(fp, mod_pattern, 1024);
+	}
+
+	// TODO : store samples, truncate to even length per sample //Signed 8-bit mono sample data.
```
sagamusix: This TODO absolutely needs to be addressed before merging, there are already enough broken MODs in the wild, and giving people the easy possibility of writing more of them is not helpful.

The way this could be addressed is adding a new `length` parameter to `csf_write_sample`. Existing call sites of the function could simply pass uint32_max as a length, and the function itself would then take `MIN(length, sample->length)` as the length to write.

sagamusix: For non-looped samples you should also make sure that the first two bytes are 0 (you can overwrite them after the `csf_write_sample` call), or they will sound broken in ProTracker and other players that implement one-shot loops (technically every MOD sample is looped, but if the first point of the loop is 0, ProTracker first plays the complete sample and then just the looping section. This is why you will typically find loop start 0, loop length 2 for non-looped samples in MOD).

fmt/mod.c
```diff
@@ -396,3 +467,192 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+				warn |= 1 << WARN_SAMPLEVIB;
+			memcpy(mod_sampleheader, song->samples[n].name, 22); // sample name
+			mod_sampleheader[22] = song->samples[n].length >> 9; // sample 11th word MSB length/2
+			mod_sampleheader[23] = song->samples[n].length >> 1; // sample 11th word LSB length/2
+			for(j = 15; j && (finetune_table[j] > song->samples[n].c5speed); --j)
+				if(((song->samples[n].c5speed) > 10000) && (j == 8))
+					break; // determine from finetune_table entry
+			mod_sampleheader[24] = (j ^ 8) & 0x0f; // sample 24th byte finetune value
+			mod_sampleheader[25] = (song->samples[n].volume + 1) / 4; // sample 25th byte sample volume value 0..64 scaled
+			if( song->samples[n].flags & CHN_LOOP ) {
```
sagamusix: In the currently missing `else` block, loop length should be set to 1, otherwise ProTracker will have issues (see comment below about one-shot loops).

fmt/mod.c
```diff
@@ -396,3 +467,192 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+	for(maxpat = i = 0; (i < nord) && (i < 128); ++i) {
+		mod_orders[i] = song->orderlist[i];
+		if(maxpat < mod_orders[i]) maxpat = mod_orders[i];
+	}
+	for(; i < 128; ++i)
+		mod_orders[i] = 0;
+
+	disko_write(fp, mod_orders, 128);
+
+	disko_write(fp, valid_tags[0][0], 4);
```
sagamusix: This should write M!K! if the highest pattern index is > 64.

fmt/mod.c
```diff
@@ -396,3 +467,192 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+				mod_pattern[joutpos] = ((m->instrument) & 0x10) | (period >> 8);
+				mod_pattern[joutpos + 1] = period & 0xff;
+				mod_pattern[joutpos + 2] = (m->instrument & 0xf) << 4;
+				mod_fx = 0; mod_fx_val = m->param;
+				if(m->voleffect == VOLFX_VOLUME) {
+					mod_fx = 0x0c;
+					mod_fx_val = m->volparam;
+				} else if (m->voleffect == VOLFX_NONE) {
+					switch(m->effect) {
+						case 0: mod_fx_val = 0; break;
```
sagamusix:
```suggestion
						case FX_NONE: mod_fx_val = 0; break;
```

sagamusix (commented):
Thanks for your contribution. There are a couple of comments that absolutely need to be addressed to avoid leaking malformed MOD files into the wild.

fmt/mod.c
retrohun: csf_write_sample() updated with truncate-and-mask argument, which is ignored when it is uint32_max

fmt/mod.c
retrohun: done

fmt/mod.c
retrohun: done in change @8f2a1fe

fmt/mod.c
retrohun: done

fmt/mod.c
retrohun: done applying >= 64 as barrier

fmt/mod.c
retrohun: done.

fmt/mod.c
```diff
@@ -396,3 +470,209 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+	}
+	for(; i < 128; ++i)
+		mod_orders[i] = 0;
+
+	disko_write(fp, mod_orders, 128);
+
+	if(maxpat < 64)
+		disko_write(fp, valid_tags[0][0], 4);
+	else {
+		warn |= 1 << WARN_MORETHAN64PATTERNS;
```
sagamusix: No need to warn here because the correct magic bytes are written now

player/csndfile.c
```diff
@@ -538,9 +538,11 @@ void csf_loop_pattern(song_t *csf, int pat, int row)
 #define SF_FAIL(name, n) \
 	({ log_appendf(4, "%s: internal error: unsupported %s %d", __FUNCTION__, name, n); return 0; })
 
-uint32_t csf_write_sample(disko_t *fp, song_sample_t *sample, uint32_t flags)
+uint32_t csf_write_sample(disko_t *fp, song_sample_t *sample, uint32_t flags, uint32_t maxlengthmask)
 {
 	uint32_t pos, len = sample->length;
+	if(maxlengthmask != UINT32_MAX)
```
sagamusix, line 544: I think it would make more sense if this parameter wasn't a mask but simply an upper limit. That would also be closer to the original ModPlug source this is taken from. You could then pass `MAX(length & ~1, 0xFFFFE)` to the function for MOD.

fmt/mod.c
```diff
@@ -396,3 +470,209 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+				mod_sampleheader[23] = 0xFF;
+				warn |= 1 << WARN_LONGSAMPLES;
+			}
+			for(j = 15; j && (finetune_table[j] > song->samples[n].c5speed); --j)
+				if(((song->samples[n].c5speed) > 10000) && (j == 8))
+					break; // determine from finetune_table entry
+			mod_sampleheader[24] = (j ^ 8) & 0x0f; // sample 24th byte finetune value
+			mod_sampleheader[25] = (song->samples[n].volume + 1) / 4; // sample 25th byte sample volume value 0..64 scaled
+			if( song->samples[n].flags & CHN_LOOP ) {
+				mod_sampleheader[26] = song->samples[n].loop_start >> 9;// loop start MSB /2
```
sagamusix, line 524: Please limit loop start and loop end accordingly when the sample length is truncated. ProTracker will do funky things otherwise (it will play parts of the following sample(s)).

fmt/mod.c
sagamusix: You'll still need to zero out the first two bytes of the sample if it's not looped. This doesn't need to be part of the csf_write_sample function, you can for example `fseek` back to the beginning of the sample and overwrite the samples afterwards.

player/csndfile.c
retrohun: This parameter has only one job - limit/truncate mod samples on passed values other than default. In the future, it might be split, but till then I think that it is correct. If it is not a cardinal showstopper, I would keep it in its current form.

fmt/mod.c
retrohun: What do you mean under "zeroing out"? Insertion or overwriting? How do you define "Zero-level": aling to/make them equal to last nibble/byte value of sample, set to signed zeroes, set to unsigned zeroes, repeat first byte etc... Also consider that preparing samples should belong to the end users as well as assuming some Paula knowledge. But IRL, I strongly agree with you ;-)

sagamusix: The first two bytes of the sample data should be set to 0. In order to avoid growing the sample each time the file is re-saved, trackers typically just overwrite the first two bytes instead of inserting extra bytes.

> Also consider that preparing samples should belong to the end users as well as assuming some Paula knowledge

No, it absolutely shouldn't. 99% of users don't know about this and are still perfectly capable of making a ProTracker-compatible MOD otherwise. As SchismTracker doesn't even have an easy-to-use sample editor that could help them fixing this, it's absolutely vital that the software automatically does it for them.  I have attached an example file to demonstrate the problem if the tracker doesn't care of zeroing the samples. Load it into ProTracker 2.3d in (Win)UAE and you will notice a high-pitched beep when the sample is finished playing. Most players don't emulate this feature so even if a user tested their MOD file in 10 different MOD players they would still be unlikely to figure out that something is technically wrong with their file.

[oneshot-loop.zip](https://github.com/schismtracker/schismtracker/files/6359823/oneshot-loop.zip)

retrohun: De facto done+pushed your recommended changes. De jure needs testing ;-)

retrohun: Seems to be okay, rebased to schismtracker:master

fmt/mod.c
```diff
@@ -396,3 +468,211 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+				if(((song->samples[n].c5speed) > 10000) && (j == 8))
+					break; // determine from finetune_table entry
+			mod_sampleheader[24] = (j ^ 8) & 0x0f; // sample 24th byte finetune value
+			mod_sampleheader[25] = (song->samples[n].volume + 1) / 4; // sample 25th byte sample volume value 0..64 scaled
+			if( (song->samples[n].flags & CHN_LOOP) && (song->samples[n].loop_start < song->samples[n].loop_end) && (song->samples[n].loop_end <= MIN(song->samples[n].length, 0x1FFFE)) ) {
+				mod_sampleheader[26] = song->samples[n].loop_start >> 9;// loop start MSB /2
+				mod_sampleheader[27] = song->samples[n].loop_start >> 1;// loop start LSB /2
+				mod_sampleheader[28] = (song->samples[n].loop_end - song->samples[n].loop_start) >> 9;// loop length MSB /2
+				mod_sampleheader[29] = (song->samples[n].loop_end - song->samples[n].loop_start) >> 1;// loop length LSB /2
+			}
```
sagamusix, line 533: the `else` block seems to have gone missing here (`mod_sampleheader[29] = 1;` in particular is missing)

retrohun: There was a small refactoring commit, which initialized the *mod_sampleheader with (((unsigned int)(floor(abs(index)/29.0)))) around line 506.

sagamusix: Can you please avoid this kind of code golf? It may look cool to you but in the end makes maintenance of the code much more complicated for no good reason.

retrohun: done.

sagamusix: Thanks!

fmt/mod.c
```diff
@@ -396,3 +468,216 @@ int fmt_mod15_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
...
+						case FX_TONEPORTAMENTO: mod_fx = 3; break;
+						case FX_VIBRATO: mod_fx = 4; break;
+						case FX_TONEPORTAVOL: mod_fx = 5; break;
+						case FX_VIBRATOVOL: mod_fx = 6; break;
+						case FX_TREMOLO: mod_fx = 7; break;
+						case FX_PANNING: mod_fx = 8; break;
+						case FX_OFFSET: mod_fx = 9; break;
+						case FX_VOLUMESLIDE:
+							mod_fx = 0x0a;
+							if ((mod_fx_val & 0xf0) == 0xf0) { // fine volslide down!
```
sagamusix: Note that DF0 and D0F are currently not handled correctly by this conversion: They are converted to EA0 and EB0, but should be converted to AF0 and A0F.

retrohun: I will push the fix in a moment.

retrohun: Done.

fmt/mod.c
```diff
@@ -75,6 +82,71 @@ static const char *valid_tags[][2] = {
 	{NULL, NULL}
 };
 
+enum {
+	WARN_MAXPATTERNS,
```
sagamusix: I think we're almost ready to go. Here you can remove unused enum values.

retrohun: Hold my water...

retrohun: Done.

retrohun:
Shall I rebase it before merging?

sagamusix:
Please rebase, and if you want to preserve meaningful commits, rewrite them as needed.

@jangler do you intend to have a look at this as well?

jangler:
I'm not a mod expert but I don't see any overt problems with the code, and it seems to work OK. Green light from me.

retrohun:
> Please rebase, and if you want to preserve meaningful commits, rewrite them as needed.
> 
> @jangler do you intend to have a look at this as well?

Done.

jangler:
Thanks for your work, @retrohun. I will try to get AppVeyor working again so this can actually be available in Windows builds.

PR #256, opened 2021-05-08
Fix Haiku's save config directory

Begasus:
Haiku doesn't save files in ~/. directory, best to save them to ~/config/settings

### DIFF
--- a/schism/config.c
+++ b/schism/config.c
@@ -52,6 +52,8 @@ int cfg_video_width, cfg_video_height;
 # define DOT_SCHISM "Library/Application Support/Schism Tracker"
 #elif defined(GEKKO)
 # define DOT_SCHISM "."
+#elif defined(__HAIKU__)
+# define DOT_SCHISM "config/settings/schism"
 #else
 # define DOT_SCHISM ".schism"
 #endif


sagamusix:
Do you think you could look into providing either a fallback or an automatic migration for existing users, or at least showing a warning if settings are found in the old location? Seems a bit harsh that right now, people would lose all their settings when upgrading and they wouldn't know why.

esaruoho:
@mrpapersonic what's going on with this one? :)

sagamusix:
What's going on is: The requested fallback mechanism is still missing.

esaruoho:
@Begasus hey could you provide the fallback or automatic migration for existing users, or show a warning if the settings are found in the old location? losing settings is not a good look.

Begasus:
Is there any mechanisme atm available for other OS's? Could have a look first there (doing other projects atm, but could have a look around).

mrpapersonic:
Not in the current code, no.

sagamusix:
You could have a look at the portable mode added recently (#449), as this also causes the config dir to change at runtime.

esaruoho:
> You could have a look at the portable mode added recently (#449), as this also causes the config dir to change at runtime.

@Begasus do you have the bandwidth to look at #449? Seems like this would be a good way to update this 3 year old PR?

Begasus:
Checked out latest release that includes #449 , looks to be working fine as fallback (as far as my skills go) :)

esaruoho:
@Begasus @mrpapersonic so.. what's going on with this PR , please, does it need to be merged it, can it be merged in.. should it be merged in?

mrpapersonic:
As stated before by saga musix, there are issues here that still need to be addressed. I don't care about Haiku, so it probably won't be me finishing this PR :)

Begasus:
Haiku is in the process on releasing a new beta, so we're a bit busy atm with getting some things up to speed, will report back later (if you can wait, if not it can be closed and we'll have it patched locally if need be) :)

mrpapersonic:
I've implemented the fallback Saga Musix requested in 9c8cc438ee97e6903c2dc50a1d7485ad478e2b71 as part of moving to the not-so-new-anymore XDG base directory specifications.

PR #257, opened 2021-05-13
#64 fix alsa/oss midi setup conflict

### LINKED ISSUE #64: "ALSA: Resource temporarily unavailable" when selecting first MIDI port on a device
In the MIDI configuration screen. When I go down the list of devices, I can only activate the second or later instance of each device. I.e. on my firewire interface with two MIDI out ports I can only activate the second port, the first one stays locked to "disabled." Using a simple 1-port USB dongle I can't enable it at all, it's stuck on "disabled" and nothing can be done. Pressing spacebar to activate the port generates an "ALSA: Resource temporarily unavailable" error in the message console.

More details & screenshot for explanation here (scroll down) :
http://chipmusic.org/forums/topic/18439/what-happened-to-schismtrackerorg/page/6/

The issue occurs for me on Fedora 21, Mint 17, Mint 18, and AVLinux 2016 (Debian), including on a freshly compiled version with no preexisting config file. I'm using the latest 2016.11.14 snapshot, compiled by myself, and it occurs on all of them. I built the exact same source tarball on my Mac (OSX 10.8) and it doesn't happen there. It was also present at least as far back as version 2016.07.31.

Happens regardless of whether the ALSA or Pulse driver is being used, but does NOT happen on the "dummy" driver. I haven't tried OSS yet.

Anyone have an idea how to fix this? I have a live act coming up early in January and I really want to incorporate Schism into it.

laamaa:
Closes #64

### DIFF
--- a/schism/midi-core.c
+++ b/schism/midi-core.c
@@ -303,10 +303,13 @@ static void _midi_engine_connect(void)
 #ifdef USE_NETWORK
 	ip_midi_setup();
 #endif
-#ifdef USE_OSS
+//Prefer ALSA MIDI over OSS, but do not enable both since ALSA's OSS emulation can cause conflicts
+#if defined(USE_ALSA) && defined(USE_OSS)
+	if (!alsa_midi_setup())
+		oss_midi_setup();
+#elif !defined(USE_ALSA) && defined(USE_OSS)
 	oss_midi_setup();
-#endif
-#ifdef USE_ALSA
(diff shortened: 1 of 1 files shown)


sagamusix:
Looks sensible to me, but as I'm not really using either Schism Tracker or Linux on the desktop, maybe @jangler can have a second look and merge it

jangler:
Looks right to me and still works on Arch, so I'll just trust that it also works on OSS systems like you said. :stuck_out_tongue:

Issue #268, opened 2021-07-20
OpenGL setting in windowed mode only shows screen in bottom left corner on a Retina display on Catalina


KungFuFurby:
This program suffers from the 'ole high DPI problem that I've been hearing about on various other Github repositories, and it affects me running Catalina since high DPI mode is the default, and I'm on a retina display, which I believe causes the high DPI problems to pop up in the first place in some circumstances.

This only affects OpenGL mode: it doesn't affect any other graphical modes, which scale correctly.

There seems to be a couple of solutions to this regarding disabling high DPI support:
* Set NSHighResolutionCapable to false in the info.plist. Requires an app bundle to be complied, though.
* On the Cocoa side of things, set wantsBestResolutionOpenGLSurface to NO. This program has Cocoa-side programming, but I don't think it is known to directly reference OpenGL yet.

The alternative is to somehow get the aspect ratio in question off of a retina display and factor that in on the C++ side.

Here's a screenshot of the problem in action right on startup:
<img width="752" alt="Schism Tracker Quarter Screen Size" src="https://user-images.githubusercontent.com/7094832/126385130-51154a43-4d6b-43c8-b12d-2e234a5a980a.png">

Note that I used a manually compiled binary: the commit I last compiled from is cfef6349ecb8188e0c0dd08db00486635480aeb7 .

jangler:
I would support a pull request adding the line `[self setWantsBestResolutionOpenGLSurface:YES];` to `sys/macosx/macosx-sdlmain.m` at an appropriate place (if it's that simple), but I don't have a way to test the result myself.

jangler:
Sorry, that should be `[self setWantsBestResolutionOpenGLSurface:NO];`, of course.

jangler:
If someone with macOS and a Retina display would like to test whether this is still an issue with the SDL2 build, I would appreciate it!

KungFuFurby:
No problem here.

PR #274, opened 2021-11-17
Set NSHighResolutionCapable to false in app bundle settings

KungFuFurby:
This merge request mentions #268. Due to only affecting app bundles, it does not close
the issue: instead one of these two need to be implemented:
- Set wantsBestResolutionOpenGLSurface to NO. Requires Cocoa-side programming.
- Somehow figure out how to grab the aspect ratio and factor that in to the
  scaling already done by the program.


KungFuFurby:
Schism Tracker 2021 Startup

Ta-da! Here's the startup screen now using the app bundle that was just compiled.

PR #277, opened 2021-11-19
Port to SDL2

vcaputo:
Just throwing this up in case it's of use, I don't currently have time to spend dragging this across the finish line.

/cc @jangler

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -59,13 +59,13 @@ dnl check endianness
 AC_C_BIGENDIAN
 
 dnl Check for SDL libs
-AC_CHECK_TOOL([SDL_CONFIG], [sdl-config])
+AC_CHECK_TOOL([SDL_CONFIG], [sdl2-config])
 AC_SUBST(SDL_CONFIG)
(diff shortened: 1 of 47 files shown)


vcaputo:
Took me a bit longer than expected to get around to dusting this off enough to be remotely presentable, feel free to use whatever is of value from it.

Again, I still think it might be worth just leaving things in SDL1.2 land and relying on a compatibility shim, if such a thing both exists already and has maintainers.  Schismtracker clearly lacks developer resources to maintain it, and there's plenty of sdl1.2 crap out there justifying a community maintained compat layer.

jangler:
Looks like there's an official one now at <https://github.com/libsdl-org/sdl12-compat>. It doesn't have a first release yet but it looks like one is in the works (just to give a ballpark on status, not that we'd need to wait for a release). I'll try it out with Schism.

jangler:
First impressions are that it seems OK on Linux but the colors are wrong in the SDL video mode.

Issue #270, opened 2021-10-10
Setting AUDIODEV= before running schismtracker has no effect


vcaputo:
I wanted to instruct schism to use my USB MBox when attached by simply telling it to via the SDL AUDIODEV variable.

After wasting way too much time trying to understand why ALSA was being uncooperative, it turns out schism is actively unsetting this variable @ https://github.com/schismtracker/schismtracker/blob/20210525/schism/audio_playback.c#L1241

By simply commenting out the _audio_set_env_vars() call @ https://github.com/schismtracker/schismtracker/blob/20210525/schism/audio_playback.c#L1280  I am finally able to get my `plughw:MBox,0,0` used properly via `AUDIODEV=plughw:MBox,0,0 ./schismtracker`.

Not clear what should be done to fix this, but it seems totally user-hostile to be throwing out user-specified SDL1.2 env vars

jangler:
Yeah, unsetting the env vars without any stated rationale is very weird. Schism does look like it has a way around this via the `--audio-driver` command-line option and equivalent `driver` config file option. Do those also work for you?

vcaputo:
It does appear I can work around this by specifying `--audio-driver=alsa:plughw:MBox,0,0`, and I suppose I should have been able to discover this via reading `configuration.md`.

But I think there's arguably a bug here in how schism is silently interfering with otherwise established SDL1.2 environment variables.

If schism insists on unsetting and mangling these, it'd be better to at least detect if they were set a priori and say something in stderr/message log about their being ignored+overridden.  Perhaps mentioning instead put what you'd normally put in AUDIODEV=, after `$driver:` in --audio-driver/config::Audio::driver.

Thanks for the quick response.

jangler:
Ran into this while testing the SDL2 port; it also ignores `SDL_AUDIODRIVER`, which isn't great either. If I had explicitly specified an `--audio-driver` or put a `driver=` line in the `[Audio]` section of my config file, that'd be one thing, but I didn't do those things, and overriding env vars with an implicit default is not something that should be happening.

mrpapersonic:
Ok, I've taken a shot at this. [This](https://github.com/mrpapersonic/schismtracker/actions/runs/2879256129) build checks if the vars are already defined or not. If they aren't, it tries sets the drivers that are listed in the config, and *then* tries the defaults.

PR #292, opened 2022-05-26
SDL2 Port

### LINKED ISSUE #33: SDL2 port (Mac OS X SDL 1.2 deprecated)
This might be related to #23 which then caused a crash, but after self-recompilation instead causes this warning to be dumped in the console:

```
2016-08-15 19:41:45.264 schismtracker[77281:6776834] 19:41:45.264 WARNING:  140: This application, or a library it uses, is using the deprecated Carbon Component Manager for hosting Audio Units. Support for this will be removed in a future release. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h.
```

Since this is just a warning (on OS X 10.11.6, I don't know the state on macOS 10.12), I don't think we need to rush to fix this, but we might keep this issue here to track it and share ideas on how to fix it.

### LINKED ISSUE #240: [RFC] SDL2 port decisions
Assuming an SDL2 port is worthwhile, after some casual attempts at the port myself, I see at least two basic decisions that must be made:

1. What to do about the existing pile of video backends schismtracker implements from YUV planes to OpenGL support

2. What to do about the removal of the .unicode member of keyboard events.

For 1, I'd like to propose just throwing out everything schismtracker implements for video stuff in favor of always using an SDL2 SDL_Renderer, use an SDL_Texture as the equivalent of the current OpenGL driver's framebuf, and basically always operate as if it were an OpenGL backend with bpp=4, but using SDL_Renderer API to upload the vgamem to the SDL_Texture framebuf, and copying that texture to the SDL_Renderer every update.  No more direct OpenGL API usage, no more YUV support, just deprecate all that cruft and expect people who still want to use that to just use an SDL1.2 old schismtracker version.  See https://wiki.libsdl.org/MigrationGuide#If_your_game_just_wants_to_get_fully-rendered_frames_to_the_screen for more information.  Basically we'd just lean entirely on SDL2's SDL_Renderer implementation to handle all the details schism's existing legacy video driver mess tries to do currently. 

For 2 I'm not sure what's best to do.  There's direct support for IME handling in Schismtracker, but SDL2 has deprecated the .unicode member in favor of a new TEXTINPUT event type which brings in a UTF-8 string having potentially multiple characters in a single event.  It really doesn't align with the existing input model of schismtracker, but there's no trivial way to get back the .unicode equivalent on SDL2.   See https://wiki.libsdl.org/MigrationGuide#Input for more information.  It's possible we could try synthesizing the .unicode value on keypresses from the keycode and modifier states, and ignore the new TEXTINPUT stuff, continuing to rely on schism's builtin IME features exclusively.  I think otherwise rearranging the input handling to fit the TEXTINPUT events would be a pretty invasive change, though it might provide better non-english input handling.  Maybe something along the lines of what's done here https://github.com/MrAlert/sdlcl/blob/master/events.c makes sense to pull into schism for producing 1.2-style unicode input.

Rather than just try make these decisions myself in isolation and submit an SDL2 port PR nobody actually likes and never merges because we never came to a consensus on throwing out a pile of legacy code beforehand, I'm hoping we can decide what to do here in discussion before attempting to merge SDL2 support.

I'm also curious if perhaps we should just ignore SDL2 and expect people to use a compatibility layer like https://github.com/MrAlert/sdlcl to run schism atop SDL2.

mrpapersonic:
Closes #240, closes #33

@vcaputo's PR with additional fixes and tweaks.
List of current changes/fixes:
* [81533c5676c337dc57ccf64da6b6c304c6a1799d]: Simple windows build fix, `info.window` to `info.info.win.window`
* [25784ced4e36ed691cc36b42f9be3f818c4bc286]: Replace hack in configure.ac with *real* OS checking
* [1b5e20f1601da2e12cbc6bb3fcad6c7b9c83b2a5] + [d98abfc0c1228fe40a7cb211e0cf18abb805bd4f]: Initial WASAPI support, thanks Vad1m for notifying me it was broken. System-dependent volume controls are removed in this commit, because AFAIK there's no way to change the volume of WASAPI in SDL2.
* [0884b44da9ac8066e41d16fa8b8b846686b02a54]: Moved from checking is_synthetic to a new int (`is_textinput`) in the key_event struct; easier to differentiate TEXTINPUT
* (multiple commits over time): Added keyboard input from SDL_TEXTINPUT
* [5804f574a4a7f1bccaf369d4a718a0766d3951d8]: key_scancode_lookup replaced by just raw scancodes, mouse segfaults are fixed by replicating what SDL 1.2 does, deleted a few unused source files


mrpapersonic:
@jangler It took a bit longer than expected, but it's fully working now (as far as I've tested).

Like with your opinion in #240, I don't think we should create diverging forks for SDL 1.2 and SDL2; however I think that we should keep a download for the last SDL 1.2 version on the website for any legacy systems that don't support SDL2 (somehow), and have a disclaimer that it's deprecated and won't be getting any updates for the forseeable future.

mrpapersonic:
bump

jangler:
@mrpapersonic, I talked this over with another contributor and we had the following questions/concerns:

- This PR is supposedly based on @vcaputo's, but his commits don't appear in this history of your branch. What's going on there?
- The commit history is a mess. Stuff like 12 commits titled "Update build.yml" (often back to back), multiple commits titled "What", etc. If this were a small PR, it wouldn't really matter since I would just squash it, but this is much too big to squash comfortably. I'm not even sure whether this would yield good results at this point, but would you be willing to soft reset back to master and redo the commits with a more logical organization?

Appreciate your contributions!

mrpapersonic:
> - This PR is supposedly based on @vcaputo's, but his commits don't appear in this history of your branch. What's going on there?
> 
> - The commit history is a mess. Stuff like 12 commits titled "Update build.yml" (often back to back), multiple commits titled "What", etc. If this were a small PR, it wouldn't really matter since I would just squash it, but this is much too big to squash comfortably. I'm not even sure whether this would yield good results at this point, but would you be willing to soft reset back to master and redo the commits with a more logical organization?

Sure! I'll soft-reset and commit my changes in a less "IDK WTF I AM DOING" way.
I was trying to merge the changes but they all just went in with all of my fullscreen changes (I probably merged the changes in a separate git folder and forgot about it). I'll attempt to merge the changes _properly_ after the soft-reset.

mrpapersonic:
Ok, I reset the SDL2 branch and imported all the changes from before, but with actual commit names now.
And @vcaputo's changes have been properly merged now.

mrpapersonic:
@jangler Sorry for mentioning you again about this, but can this be reviewed again when you have the time? I believe I've patched some of the major holes (mouse segfaults, text entry playing samples, ascii input barely functioning, pattern editor not even getting the right notes) to name a few.

jangler:
Ok, I'll have a look soon and poke around with the build. Will probably take a while.

jangler:
First impressions are that I love the results of the new SDL2 video code. It doesn't crash when resizing the window on Wayland, it doesn't discolor after resizing the window on X11, and we get nearest-neighbor resizing independent of a specific backend. I played around with basic stuff and everything seemed fine.

Sometime in the next week I'll have a look at the code changes and lightly regression test the relevant parts of the program.

jangler:
@mrpapersonic there's a note in #270 that says you referenced that issue from this PR, but it looks like the reference is gone now – do you remember what it was? Asking because I ran into vcaputo's issue while testing the new SDL2 audio drivers (JACK, PipeWire).

mrpapersonic:
> @mrpapersonic there's a note in #270 that says you referenced that issue from this PR, but it looks like the reference is gone now – do you remember what it was? Asking because I ran into vcaputo's issue while testing the new SDL2 audio drivers (JACK, PipeWire).

That was actually a typo. I meant #240 :p

jangler:
My notes on the code changes are:
- `cfg_video_gl_bilinear` should be removed from `include/it.h`
- `video_report()` should either be uncommented or removed

Going to do some testing tomorrow (probably).

jangler:
Haven't found any issues. If you don't expect to push any more changes, I'm comfortable merging this.

mrpapersonic:
I don't plan on editing it any more. Feel free to merge!

vcaputo:
Awesome, never imagined this would actually happen.  Apparently that half-baked port attempt wasn't a waste of time after all!

PR #306, opened 2022-08-14
Actions: install Git on Windows enviroments

mrpapersonic:
Before this, Schism Tracker was building with an ugly "Schism Tracker built...", instead of a "Schism Tracker 20220807".

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
       with:
         msystem: ${{ matrix.sys }}
         update: true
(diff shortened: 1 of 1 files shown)


jangler:
Is this okay to merge? Those last two commits look like they might have been added by accident.

mrpapersonic:
> Is this okay to merge? Those last two commits look like they might have been added by accident.

...oops. Those were meant to be pushed to a different branch fixing a few of the bugs with SDL2. I'll reset to the first commit and then it can be merged.

PR #312, opened 2022-09-01
Lua Support

### LINKED ISSUE #105: script plugins support
would be great if schismtracker could support script plugin support, in languages like Lua, Python, etc. - there we could create our own plugins, for processing information, for import/export, etc.

guysv:
## Intro
So I thought it might be cool to integrate a scripting REPL into schism.
Applications in mind:
* macro pattern insertions
* custom keybinds
* custom MIDI commands handling
  * to track with my MPK mini..
* unstandardized "lua" pattern effect to trigger a function
* schism lua init script snippets to share with your pals

.. you get the point. Also closes https://github.com/schismtracker/schismtracker/issues/105

## What do I have right now?
schismlua
* Added lua 5.4.4 lib as an optional dependency in configure.
* Added the lua engine module, that's in charge of interacting with the `lua_State L`
* Added the lua console page that's implementing basic REPL graphics and input
* Set the REPL scrollback as the backend for lua's `print()`
* Implemented some basic bindings
* **tl;dr, still can't track with lua >:(**

## An obvious defect:
* The Lua interpreter runs on the main thread, so executing an infinite loop will freeze schism.
  * I couldn't assess what parts of schism will need to be mutex-proofed so I delayed that task to another day.
  * A bandaid would be to sandbox lua and quota lua instruction, but that's barbaric lol.

## How to try it out:
```
# clone my fork
git clone https://github.com/guysv/schismtracker.git schismtracker-lua

cd schismtracker-lua
autoreconf -i
mkdir -p build
cd build
../configure --with-lua && make
./schismtracker
```  
Go to pattern editor page, then press `alt+F12`

-----
## So, what do you think?
Could we outline some roadmap towards getting this merged?
What else do you wanna see being developed?


sagamusix:
> Could we outline some roadmap towards getting this merged?

Well, without a clearly defined scripting API that can do something more useful than starting and stopping song playback, there isn't much of a point in merging this. Some ideas what this API should probably be capable of doing could be taken from this thread: https://forum.openmpt.org/index.php?topic=4078.0

> The Lua interpreter runs on the main thread, so executing an infinite loop will freeze schism.

One way to work around this is installing a hook that is called after every X Lua VM operations and then for example check in this hook if the user pressed ESC and then terminate the script from this hook (`lua_sethook(lua, HookFunc, LUA_MASKCOUNT, 1000);` would do this after every 1000 operations).

guysv:
>One way to work around this is installing a hook that is called after every X Lua VM operations and then for example check in this hook if the user pressed ESC and then terminate the script from this hook (`lua_sethook(lua, HookFunc, LUA_MASKCOUNT, 1000);` would do this after every 1000 operations).

Looking into the subject, I thought that multi-tasking schism and the running lua code as coroutines could be nice. this looks promising: https://github.com/evg-zhabotinsky/yieldhook

>Well, without a clearly defined scripting API that can do something more useful than starting and stopping song playback, there isn't much of a point in merging this.

Definitely. This is work in progress unfortunately. Still thinking about how I should access the static variables of `page_patedit.c` from `lua-engine.h`. Stay tuned for updates i guess ;)

vanfanel:
Will LUA support be optional at build time, please? Some of us just want old, classic IT that runs anywhere without slow interpreted languages installed on our systems.

guysv:
> Will LUA support be optional at build time, please? Some of us just want old, classic IT that runs anywhere without slow interpreted languages installed on our systems.

Of course it is optional, see the new `--with-lua` configure argument...

vanfanel:
> > Will LUA support be optional at build time, please? Some of us just want old, classic IT that runs anywhere without slow interpreted languages installed on our systems.
> 
> Of course it is optional, see the new `--with-lua` configure argument...

Ah, yes, my fault. Thanks.

guysv:
## Quick update i guess

### lua multi-tasking
I pretty much put all my money on abusing the lua-instruction-counter-debug-hook to preemptively multitask lua code with the schism event loop. Had couple of nasty lua-stack-balance bugs on the way there, still wonder if threads are not the way to go, anyway it's technically cooler this way so what the hell ;) 

### pattern editing API
Added some basic functions to retrieve pattern position, and note/param setters

### lua callbacks, MIDI control message handling
Added a lua task queue that is unloaded when the lua console is idle. Now schism C code can post lua execution requests, that eventually invoke a lua global function the user can overwrite in the console.

For MIDI, that's `_on_midi(value, param)` (Only MIDI control messages for now)

### A Humble Demo
Ok so now that I got all that running, I wanted to see if all that lua scripting stuff is actually practical.

Like I mentioned earlier I have some MPK mini lying around. Inspecting the keyboard with a MIDI monitor shows that the knobs emit MIDI control messages with value and param (mapped to knob position).

I then tried this midi callback:
```
cp = current_pattern
cc = current_channel
cr = current_row
_on_midi = function(v,p) pattern.set_param(cp(), cc(), (cr()+1)%64, v*2) end
```
(looks like clipboard paste into the lua console is broken, WIP...)

Anyway, combine this with schism MIDI pattern trigger. I can now sweep the channel effect param value with the knob! I was even able to quickly compose a tune that was actually _listenable_. Check out the [track](https://modarchive.org/index.php?request=view_by_moduleid&query=199123) 😎 !

You might be wondering about the `(cr()+1)%64`. The point was to make the param modification happen ahead of the cursor, so I could immediately listen to the track _while_ i'm jamming. This also means this hack sacrifices some hit accuracy for usability I guess this is an inherent issue with the matrix model of the tracker.

guysv:
Short update, I added a user script loader, so your lua macros could live across schism sessions. Useful to populate your hooks on startup. This should also make your life easier sharing scripts with the world.

coderofsalvation:
@guysv this is incredibly cool. I really like this. Perhaps we can have a small chat somewhere?
There's a LUA interpreter in a similar state for Milkytracker, perhaps we could put function-names next to eachother, and hash out a spec which makes .lua scripts work for both?
Feel free to ping me on mastodon/twitter (see my profile) to brainstorm a bit.

> One way to work around this is installing a hook that is called after every X Lua VM operations and then for example check in this hook if the user pressed ESC and then terminate the script (...)

@sagamusix really like this clever approach

esaruoho:
how could this proceed somehow?

coderofsalvation:
@sagamusix would you be interested to jump on a call with me and @guysv at some point, and brainstorm a bit?
@esaruoho would you be interested in implementing these efforts into ztracker_mac as well?

esaruoho:
> @esaruoho would you be interested in implementing these efforts into ztracker_mac as well?

ztracker_mac does not currently even compile. I was hoping to find someone who could help with it, but it seems that free devs that know macOS and MIDI and care about trackers are nonexistent :)

Good idea otherwise. I have a hard time believing that zTracker will ever make it to macOS unless if i can find and pay a developer to convert it from Windows to macOS.

guysv:
@coderofsalvation umm sure i can hop into some call. i'd be happy to chat with you about this as this topic is indeed dear to my heart.

sagamusix:
Sure, why not. Do not though that OpenMPT's WIP API is object-oriented similar to Renoise's API, as my goal too was to make it mimic something that already existed.

Here's a very WIP documentation of the API: https://github.com/sagamusix/openmpt/blob/master/packageTemplate/Scripting/api.lua (some of those things mentioned there do not exist yet but should be part of the final product)

coderofsalvation:
OK, what timezone's are we in? (Im in UTC+1)

How about we explore some directions of interest (and dis-interest) regarding to the (pretty extensive) renoise API.
Even if zero actions will follow from it, I think it's still an important conversation to have.

Afaik current interests are:

| topic | usecase |  comment |
|-|-|-|
| midi | import (.mid) file | escape-hatch for other software |
| midi | export (.mid) file | escape-hatch for other software |
| midi | CC/NOTE/start/stop hooks | alternate ways to control UI/pattern |
| sampler | generate/process waveform | @coderofsalvation did some experiments with lua-driven synths/fx using<br>using [function(x,i,samples,srate)](https://github.com/coderofsalvation/MilkyTrackerX/blob/master/patch/src/lua/config.lua#L122C1-L124C4) (mono though) |
| (slider)dialog | meta control of UI & lua-scripts | @coderofsalvation instead of offering draw-anything dialogs, sliders cover most usecases |

> some thoughts: I'm not sure till which degree following the renoise API will make sense for schismtracker/milkytracker.
Their codebase/fileformats/UI imposes various restrictions, therefore making compatibility a pipedream for the near future.
Perhaps we could come up with an 'overlay' API  which easily maps on top of RENOISE's/OpenMPT OO structure.
That way each tracker can keep their unique lua API, but still enjoy an extra layer which supports lua-plugins. For example in renoise's case:

```
renoise.UCP = {...the UCP implementation..}

# tracker-specfic mappings (this part differs from tracker to tracker)
renoise.UCP.Sample.play = renoise.Song.Sample.play  # play(note [, volume [, panning]])
```

> btw. UCP is something I just made up (Universal Tracker Plugins).

coderofsalvation:
apoligies, life got a bit in the way.
Anyways, I'd be up for a call tomorrow (wednesday 27 march) around 20:00 PM UTC+8.
Or perhaps it's better next week's tuesday or wednesday?
WDYT?

Here's a secure [Matrix link](https://call.element.io/room/#/lua--xmitmod-brainstorm?roomId=!JFoidnLnxuLrMbcgDT:call.ems.host&password=SIwXupGW2gMvMFX_sugExw) for the call

sagamusix:
Sorry, way too spontaneous. Maybe sometime in April. European evening hours would work here.

coderofsalvation:
np, there's no hurry.

guysv:
@coderofsalvation where can I DM you? didn't seem to reach you on mastodon.

coderofsalvation:
update: meanwhile we've connected

catbusconductor:
Something still happening here? Would be really cool :)

guysv:
Unfortunately the project got stalled a bit as I just kept bumping into (and fixing) older schism bugs. In the meanwhile I am taking a look on milkytracker on which @coderofsalvation have made some progress towards lua tracking on his own. I'm afraid we are planning on joining forces to get lua done in milky properly first.

esaruoho:
@guysv how's that going? any way this could be updated, too? :)

Issue #311, opened 2022-08-20
Aspect ratio option?


vanfanel:
Hi,

Even if Impulse was originally 640x400, fullscreen should have an option to be aspec-corrected on modern 16:9 screens: it originally filled 4:3 monitors.

It's very easy to do, using https://wiki.libsdl.org/SDL_RenderSetLogicalSize

You seem to be doing ratio correction here already:
https://github.com/schismtracker/schismtracker/blob/26066177bfdb53913534ac412ed3ffff7b70370f/schism/video.c#L277

..but you do it using the host screen size. An solution could be that `SDL_RenderSetLogicalSize()` in **video_resize()** uses 4:3 if the hypothetical aspect ratio option is ON.

For now, I will force 4:3 on my own builds :)

vanfanel:
@jangler Did you consider this?

jangler:
It's a fine suggestion; I just haven't had much time for Schism dev lately. I would be happy to merge a PR that introduces such an aspect ratio option.

PR #322, opened 2022-09-30
Add working aspect ratio correction option (want_fixed)

vanfanel:
This fixes issue: https://github.com/schismtracker/schismtracker/issues/311

I have used the name `want_fixed` for the option to match the docs, which mention it here:
https://github.com/schismtracker/schismtracker/blob/master/docs/configuration.md
but it wasn't implemented before.

### DIFF
--- a/schism/video.c
+++ b/schism/video.c
@@ -276,7 +276,13 @@ void video_startup(void)
 
 void video_resize(unsigned int width, unsigned int height)
 {
-	SDL_RenderSetLogicalSize(video.renderer, width, height);
+	/* Aspect ratio correction if it's wanted */
+	if (cfg_video_want_fixed)
+		SDL_RenderSetLogicalSize(video.renderer,
+					 NATIVE_SCREEN_WIDTH * 5, NATIVE_SCREEN_HEIGHT * 6); // 4:3
+	else 
+		SDL_RenderSetLogicalSize(video.renderer, width, height);
+
 	video.prev.width = video.width;
 	video.prev.height = video.height;
 	video.width = width;
--- a/include/it.h
+++ b/include/it.h
@@ -246,6 +246,7 @@ extern int show_default_volumes;        /* pattern-view.c */
 extern char cfg_video_interpolation[];
 /* TODO: consolidate these into cfg_video_flags */
 extern int cfg_video_fullscreen;
+extern int cfg_video_want_fixed;
 extern int cfg_video_mousecursor;
 extern int cfg_video_width, cfg_video_height;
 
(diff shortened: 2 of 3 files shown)


schism/video.c
```diff
@@ -276,7 +276,10 @@ void video_startup(void)
 
 void video_resize(unsigned int width, unsigned int height)
 {
-	SDL_RenderSetLogicalSize(video.renderer, width, height);
+	/* Aspect ratio correction if it's wanted */
+	if (cfg_video_want_fixed)
+		SDL_RenderSetLogicalSize(video.renderer, 4, 3);
```
vcaputo: Does this actually work correctly in practice with no ill artifacts?

It probably makes more sense to use the NATIVE_SCREEN_{WIDTH,HEIGHT} constants as this API isn't strictly about specifying an aspect ratio AIUI.  It's size *and* ratio.

vanfanel: It does work correctly without any artifacts. It works by using the renderer to scale the image, think of it as "post-processing". Other FOSS projects use this, as SDLPoP here:
https://github.com/NagyD/SDLPoP/blob/3e74ba883c72969cbde7037087475b258488245b/src/seg009.c#L2379

As for using NATIVE_SCREEN_{WIDTH,HEIGHT} constants , you're probably right.
I have amended the commit and everything is in place now.

vcaputo: It's neat that this works, and looking at SDL2's implementation of SDL_RenderSetLogicalSize(), specifically UpdateLogicalSize() @ https://github.com/libsdl-org/SDL/blob/release-2.24.0/src/render/SDL_render.c#L2285 it does appear to strictly consider just the aspect ratio of the logical size operating entirely using floats - in *this* function.

But looking elsewhere in that listing, like the events code @ https://github.com/libsdl-org/SDL/blob/release-2.24.0/src/render/SDL_render.c#L780 it's unclear to me at a glance if this is uniformly working beyond appearances.  Admittedly I'm not familiar with this area of SDL2, and lack the time right now to dig further.

Other considerations like principle of least surprise being if anyone ever starts calling SDL_RenderGetLogicalSize() in schism it's very unlikely they'd expect 4x3 logical pixels to come back.  And yeah, the named symbolic constants of our internal resolution are inherently the aspect ratio defining values, and we should avoid sprinkling unnamed magic numbers throughout the code anyways.

Since you've already changed it to use the constants, this comment is mostly just academic/for posterity.

vanfanel: Really, this is the standard way to do aspect ratio correction in SDL2, there's not much more to it.
It works everywhere for what I have seen, this is an stable SDL2 function that shows good results where SDL2 runs.

jangler: I was going to mention that the native screen constants aren't 4:3, but it looks like you compensated for that already.

I haven't tested this yet, but using `NATIVE_SCREEN_*` instead of `width` and `height` in the `else` case looks wrong to me. IMO better to assume that someone does not want *any* aspect correction if the config var isn't set.

vanfanel: Yep, I agree on that. Now using `width` and `height` in the `else`.
No correction means no correction, indeed! :)

PR #324, opened 2022-10-02
Disable WM composition bypass.

drradio:
Set SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR to "0" to avoid composition bypassing.


jangler:
Technically a value of 0 indicates "no preference", according to <https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm46409506222736>. A value of 2 would hint the WM not to disable compositing.

I think 0 is a sensible default for Schism, although I will wrap this in an `if` so it doesn't override an existing environment variable.

jangler:
Never mind; [the SDL_SetHint documentation](https://wiki.libsdl.org/SDL_SetHint) says:
> Hints will not be set if there is an existing override hint or environment variable that takes precedence.

PR #325, opened 2022-10-02
Fix segfault when SDL2 runs on Wayland or KMS/DRM.

vanfanel:
There's a long-standing bug on MESA that causes a **segfault** if `SDL_Quit()` is called inside an `atexit()` handler, see:

https://github.com/libsdl-org/SDL/issues/3184
(There you can see the many SDL2 projects that are affected by this, and the information on the MESA mailing list where it was discussed).

It was ultimately left from SDL2 to MESA, where the MESA devs concluded that **freeing the GBM stuff in an atexit() handler is generally a bad idea.**
NOTE: _GBM is used by MESA and KMS/DRM on GNU/Linux for video buffer management. Calling SDL_Quit() implies freeing the GBM stuff._

This fixes several **segfaults** on `exit()` calls in Schismtracker if using SDL2 on Wayland (modern GNU/Linux graphics protocol) or KMS/DRM (dos-like SDL2 backend without X11 or Wayland).

This will have to be faced sooner or later, now that GNU/Linux distros are adopting Wayland and X11 is going away for good.

### DIFF
--- a/include/it.h
+++ b/include/it.h
@@ -455,6 +455,9 @@ void memused_songchanged(void);
 
 void memused_get_pattern_saved(unsigned int *a, unsigned int *b); /* wtf */
 
+/* Shutdown the SDL2 system from anywhere without having to use atexit() */
+void schism_exit(int status);
+
 /* --------------------------------------------------------------------- */
 
 #endif /* ! IT_H */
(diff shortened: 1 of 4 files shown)


jangler:
Is there a reason you didn't address the exit paths in `schism/page.c`? If they were simply overlooked I can handle them.

vcaputo:
Since there are no existing users of schism_shutdown() besides the atexit() callback registration, why not just rename it to schism_exit(int status), put the exit(status) at its end, and change all the existing exit(status) call sites to schism_exit(status)?

vanfanel:
@vcaputo Good idea. Done!

@jangler Ah, yes, `schism/page.c` exit paths were an omission. It's now taken care of ;)

PR #328, opened 2022-10-20
Ignore synthetic keystrokes in the pattern editor

### LINKED ISSUE #317: Duplication when typing a digit in the pattern editor
Hi. In the last two releases, in the pattern editor when you enter a digit in the volume or effect column, the digit is entered twice (it is repeated in the next column or if it is the second column it is repeated below).

Example:   
C-501  00  O33  <-- When typing the first "3", a second 3 is automatically typed next to it 

C-501  00  O03  <-- In this case, when a 3 is entered in the second column, a 3 is automatically entered below.
                    30

mrpapersonic:
Closes #317.

### DIFF
--- a/schism/page_instruments.c
+++ b/schism/page_instruments.c
@@ -957,6 +957,8 @@ static int note_trans_handle_key(struct key_event * k)
 
 			case 2:        /* instrument, first digit */
 			case 3:        /* instrument, second digit */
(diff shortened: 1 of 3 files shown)


jangler:
Confirmed that this fixes all pattern editor columns, the order list, and the instrument editor note translation widget.

PR #332, opened 2022-11-22
Quantize early MIDI notes to the next row

guysv:
So for some time now I suspect that MIDI pattern insertions while the track is running are inaccurate. From inspecting the code I see that even with MIDI quantization being on, late notes are snapping to the current row in the pattern, even if it's a tick away from traversing to the next row.

Here's a gif demonstrating the fix, I recorded the latest schism released with my locally-compiled instance acting on the same MIDI signals:
quantize-captions

I think you'd agree that's an improvement 😄


schism/page_patedit.c
```diff
@@ -2958,28 +2958,41 @@ static int patedit_record_note(song_note_t *cur_note, int channel, UNUSED int ro
...
 	speed = song_get_current_speed();
 	tick = song_get_current_tick();
+
+	/* correct late notes to the next row */
+	/* tick + 1 because processing the keydown itself takes another tick */
+	if (midi_flags & MIDI_TICK_QUANTIZE && !first_note && tick + 1 < speed / 2) {
+		r = (r + 1) % song_get_rows_in_pattern(current_pattern);
+		quantize_next_row = 1;
+	}
+
```
jangler, line 3006: If a MIDI note is entered on the last row of a pattern and should be quantized to the next row, does it just write to the first row of the current pattern then?

guysv: well.. yeah. What other behavior do you prefer?
* clamp row to max-row
* write note to next pattern

jangler: Ideally it would write to the next row that would be played, which requires some nontrivial logic (which could hopefully be factored out of wherever it exists in the codebase already):

- If the current pattern is being looped, the correct behavior is to write to the first row of the current pattern
- If the song is being played normally, the correct behavior is to write to the first row of the next pattern in the order list
- Jump/break commands (Bxx, Cxx, SBx) should be correctly accounted for (in general, not just at the end of a pattern)

guysv: very interesting! will work on this soon.

guysv:
hi @jangler I pushed more code to move the row-shifting logic to a song-lib function.

I support the pattern loop/song play cases but didn't bother considering row-jump effects. Supporting those is a minor use case, I don't see anyone actually run into trouble with. The user could always record notes and move them around the track later.

a-bit off-topic for this PR, but I set the `song_keydown` to play the actually-selected sample when striking midi key. the previous value was `KEYJAZZ_NOINST` which actually most of the time play nothing. I couldn't figure out what was the point having `KEYJAZZ_NOINST` in the first place.

also, midi note-off quantization is also fixed now.

jangler:
Hmm, okay. I know some people who use Bxx and SBx a lot, but I'm guessing they're not the same people who would be interested in recording MIDI into the tracker. Cxx is very common for me, since it's the most straightforward way to have patterns with an effective number of rows other than 64 in S3M format, but that's less of a problem than an unhandled Bxx or SBx, since the note would just be written into the dead space at the end of the pattern instead of potentially overwriting something else. Let me take a look at the code.

schism/page_patedit.c
```diff
@@ -2958,28 +2960,50 @@ static int patedit_record_note(song_note_t *cur_note, int channel, UNUSED int ro
...
 			playback_tracing = 1;
 			break;
 		};
+
+		first_note = 1;
 	}
 
+	/* correct late notes to the next row */
+	/* tick + 1 because processing the keydown itself takes another tick */
+	if (midi_flags & MIDI_TICK_QUANTIZE && SONG_PLAYING && !first_note
```
jangler: This the only place `first_note` is used, so making this conditional an `else if` of the previous one would be simpler and easier to understand.

schism/mplink.c
```diff
@@ -222,6 +222,29 @@ int song_find_last_channel(void)
 
 // ------------------------------------------------------------------------
 
+int song_get_pattern_offset(int * n, song_note_t ** buf, int * r, int offset)
```
jangler: I'd prefer the name `row` instead of `r` here. It's plenty short and much easier to read.

schism/mplink.c
```diff
@@ -222,6 +222,29 @@ int song_find_last_channel(void)
...
+{
+	int len, tot;
+	if (song_get_mode() & (MODE_PATTERN_LOOP)) {
```
jangler: Not a big deal, but `MODE_PATTERN_LOOP` doesn't need to be parenthesized here.

schism/mplink.c
```diff
@@ -222,6 +222,29 @@ int song_find_last_channel(void)
...
+		// just wrap around current rows
+		*r = (*r + offset) % song_get_rows_in_pattern(*n);
+		return song_get_pattern(*n, buf);
```
jangler, line 236: This doesn't change the contents of `buf`, right, since we're not changing patterns? In that case you could pass `NULL` instead of `buf` here.

schism/mplink.c
```diff
@@ -222,6 +222,29 @@ int song_find_last_channel(void)
...
+	}
+
+	tot = song_get_rows_in_pattern(*n);
+	while (offset + *r > tot) {
+		offset - tot;
```
jangler: This statement doesn't do anything.

schism/mplink.c
guysv: what do you mean by
> since we're not changing patterns

this function replaces the original `song_get_pattern` call in `pattern_editor_insert_midi` to select and instantiate the pattern buffer with the right #n pattern according to the offset.

`buf` is still NULL here for that matter

schism/page_patedit.c
guysv: guess you're right. fixing.

schism/mplink.c
guysv: fixing.

schism/mplink.c
guysv: right. fixing.

schism/mplink.c
guysv: correct. i meant `offset -= tot` fixing.

guysv:
@jangler this is ready for merge.

schism/page_patedit.c
guysv: there is still a bug here. what if `midi_start_record` is zero and the song is playing? keyjazz will be quantized to next row based on tick, even if the trigger mode (ctrl-z) is none. maybe add `&& midi_start_record`

schism/mplink.c
jangler: Never mind, I disagree with what I said originally.

jangler:
What did you decide about this?

> there is still a bug here. what if midi_start_record is zero and the song is playing? keyjazz will be quantized to next row based on tick, even if the trigger mode (ctrl-z) is none. maybe add && midi_start_record

guysv:
>What did you decide about this?
>
>>there is still a bug here. what if midi_start_record is zero and the song is playing? keyjazz will be quantized to next row >>based on tick, even if the trigger mode (ctrl-z) is none. maybe add && midi_start_record

well i got rid of the first_note variable, and moved the quantization check to the `else` of the `midi_start_record` so it will not qunatize when `midi_start_record && !(song_get_mode() & (MODE_PLAYING|MODE_PATTERN_LOOP))` (i.e. first note)

esaruoho:
what's the endresult of this? is it gonna get in?

guysv:
I kinda bailed on schism for milky tbh. feel free to close.

esaruoho:
@jangler @mrpapersonic any take on how this could proceed if the PR maker has abandoned it (unfortunately)?
it seems like it provides features that would be useful for the non Bxx SBx crowd.

mrpapersonic:
if and when the time comes I'll likely import the branch do my own changes on it, similarly to what I did with Vito's SDL2 branch

Whether that actually happens or not is based off mainly when I get the interest to :)

jangler:
I'm actually not sure why I didn't merge this a year ago

PR #340, opened 2022-12-02
Fix bashism in configure.ac sdl2-config check

thesamesam:
configure scripts need to be runnable with a POSIX-compliant /bin/sh.

On many (but not all!) systems, /bin/sh is provided by Bash, so errors like this aren't spotted. Notably Debian defaults to /bin/sh provided by dash which doesn't tolerate such bashisms as '=='.

This retains compatibility with bash.


jangler:
Appreciate it!

thesamesam:
Thanks!

PR #342, opened 2022-12-06
FLAC loading support

mrpapersonic:
Can be disabled at build time with `./configure --without-flac`, and is already disabled by default if you don't have the FLAC libraries installed.


fmt/flac.c
```diff
@@ -0,0 +1,343 @@
...
+	} raw_data;
+	uint8_t* uncomp_buf;
+};
+static uint32_t samples_decoded = 0, samples_read = 0;
+static size_t total_size = 0;
+
+static void on_meta(const FLAC__StreamDecoder *decoder, const FLAC__StreamMetadata *metadata, void *client_data) {
+	struct flac_file* flac_file = (struct flac_file*)client_data;
+    switch (metadata->type) {
+        case FLAC__METADATA_TYPE_STREAMINFO:
```
jangler: Seems like spaces keep creeping into your commits somehow ;)

fmt/flac.c
```diff
@@ -0,0 +1,343 @@
...
+
+	if (!flac_file->streaminfo.total_samples || !flac_file->streaminfo.channels)
+		return FLAC__STREAM_DECODER_WRITE_STATUS_ABORT;
+
+	if (flac_file->streaminfo.channels > 2)
+		return FLAC__STREAM_DECODER_WRITE_STATUS_ABORT;
+
+	if (!(flac_file->streaminfo.bits_per_sample == 8 || flac_file->streaminfo.bits_per_sample == 12 || flac_file->streaminfo.bits_per_sample == 16
+		|| flac_file->streaminfo.bits_per_sample == 20 || flac_file->streaminfo.bits_per_sample == 24 || flac_file->streaminfo.bits_per_sample == 32))
+		return FLAC__STREAM_DECODER_WRITE_STATUS_ABORT;
```
jangler: This would be more readable as a switch statement with the return in a default case

fmt/flac.c
```diff
@@ -0,0 +1,343 @@
...
+		case 20: case 24: case 32: {
+			int16_t *buf_ptr = (int16_t*)flac_file->uncomp_buf + samples_decoded;
+			uint32_t bit_shift = flac_file->streaminfo.bits_per_sample - 16;
+			for (i = 0, j = 0; i < block_size; j++) {
+				buf_ptr[i++] = buffer[0][j] >> bit_shift;
+				if (flac_file->streaminfo.channels == 2)
+					buf_ptr[i++] = buffer[1][j] >> bit_shift;
+			}
+			break;
+		}
```
jangler, line 203: I don't see a way to combine these cases without (probably) degrading performance, so the duplication here is okay

jangler:
Tested with and without libflac-dev, works as expected for the files I have.

mrpapersonic:
I tested it on [these test files](https://github.com/ietf-wg-cellar/flac-test-files) and everything seems to work as expected. One of the metadata extremes fails because there's no total samples block, which is used to calculate how much memory to allocate for the uncompressed audio.

esaruoho:
I'd love to know how to get into this. I've never built a Schism Tracker locally. Would this work on both Apple Silicon macOS Monterey, and Intel macOS Mojave?

mrpapersonic:
> I'd love to know how to get into this. I've never built a Schism Tracker locally. Would this work on both Apple Silicon macOS Monterey, and Intel macOS Mojave?

Schism is being built for both architectures, and on Intel it should be compatible to at least Mavericks. There's pre-built binaries [here](https://github.com/schismtracker/schismtracker/actions/runs/3630935839), scroll down to Artifacts and download the macOS version.

jangler:
Looks good. I'll merge tomorrow to leave others a chance to comment.

Issue #319, opened 2022-09-13
Pressing space to edit Instrument List names also inserts space


jangler:
Broken in `20220905`, worked in `sdl1`.

BKVad1m:
also broken in pattern editor

https://user-images.githubusercontent.com/76803825/194276562-7fbdf93b-9cc4-4a2d-b7e2-d76877475007.mp4

after fixed give me code to compile my tracker

Issue #318, opened 2022-09-13
Can't tab from sample name field in Sample List to other controls on page


jangler:
Broken in `20220905`, worked in `sdl1`.

PR #343, opened 2022-12-09
Native SDL_TEXTINPUT handling

mrpapersonic:
Before, I made a compatibility layer that would attempt to make "fake" key inputs with the text provided by SDL_TEXTINPUT, but this didn't work very well, and posed a number of problems. See: #317, #318, and #319.
So, this is an attempt at making a whole new text input handler. So far, it seems to work fine with everything I tried. :)


jangler:
Some issues I found:

1. Keybindings that open text input dialogs in the sample list or instrument list *without* the sample/instrument name field focused also enter the key into the dialog, ex. Alt+Shift+Z in the sample list enters a capital Z into the dialog.
2. Key bindings in the instrument list don't seem to work at all if the name field is focused, even if the field isn't in text input mode.
3. The first key I enter when editing an instrument name is eaten.

mrpapersonic:
Note: Issue 1 looks to be platform-specific, it only happens on macOS and Linux.

jangler:
Verified that the issues I mentioned earlier are fixed (on WSL).

A regression test is needed.

PR #348, opened 2023-01-14
drop MIDI note-offs that don't match last note

guysv:
imagine MIDI messages coming in this order into the midi insert routine:

image

this happens frequently due to how I play the piano, when I press other keys before I release previous ones. Unfortunately, in schism, the C3 note-off is read as the C3# note-off, as they are both written to the same channel.

now, note-offs are recorded only if their note matches the last-note played. if there's no match, the note-off is dropped.

### DIFF
--- a/schism/page_patedit.c
+++ b/schism/page_patedit.c
@@ -105,6 +105,7 @@ int midi_playback_tracing = 0;
 static int panning_mode = 0;            /* for the volume column */
 int midi_bend_hit[64];
 int midi_last_bend_hit[64];
+int midi_last_note[64];
 
 /* blah; other forwards */
 static void pated_save(const char *descr);
@@ -2983,6 +2984,11 @@ static int pattern_editor_insert_midi(struct key_event *k)
 	if (k->midi_note == -1) {
 		/* nada */
 	} else if (k->state == KEY_RELEASE) {
+		/* don't record noteoffs of non-matching notes */
+		if (k->midi_note != midi_last_note[c]) {
+			return 0;
+		}
+
 		c = song_keyup(KEYJAZZ_NOINST, KEYJAZZ_NOINST, k->midi_note);
 
 		/* don't record noteoffs for no good reason... */
@@ -3006,7 +3012,7 @@ static int pattern_editor_insert_midi(struct key_event *k)
 		if (!((song_get_mode() & (MODE_PLAYING | MODE_PATTERN_LOOP)) && playback_tracing)) {
 			tick = 0;
 		}
-		n = k->midi_note;
+		midi_last_note[c] = n = k->midi_note;
 		c = song_keydown(KEYJAZZ_NOINST, KEYJAZZ_NOINST, n, v, current_channel);
 		cur_note = pattern + 64 * current_row + (c-1);
 		patedit_record_note(cur_note, c, current_row, n, 0);


vcaputo:
Hmm, it's too bad IT's NNAs aren't powerful enough to just have the note-off occur for each of the notes when released.

What you've proposed seems like probably the best compromise..

guysv:
what is this NNA you are talking about? I tried reading the source but couldn't figure how to use the feature..

In the end, the bug is inherent due to the fact all MIDI messages are handled on the same track channel. (unlike actual piano where each key pretty much gets its own dedicated "channel"). are NNA supposed to fix this on playback level?

vcaputo:
NNAs is an Impulse Tracker feature that Schism Tracker replicates, being an IT clone.  Here's some of the relevant text from `it214v5/IT.TXT` found in the old DOS program's .zip, FYI:

```
          S73   Set NNA to note cut
          S74   Set NNA to continue
          S75   Set NNA to note off
          S76   Set NNA to note fade
        
                  These options allow you to override the default NNA for a
                  particular NOTE (ie. it does not affect the instrument)
                  See section 2.5 for an explanation of NNAs.
         
                  Note: You *MUST* be in instrument mode for these to have
                        any effect.
```

and

```
         The powerful feature of this tracker, though, is not the
         envelopes by themselves - it's the New Note Actions! What these
         options do is allow you to select what should happen to an
         instrument when another note is played in the same column. If NNA
         "Cut" is selected, then the previous note will immediately be
         stopped (like in all other trackers). If NNA "Continue" is
         selected, then the note will continue playing! This is especially
         useful for Drum Parts, where there may be a long-ish snare sample
         - you can go on ahead and put a bass drum in the very next row
         after it - the snare drum will still complete playing! NNA "Note
         Off" issues a note off command to a note when a new note is played
         in the channel. This is particularly useful in combination with
         volume envelopes and volume envelope sustain loops. NNA "Note
         Fade" causes the current note to fade out with the fadeout value
         when a new note is played.

        WARNING!!
         New Note Actions are EXTREMELY powerful, but they CAN cause
         problems if you are not careful. Selecting NNA Fade with a fade
         value of 0, or note continue with a looped sample (no volume
         envelope) or anything which can cause a quick build up of
         allocated channels can easily hang a slow computer (or even a
         quick one, for that matter!) I tried to put checks against this,
         but they always triggered too late - "past the point of no return"
         where the CPU becomes so bogged down with processing the
         information, that it can't do or try to do anything else. (The
         tracker has *MANY* calculations to do - the most time consuming
         are the mixing routines, which require the processor to process as
         many bytes as the mixing speed per second for EACH note ie. a
         mixing speed of 44kHz means that for each note playing, 44000
         calculations have to be made EVERY second... so with 64 channels
         'active' at the maximum mixing rate for a SB16, almost 3 MILLION
         calculations have to be done EVERY second to produce the sound (in
         mono)!!! (My 486 can cope with this, but my 386 just dies!)
```

If you download an Impulse Tracker .zip like from https://archive.org/details/demoscene_ImpulseTracker214 check out IT.TXT, much of it applies to Schism usage.

guysv:
cool. I think I did ran into those file at some point.

this have nothing to do with the issue at hand right? NNAs are not related to the fact that \*MIDI note-offs\* from multiple notes write to the same channel without regard to whether their note is still actually playing.

vcaputo:
Well, they're just not powerful enough to help here.

They seem semi-related in my mind since the NNA influences what happens to the existing note being played in the same channel...  That's similar to what you're addressing, but we can't exploit them to do it better AFAICT.

guysv:
@jangler WDYT?

PR #351, opened 2023-01-30
libs: remove vestigial libSDL-1.2.0 artifact

vcaputo:
This is irrelevant since transitioning to SDL2


vcaputo:
I'm unsure if this was ever particularly important, but it sure seems pointless now...

Is there need to get an SDL2 dylib binary committed in its place?  We seem to be getting on without it just fine AFAICT

jangler:
I don't think there's any need for a SDL2 dylib in the repo. Building SDL2 is part of the macOS build routine.

PR #353, opened 2023-02-26
rocket: implement GNU Rocket integration

vcaputo:
This adds a --with-rocket configure flag for enabling
experimental GNU Rocket support in Schism.

When enabled in the build, a series of new runtime flags are 
added:

--rocket (enable rocket mode)
all these overrides imply --rocket:
--rocket-host=HOST_OVERRIDE
--rocket-port=PORT_OVERRIDE
--rocket-bpm=BPM_OVERRIDE (beats per minute)
--rocket-rpb=RPB_OVERRIDE (rows per beat)

BPM and RPB are RocketEditor concepts, which the GNU Rocket
protocol is totally ignorant of.  But they're provided to aid in
keeping the Editor and Schism in agreement on what Rocket row 
maps to what millisecond within the song and within the Rocket
track data.  This doesn't override the speed/tempo setting for 
the song within Schism, and you'll have to take care to pick
those values such that it's consistent with the Rocket values, if
you want the RocketEditor's patterns and current row to align
with Schisms.  It's a little awkward due to how IT defined the 
speed/tempo semantics.  The default Rocket values of 125BPM/8RPB
do seem to align with the default IT/Schism speed/tempo of 6/125
though.

As-is this commit only attempts to connect to the Editor once at
startup.  So you'll need to have RocketEditor running before
starting SchismTracker w/--rocket.  This area could be made more
robust before merging upstream into SchismTracker, but it is
already usable.  If viewed as an experimental demo developer
mode, this is probably good enough to merge as-is.

See https://github.com/emoon/rocket/pull/165 for more context on
what this is all about.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "gnurocket"]
+	path = rocket
+	url = https://github.com/rocket/rocket
--- a/Makefile.am
+++ b/Makefile.am
@@ -228,7 +228,12 @@ libs_flac=-lFLAC
 endif
 
 if USE_ROCKET
+# TODO: it's fragile to be intimately grabbing rocket/lib/*.[ch] stuff here,
+# and if we continue doing that we need to DTRT regarding os-specific defines
+# like USE_NODELAY
 files_rocket = \
+	include/rocket.h		\
+	schism/rocket.c			\
 	rocket/lib/base.h		\
 	rocket/lib/device.c		\
 	rocket/lib/device.h		\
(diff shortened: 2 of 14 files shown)


vcaputo:
This is something I'm experimenting with for streamlining the .IT-file-utilizing demo making process w/RocketEditor.

Figured it couldn't hurt to post here as a draft to give folks time to get on the same page and get some sense of how welcome this kind of feature is upstream.

But ultimately this branch was pushed first and foremost so @emoon and others could potentially use it for evaluating  https://github.com/emoon/rocket/pull/165, since this whole multi-client GNU Rocket scenario is totally novel and there's no other non-editor tooling style clients in existence for testing it.

PR #354, opened 2023-02-26
schism: drop NORETURN attribute use

vcaputo:
This is just generating pointless noise from gcc:

```
gcc -DHAVE_CONFIG_H -I. -I..  -D_USE_AUTOCONF -D_GNU_SOURCE -I../include -I.  -I/usr/include/SDL2 -D_REENTRANT -DUSE_ALSA -DUSE_DLTRICK_ALSA -DUSE_OSS -DUSE_NETWORK -DUSE_X11 -DUSE_XV -DHAS_YM3812=1 -DHAS_Y8950=0 -DHAS_YM3526=0     -DUSE_FLAC -DUSE_ROCKET -g -O2 -MT schism/main.o -MD -MP -MF $depbase.Tpo -c -o schism/main.o ../schism/main.c &&\
mv -f $depbase.Tpo $depbase.Po
../schism/main.c: In function ‘event_loop’:
../schism/main.c:1086:1: warning: ‘noreturn’ function does return
 1086 | }
      | ^
```

I see no obvious reason why it's important for the compiler to be informed of event_loop()'s not returning, and right now it's just adding a false-positive warning to the build.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -53,9 +53,6 @@
 # ifndef UNUSED
 #  define UNUSED __attribute__((unused))
 # endif
-# ifndef NORETURN
-#  define NORETURN __attribute__((noreturn))
-# endif
 # ifndef PACKED
 #  define PACKED __attribute__((packed))
 # endif
@@ -68,9 +65,6 @@
 # endif
 # ifndef PACKED
 #  define PACKED
-# endif
-# ifndef NORETURN
-#  define NORETURN
 # endif
 # ifndef LIKELY
 #  define LIKELY(x)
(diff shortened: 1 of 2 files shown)


vcaputo:
If this is actually important for some reason, maybe turn event_loop() into an infinite loop so the compiler doesn't think it returns?  `for(;;);` at the end?

jangler:
Since the compiler disagrees with the annotation and it doesn't seem to be breaking anything, I think it's safe to assume we can get rid of it.

vcaputo:
I just remembered how the `exit(0)` was switched to `schism_exit(0)` by 76ec801a9b047b176d230963efda955e81a316f6, maybe that's when the warning appeared?

Either way, it's unclear to me what value marking `event_loop()` as noreturn has.. it's not like it's performance sensitive.

In case anyone decides it needs to go back in: maybe `schism_exit()` needs to be marked noreturn as well to silence the warning.

PR #365, opened 2023-06-02
Better keyjazz note-off matching

guysv:
So after doing #348 I noticed there's probably a better way to tackle the issue at hand.
- my previous PR solved the issue only for the pattern editor
  - this works seamlessly in sample/instrument page for example.
- this version also keeps all keyjazzing state arrays near each other which is also a plus

I also fixed some interesting bug here related to key repeats: I noticed that in master, holding two keys, then releasing one outputs a keyoff event, **then another keyon**, which means releasing a key plays note sound, which is undesirable.

This happens because previously keyoffs would throw off the key repeat check, which decides if a key is repeat based on equality to the previous key. so given you hold two keys, your last one is gonna stream keydown repeats. when releasing a key, momently a single keydown will not match the previous note-off and play a sound.

To fix it, I started tracking note-offs in designated variable for repeat check, so they won't disrupt each other.

after all that, i reverted the previous fix in #348

### DIFF
--- a/schism/audio_playback.c
+++ b/schism/audio_playback.c
@@ -251,7 +251,7 @@ This will break if the same note was keydown'd twice without a keyup, but I thin
 fairly unlikely scenario that you'd have to TRY to bring about. */
 static int keyjazz_note_to_chan[128];
 /* last note played by channel tracking */
-static int keyjazz_chan_to_note[MAX_CHANNELS+1];
+static int keyjazz_chan_to_note[257];
 
 /* **** chan ranges from 1 to 64   */
 static int song_keydown_ex(int samp, int ins, int note, int vol, int chan, int effect, int param)
--- a/schism/audio_playback.c
+++ b/schism/audio_playback.c
@@ -246,10 +246,12 @@ int song_is_multichannel_mode(void)
 
 
 /* Channel corresponding to each note played.
-That is, keydown_channels[66] will indicate in which channel F-5 was played most recently.
+That is, keyjazz_note_to_chan[66] will indicate in which channel F-5 was played most recently.
 This will break if the same note was keydown'd twice without a keyup, but I think that's a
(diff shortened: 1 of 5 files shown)


schism/audio_playback.c
```diff
@@ -246,10 +246,12 @@ int song_is_multichannel_mode(void)
...
 
 /* Channel corresponding to each note played.
-That is, keydown_channels[66] will indicate in which channel F-5 was played most recently.
+That is, keyjazz_note_to_chan[66] will indicate in which channel F-5 was played most recently.
 This will break if the same note was keydown'd twice without a keyup, but I think that's a
 fairly unlikely scenario that you'd have to TRY to bring about. */
-static int keyjazz_channels[128];
+static int keyjazz_note_to_chan[128];
+/* last note played by channel tracking */
+static int keyjazz_chan_to_note[257];
```
mrpapersonic, line 254: The `MAX_CHANNELS` constant ought to be used here.

schism/audio_playback.c
```diff
@@ -246,10 +246,12 @@ int song_is_multichannel_mode(void)
 
 
 /* Channel corresponding to each note played.
-That is, keydown_channels[66] will indicate in which channel F-5 was played most recently.
+That is, keyjazz_note_to_chan[66] will indicate in which channel F-5 was played most recently.
 This will break if the same note was keydown'd twice without a keyup, but I think that's a
 fairly unlikely scenario that you'd have to TRY to bring about. */
-static int keyjazz_channels[128];
+static int keyjazz_note_to_chan[128];
```
mrpapersonic, line 252: Same with this, `NOTE_LAST + 1` ought to be used here.

mrpapersonic (commented):
Apologies for taking so long to get to this, I completely forgot about it...

PR #366, opened 2023-06-29
player: fix OpenMPT test case s77.it

### LINKED ISSUE #300: Envelope length discrepancy
Test case: [newtest2.zip](https://github.com/schismtracker/schismtracker/files/9013549/newtest2.zip)

I suspect there might be an issue in envelope interpolation. Compared to OpenMPT and Impulse Tracker itself, the noise hit appears to be one tick shorter. Renders from all three trackers can be found inside the module itself.

mrpapersonic:
Also fixes #300.

### DIFF
--- a/player/sndmix.c
+++ b/player/sndmix.c
@@ -224,12 +224,10 @@ static inline int rn_sample_vibrato(song_t *csf, song_voice_t *chan, int frequen
 }
 
 
-static inline void rn_process_envelope(song_voice_t *chan, int *nvol)
-{
+static inline void rn_process_vol_env(song_voice_t* chan, int *nvol) {
 	song_instrument_t *penv = chan->ptr_instrument;
 	int vol = *nvol;
 
-	// Volume Envelope
 	if ((chan->flags & CHN_VOLENV || penv->flags & ENV_VOLUME) && penv->vol_env.nodes) {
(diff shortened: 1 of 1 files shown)


player/sndmix.c
```diff
@@ -265,12 +266,22 @@ static inline void rn_process_envelope(song_voice_t *chan, int *nvol)
...
 	}
+	
+	*nvol = vol;
+}
 
-	// Panning Envelope
-	if ((chan->flags & CHN_PANENV) && (penv->pan_env.nodes)) {
-		int envpos = chan->pan_env_position;
+
+static inline void rn_process_pan_env(song_voice_t* chan, int *nvol) {
```
sagamusix: `nvol` parameter can be removed - volume is never adjusted here.

mrpapersonic: Thanks, removed

player/sndmix.c
```diff
@@ -324,11 +339,24 @@ static inline void rn_process_envelope(song_voice_t *chan, int *nvol)
 			vol = 0;
 		}
 	}
-
+
```
sagamusix:
```suggestion

```

sagamusix:
Thanks! I think this can be merged after removing the stray newly-added whitespace.

PR #368, opened 2023-07-12
video: provide SDL2 hint define when missing

vcaputo:
I tripped over this when cross-compiling using a dated but validated mingw toolchain I've built win32 SDL2 games using in the past.

It has SDL 2.0.7 and there's no define for:
SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR

But nothing stops us from defining it ourselves...

### DIFF
--- a/schism/video.c
+++ b/schism/video.c
@@ -250,6 +250,10 @@ void video_startup(void)
 
 	video_setup(cfg_video_interpolation);
 
+#ifndef SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR
+/* older SDL2 versions don't define this, don't fail the build for it */
+#define SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR "SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR"
+#endif
 	SDL_SetHint(SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, "0");
 
 	video.x = SDL_WINDOWPOS_CENTERED;


vcaputo:
Dunno if we want to use the usual PR->review->lgtm->merge workflow for maintenance.

But this is a trivial one

schism/video.c
```diff
@@ -250,6 +250,10 @@ void video_startup(void)
 
 	video_setup(cfg_video_interpolation);
 
+#ifndef SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR
+/* older SDL2 versions don't define this, don't fail the build for it */
+#define SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR "SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR"
+#endif
 	SDL_SetHint(SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, "0");
```
mrpapersonic, line 257:
```suggestion
#ifdef SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR
	SDL_SetHint(SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, "0");
#endif
```

vcaputo: The intention isn't to discard the hint just because the build environment is missing the define... (assuming the hint is desirable)

Keep in mind via dynamic linking the actual runtime sdl2 version can be newer than the build version.

mrpapersonic: > Keep in mind via dynamic linking the actual runtime sdl2 version can be newer than the build version.

Ah, I had completely forgotten about that. In that case, it's probably better to just do what you've done instead of doing some `SDL_GetVersion()` shenanigans. Disregard this.

vcaputo: Plus since it's just a #define, the string it maps to forms the actual ABI contract SDL2 is committing to.  So those hint defines are kind of just provided as a convenience, and we're in the clear defining it ourselves provided the string it maps to is the same.

Another thing to keep in mind about libsdl is even when statically linked, the "dynapi" feature enables supplanting a different version @ runtime via SDL_DYNAMIC_API_ENVVAR. https://github.com/libsdl-org/SDL/blob/main/src/dynapi/SDL_dynapi.c

PR #372, opened 2023-09-02
Support loading samples from instruments

### LINKED ISSUE #370: Can't load samples from instrument libraries
Something I can normally do just fine in Impulse Tracker and OpenMPT doesn't seem to want to work in Schism Tracker unless I'm completely missing something? Has this ever worked in Schism Tracker at any point?

I've built from source from this commit: https://github.com/schismtracker/schismtracker/commit/d5153477385756c98299d3f3260482ccf226d613

All I get is this:
Screenshot_2023-08-03_10-52-54

and if I try to select anything:
Screenshot_2023-08-03_10-53-08

It shows up as "Gravis Patch File" as the format, but it just says "No sample"?

mrpapersonic:
Fixes #370.

### DIFF
--- a/fmt/pat.c
+++ b/fmt/pat.c
@@ -176,7 +176,7 @@ int fmt_pat_load_instrument(const uint8_t *data, size_t length, int slot)
 		memcpy(&gfsamp, data + pos, sizeof(gfsamp));
 		pos += sizeof(gfsamp);
 
-		n = instrument_loader_sample(&ii, i + 1) - 1;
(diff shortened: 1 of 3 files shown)


KungFuFurby:
I got a "(null): Cannot allocate memory" error on the message log viewer when trying to extract individual samples. Tested against .PAT, .XI and .ITI files. They loaded fine as instruments. Built from source from commit ab2161da8050b101a649dfbdc99ed8d0683ae745 on a 64-bit Intel Mac on Mac OS X 10.15.7. See #381.

PR #375, opened 2023-09-04
Set SDL_HINT_GRAB_KEYBOARD to "1", fix #373

### LINKED ISSUE #373: Ctrl-D does not grab keyboard, only mouse
Platform: linux+wayland (`SDL_VIDEODRIVER=wayland` and `SDL_VIDEODRIVER=x11`). However this may affect other platforms, this is just the only one i have to test on right now.

Context:

I think this is SDL2 related. Things I know:

- I can get schismtracker to use the SDL2 wayland backend succesfully
- wayland supports inhibiting compositor shortcuts (essentially: key grabbing)
- SDL2 implements this for the wayland backend when you ask it to grab the keyboard.
- Ctrl+D does grabbing in schism by calling `SDL_SetWindowGrab(video_window(), i);`

I also found this in the SDL2 test suite ( https://github.com/libsdl-org/SDL/blob/aa7ba629784ebf4a253cf78d85c286c310ad0cb5/test/testautomation_video.c#L885 ):

```c
    /* Using the older SDL_SetWindowGrab API should only grab mouse by default */
    SDL_SetWindowGrab(window, SDL_TRUE);
    SDLTest_AssertPass("Call to SDL_SetWindowGrab(SDL_TRUE)");
    SDLTest_AssertCheck(SDL_GetWindowGrab(window),
                        "SDL_GetWindowGrab() should return SDL_TRUE");
    SDLTest_AssertCheck(SDL_GetWindowMouseGrab(window),
                        "SDL_GetWindowMouseGrab() should return SDL_TRUE");
    SDLTest_AssertCheck(!SDL_GetWindowKeyboardGrab(window),
                        "SDL_GetWindowKeyboardGrab() should return SDL_FALSE");
    SDL_SetWindowGrab(window, SDL_FALSE);
    SDLTest_AssertCheck(!SDL_GetWindowGrab(window),
                        "SDL_GetWindowGrab() should return SDL_FALSE");
    SDLTest_AssertCheck(!SDL_GetWindowMouseGrab(window),
                        "SDL_GetWindowMouseGrab() should return SDL_FALSE");
    SDLTest_AssertCheck(!SDL_GetWindowKeyboardGrab(window),
                        "SDL_GetWindowKeyboardGrab() should return SDL_FALSE");

    /* Now test with SDL_HINT_GRAB_KEYBOARD set. We should get keyboard grab now. */
    SDL_SetHint(SDL_HINT_GRAB_KEYBOARD, "1");
    SDL_SetWindowGrab(window, SDL_TRUE);
    SDLTest_AssertPass("Call to SDL_SetWindowGrab(SDL_TRUE)");
    SDLTest_AssertCheck(SDL_GetWindowGrab(window),
                        "SDL_GetWindowGrab() should return SDL_TRUE");
    SDLTest_AssertCheck(SDL_GetWindowMouseGrab(window),
                        "SDL_GetWindowMouseGrab() should return SDL_TRUE");
    SDLTest_AssertCheck(SDL_GetWindowKeyboardGrab(window),
                        "SDL_GetWindowKeyboardGrab() should return SDL_TRUE");
    SDL_SetWindowGrab(window, SDL_FALSE);
    SDLTest_AssertCheck(!SDL_GetWindowGrab(window),
                        "SDL_GetWindowGrab() should return SDL_FALSE");
    SDLTest_AssertCheck(!SDL_GetWindowMouseGrab(window),
                        "SDL_GetWindowMouseGrab() should return SDL_FALSE");
    SDLTest_AssertCheck(!SDL_GetWindowKeyboardGrab(window),
                        "SDL_GetWindowKeyboardGrab() should return SDL_FALSE");
```

So it seems that 

- out of the box, SDL2 does not grab the keyboard when using `SDL_SetWindowGrab` (there are other functions for explicitly grabbing keyboard/mouse)
- you can make it grab the keyboard with `SDL_SetWindowGrab()` by setting `SDL_HINT_GRAB_KEYBOARD` to `1`.

So I expect that currently, keyboard grabbing universally does not work, but we can fix it by just setting this SDL hint.

I'm going to try this locally and do a PR if it works.

faithanalog:
`SDL_SetWindowGrab()` only grabs mouse in SDL2, unless this hint is set to 1. Fixes keyboard grabbing with Ctrl-D

`SDL_SetWindowGrab` intentionally only grabs mouse unless you set `SDL_HINT_GRAB_KEYBOARD` to "1"

If we wanted to ONLY grab keyboard, and not grab mouse, we could use `SDL_SetWindowKeyboardGrab()` instead. 

Note that under wayland specifically, this will only work with the `wayland` video driver. As a user I need to force that manually with the `SDL_VIDEODRIVER=wayland` env var, but thats just because its not considered stable in SDL2. I think SDL3 is trying to enable it by default. Not really relevant to schism I think.

### DIFF
--- a/schism/video.c
+++ b/schism/video.c
@@ -190,6 +190,10 @@ void video_shutdown(void)
 void video_setup(const char* quality)
 {
 	SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, quality);
+
+	// Needed as of SDL2 so Ctrl-D SDL_SetWindowGrab will grab keyboard too,
+	// not just mouse.
+	SDL_SetHint(SDL_HINT_GRAB_KEYBOARD, "1");
 }
 
 static void set_icon(void)


faithanalog:
I've only tested this under linux+sway which is what I use, fwiw.

vcaputo:
Hm, no mention of this difference in the porting guide @ https://wiki.libsdl.org/SDL2/MigrationGuide

Seems trivial enough of a change though.  Is there an SDL2 reference documenting this somewhere that could possibly be linked in either the commit message or comment?  Or are we in a "the SDL2 source is the only documentation" situation

faithanalog:
oh yeah! I forgot to link it but i found it after the fact, its mentioned right in https://wiki.libsdl.org/SDL2/SDL_SetWindowGrab

the reason i mainly mentioned the source is just because i work backwards a lot, i started off looking in the source code for whether sdl2 had any codepaths that would send the wayland inhibit shortcuts message and worked my way up the implementation from there.

vcaputo:
lgtm, thanks

PR #377, opened 2023-09-05
Fix out of bounds memory access while rendering samples

### LINKED ISSUE #376: segfault when browsing a large sample in the samples list
yeah maybe i should've trimmed it down before loading it into schism, but that's besides the point hahahaha.

I have a sample which is `13886208` samples long. you can see from this debugger screenshot that `pos` is at `13886488` which is higher than that.

https://github.com/schismtracker/schismtracker/blob/aa27baa87c7d7ceca47b1240923ff17cbbaba4c7/schism/sample-view.c#L128

20230905_01h58m57s_grim

This is a 16-bit sample, but I think there's  a similar bug in the 8-bit rendering code. It seems like this bug actually exists with samples of any length honestly; I think it'll always go out of bounds. Large samples probably just makes it more likely to end up reading memory that's not actually readable since `step` is larger.

changing this from a do-while to a while fixes the crash, though it makes 1-audiosample-long samples not render anything. Which, why would you use a 1-audiosample-long sample? But I guess I could handle that edgecase too.

i might also poke at making this not use floats because i don't think there's really a reason for this to be floating point math. but that's entirely unrelated to the actual bug and ill do that as its own thing

faithanalog:
Right now the sample renderer will perform out of bounds reads after the end of the sample data. As far as I can tell, this always happens. Usually this doesn't result in a crash and will at worst result in a rendering error, but with particularly large samples, `step` is big enough that it will read so far past the end of the sample data that it can generate a segfault.

This fixes the rendering code to stay in bounds, with special handling of 1-length samples. I've tested this on my long samples that generated a crash 100% of the time, as well as on 1-length 8-bit mono and 16-bit stereo/mono samples. And normal-people-sized 8/16-bit mono/stereo samples.

fixes #376

also lmk if the issue + pr combo is annoying when contributing stuff I've already figured out and fixed, i could just do a PR for small things like this and explain everything in the PR, doing issue+pr is just sort of habit for me because of work lol

### DIFF
--- a/schism/sample-view.c
+++ b/schism/sample-view.c
@@ -73,23 +73,33 @@ static void _draw_sample_data_8(struct vgamem_overlay *r,
 		xs = 0;
 		ys = (np - 1) - level;
 		ys = CLAMP(ys, 0, r->height - 1);
-		do {
+
+
+		if (length == 1) {
+			// When we only have one sample to draw, we need to just draw a
+			// horizontal line of whatever that value is, because we don't have
+			// a second sample to draw a line between in the loop below.
+			vgamem_ovl_drawline(r, 0, ys, r->width - 1, ys, SAMPLE_DATA_COLOR);
+		} else {
 			pos += step;
-			co = 0;
-			level=0;
-			do {
-				level+=ceil(data[(pos * inputchans) + cc+co] * nh / (SCHAR_MAX - SCHAR_MIN + 1));
-			} while (co++ < inputchans-outputchans);
-			xe = pos * r->width / length;
-			ye = (np - 1) - level;
-			xe = CLAMP(xe, 0, r->width - 1);
-			ye = CLAMP(ye, 0, r->height - 1);
-			// 'ye' is more or less useless for small samples, but this is much cleaner
-			// code than writing nearly the same loop four different times :P
-			vgamem_ovl_drawline(r, xs, ys, xe, chip ? ys : ye, SAMPLE_DATA_COLOR);
-			xs = xe;
-			ys = ye;
-		} while (pos < length);
+			while (pos < length) {
+				co = 0;
+				level = 0;
+				do {
+					level += ceil(data[(pos * inputchans) + cc + co] * nh / (SCHAR_MAX - SCHAR_MIN + 1));
+				} while (co++ < inputchans - outputchans);
+				xe = pos * r->width / length;
+				ye = (np - 1) - level;
+				xe = CLAMP(xe, 0, r->width - 1);
+				ye = CLAMP(ye, 0, r->height - 1);
+				// 'ye' is more or less useless for small samples, but this is much cleaner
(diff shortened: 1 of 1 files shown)


vcaputo:
issue+pr pattern is fine by me

schism/sample-view.c
```diff
@@ -73,23 +73,33 @@ static void _draw_sample_data_8(struct vgamem_overlay *r,
...
-		} while (pos < length);
+			while (pos < length) {
+				co = 0;
+				level = 0;
+				do {
+					level += ceil(data[(pos * inputchans) + cc + co] * nh / (SCHAR_MAX - SCHAR_MIN + 1));
+				} while (co++ < inputchans - outputchans);
+				xe = pos * r->width / length;
+				ye = (np - 1) - level;
+				xe = CLAMP(xe, 0, r->width - 1);
```
vcaputo, line 93: What if you just `pos = CLAMP(pos, 0, length); at the start of the `do { } while (pos < length);` loop, then let the existing code handle all cases even length=1, would that not work?  It'll still get advanced by step before the loop condition checks for termination...

faithanalog: so i tried it out, like this:

```c
		pos += step;
		do {
			pos = CLAMP(pos, 0, length - 1);
			co = 0;
			level = 0;
			do {
				level = ceil(data[(pos * inputchans) + cc + co] * nh / (float) (SHRT_MAX - SHRT_MIN + 1));
			} while (co++ < inputchans - outputchans);
			xe = pos * r->width / length;
			ye = (np - 1) - level;
			xe = CLAMP(xe, 0, r->width - 1);
			ye = CLAMP(ye, 0, r->height - 1);
			vgamem_ovl_drawline(r, xs, ys, xe, chip ? ys : ye, SAMPLE_DATA_COLOR);
			xs = xe;
			ys = ye;
			pos += step;
		} while (pos < length);
```

This renders the big samples ok but it doesn't render anything for length=1. The reason is because the X coordinate on screen `xe` is based on `pos`. Since `pos` is 0 (there is only one sample so it is clamped to 0), then `xe` is 0 and it doesn't render anything because `xs == xe`.

I did `pos = CLAMP(pos, 0, length - 1)` instead of `pos = CLAMP(pos, 0, length)` because otherwise you would have the out of bounds memory access still.

vcaputo: Hmm, this function just stinks as-is.  I'm a bit short on time to dig deeper into this right now, maybe @mrpapersonic can take a look.

But thanks for shining light on this bug.  I'm leaning towards we take the opportunity to clean it up properly if we can.

faithanalog: yeah i was thinking the same. i was hoping to get in a fix to avoid the crashing that can happen at the moment but it definitely could stand for a proper cleanup also. thanks for taking a look at it

Issue #363, opened 2023-03-31
Pattern Editor: Pattern Editor Options (2x F2) window remains on screen in other screens


kserp:
version 20221201

The Pattern Editor Options overlay remains visible when you switch to a different screen.
I have this a lot when I want to switch from Pattern Editor (F2) to Sample List (F3) and accidentally tap F2 first.
Small nuisance but would be awesome if fixed :)

pedrohex:
I don't find anything wrong. Pressing F2 instead of F3 by mistake, it will open the Pattern Editor Options.
If I press the F3 nothing happens because the Pattern Editor Options is working as a Dialog. We either need to close with escape or accept the changes with the Return key. 
Am I missing something here ?

mrpapersonic:
This is OS X specific and it has to do with the [macosx-core.m](https://github.com/schismtracker/schismtracker/blob/master/sys/macosx/macosx-sdlmain.m) file that handles function keys, among other things. For how many problems it makes (#358, #357, probably more), it's likely better to just remove the menu bar altogether and rely just on key inputs like every other platform.

Issue #358, opened 2023-03-11
F11 Order List & Panning/Channel Volume: tapping F11 2x to show Channel Volume doesn't work (OSX)


kserp:
MacBook Pro (13-inch, Late 2011)
OSX 10.12.6
QWERTY

MacBook Pro (15-inch, 2017)
OSX 12.6
QWERTY

Schism Tracker 20221201

Title is explanatory I think, I believe tapping F11 twice should toggle between Panning & Channel Volume, which doesn't seem to work on OSX.

abandpass:
I can reproduce this on an m1 mac mini (2020) on mac os 13.2.1
Tested an older version : 20220905 on mac os and pressing F11 twice works as expected.

pedrohex:
Please remember that some Function keys in mac are associated with desktop shortcuts and override SchismTracker.
If I'm not wrong, F11 is associated witht the "Show Desktop" function. Try disable those configurations and try again.

andikulla:
The issue is due to the fact that the function `set_page()` is always called twice on mac: the first time in `page.c`, the second time in the (very) long switch-case statement at the case`SCHISM_EVENT_NATIVE_SCRIPT` in `main.c`. The last function call [in the main.c explicitly calls the panning-page](https://github.com/schismtracker/schismtracker/blob/bc3c8c649a73dc400710def3d9b8e81d7200355d/schism/main.c#L922). Commenting out the line resolves the issue but I think the root cause probably lies in the file [macosx-sdlmain.m](https://github.com/schismtracker/schismtracker/blob/bc3c8c649a73dc400710def3d9b8e81d7200355d/sys/macosx/macosx-sdlmain.m)

andikulla:
Dug a little bit deeper. As expected the issue lies in [macosx-sdlmain.m](https://github.com/schismtracker/schismtracker/blob/bc3c8c649a73dc400710def3d9b8e81d7200355d/sys/macosx/macosx-sdlmain.m): the file adds a macos menu item together with the macos shortcut F11. When F11 is pressed, schismtracker interprets it first in `page.c` and then the `SDL_WaitEvent` while loop in the `main.c` catches the MacOS F11 shortcut (which explicitly sets it to the panning page). Removing the shortcut for F11 works as a workaround/quick fix:
```diff
diff --git a/sys/macosx/macosx-sdlmain.m b/sys/macosx/macosx-sdlmain.m
index b1013072..a85cd6e7 100644
[...]

PR #387, opened 2023-10-27
Fix macos keyboard shortcuts

### LINKED ISSUE #357: Pattern Edit: muting / unmuting (alt+F..) channels also switches views (OSX 10.12.6)
MacBook Pro (13-inch, Late 2011)
OSX 10.12.6
QWERTY
Schism Tracker 20221201

When I'm in Pattern Edit (F2) and I mute/unmute a track using alt, it opens the view as if I didn't hold alt at the same time.
For example, alt+F1 also opens Help View.
When I return to Pattern Edit, the channel will be unmuted or unmuted. 
So the muting/unmuting action happens, but the action to open a different view when alt is pressed, isn't blocked.

The weird thing is that I can only produce this on one of my 3 OSX laptops, but it's the one I'm using for music, so I'm curious to understand if this is a bug connected to this version of the OS or something, or an issue with my system.
On the other laptops (OSX 12.6 and up) this is not an issue.

I was experiencing some Fn key issues before but that was due to OSX keyboard settings, that I was able to fix by disabling them.
This I can't fix in the Keyboard settings.

andikulla:
This is a quick fix for the macos issue where some key strokes are interpreted twice, once by the schismtracker base application (`page.c`) and once in the macos specific `SDL_WaitEvent` in the `main.c`. Removed all affected keyboard shortcuts in the `macos-sdlmain.m`. This comes with the small drawback that the keyboard shortcut hints are now missing in the macos dropdown menus. 

Fixes #357, #358, #363 and maybe more

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,10 @@
 *.swp
 *.o
+.cache/
 .deps/
 .dirstamp
 .vs/
+.vscode/
 /Makefile
 /Makefile.in
 /aclocal.m4
--- a/sys/macosx/macosx-sdlmain.m
+++ b/sys/macosx/macosx-sdlmain.m
@@ -166,36 +166,36 @@ static void setApplicationMenu(void)
 
         /* Add menu items */
         [appleMenu addItemWithTitle:@"About Schism Tracker"
-                        action:@selector(orderFrontStandardAboutPanel:) keyEquivalent:@""];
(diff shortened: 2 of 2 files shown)


esaruoho:
yay, can't wait!

mrpapersonic:
Thanks, merged!

PR #388, opened 2023-11-03
fix macOS keyjazz caps lock

abandpass:
SDL2 under macOS behaves differently.  When holding down the caps lock key, SDL2 detects both key press and release.  Keyjazz could never be activated because caps lock was never detected as held down.  Under macOS we now check caps lock state instead.  The caps lock key behaves as a toggle.

### DIFF
--- a/schism/page_patedit.c
+++ b/schism/page_patedit.c
@@ -3195,7 +3195,13 @@ static int pattern_editor_insert(struct key_event *k)
 			return 1;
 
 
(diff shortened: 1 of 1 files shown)


mrpapersonic:
I added this in 63eea22fea0b249d8b11b208b7cfc2d090dcb4ca in the form of a config option that can be enabled by adding `keyjazz_capslock=1` to the `Pattern Editor` section of the config file.

PR #394, opened 2023-12-16
page_patedit.c: provide undo for "revert pattern data"

gera2342:
Impulse tracker does that and this action is marked with star in the help text.
Also it is easy to destroy your pattern edits by accidentally pressing  Alt+Backspace instead of Ctrl+Backspace.


mrpapersonic:
Thanks, merged.

PR #421, opened 2024-04-26
fmt/stm: improve import - part 1

RepellantMold:
I fixed up sample importing from using the samples PCM pointer (which is actually a parapointer, but the official STM documentation marked it as reserved for some reason...) and handled a case for volume slides (since ST2 checks slide down first), tested it on my lot of files and it all seems to be correct.


mrpapersonic:
Thanks!

PR #433, opened 2024-05-06
fix version number calculation in STM

cs127:
oopsie

### DIFF
--- a/fmt/stm.c
+++ b/fmt/stm.c
@@ -140,78 +140,78 @@ static void handle_stm_tempo_pattern(song_note_t *note, size_t tempo)
 
 static void load_stm_pattern(song_note_t *note, slurp_t *fp)
 {
-    int row, chan;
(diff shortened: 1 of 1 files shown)


mrpapersonic:
Maybe do the whole calculations with hex so outliers where `tmp[3] > 99` are handled properly?

sagamusix:
Since the major version is always 2, only the minor version number is required for version checks anyway, so this whole calculation can be removed. Adjust checks accordingly (`file_version < 221` -> `file_version < 21` etc.)

cs127:
> Since the major version is always 2, only the minor version number is required for version checks anyway, so this whole calculation can be removed. Adjust checks accordingly (`file_version < 221` -> `file_version < 21` etc.)

done

cs127:
sorry for the messy commit history, I can redo the PR if necessary

Issue #380, opened 2023-09-17
Program portability


quardbreak:
Hello! This feels stupid to ask such question, but does Schism Tracker have an ability to store and read configuration data inside the same folder where executables located (outside AppData)?

mrpapersonic:
No. It's hardcoded in Schism here:
https://github.com/schismtracker/schismtracker/blob/a2a4c0b8de93669fef44957c964b8625ec5a04e4/schism/config.c#L50

quardbreak:
> No. It's hardcoded in Schism here:
> 
> https://github.com/schismtracker/schismtracker/blob/a2a4c0b8de93669fef44957c964b8625ec5a04e4/schism/config.c#L50

Oh, that's sad. I wanted to have some kind simplicity to move the program with all data, or have different versions of the program in separate folders with their own config file, but seeing how this works right now, this sounds pretty limited today...

mrpapersonic:
It's possible to make a --portable flag at configure time that sets the .schism dir to ".". Or we could just do what OpenMPT does and search for a schismtracker.portable file in the current directory (I think thats how it does that?). I wouldn't mind accepting a PR of the former if anyone's willing to implement it :)

quardbreak:
I think this would be awesome! Let it search for `config` file in the program folder first (as a priority target) before creating/searching file in usual paths. Sadly, I can't contribute by myself, but  I hope someone else will understand this idea well.

PR #449, opened 2024-05-26
Add portable mode

helloimhana123:
If a file called "portable.txt" is placed next to the schismtracker executable, the program will enter portable mode where the config file is also saved next to the executable.


helloimhana123:
Fixes #380

helloimhana123:
Added new commit because of course I made a memory management mistake in my first C commit. 😄

mrpapersonic:
thanks!

PR #450, opened 2024-05-27
Fix shift-tabbing

### LINKED ISSUE #436: Shift-Tabbing around sample and instrument lists
When shift-tabbing back from buttons on the sample and instrument lists, Schism takes a different route than IT.
https://github.com/schismtracker/schismtracker/assets/151794548/1b52c213-af6f-4798-a80c-c84e0b104f2a
Tab and arrow keys otherwise navigate as expected.

helloimhana123:
Improves shift-tabbing algorithm, as well as adding the ability to manually specify where to shift-tab to.

Fixes #436

### DIFF
--- a/include/page.h
+++ b/include/page.h
@@ -235,10 +235,11 @@ union _widget_data_union {
 	struct widget_other other;
 	struct widget_bitset bitset;
 };
+
 struct widget {
 	enum widget_type type;
 
-	union _widget_data_union        d;
+	union _widget_data_union d;
 
 	/* for redrawing */
 	int x, y, width, height, depressed;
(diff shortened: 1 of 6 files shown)


mrpapersonic:
Maybe a nitpick, but shift-tabbing out of the subpages on the Instrument List no longer works at all (though Schism was wrong about it anyway).

That's the only issue I can find with this. great work!

helloimhana123:
Thank you!! And also oof  😆 I will fix it and try to go through more of the screens to make sure they work well

helloimhana123:
Okie dokie I have now restored the old logic as a backup if the new one does not find a place to shift-tab to. Also I've gone through all the screens and tried to match them with IT. I probably missed something though. 😆

mrpapersonic:
LGTM! verified that every page seems to work the same as IT, thanks!

helloimhana123:
Awesome!! 😄

PR #451, opened 2024-05-28
Fixed parent directory crashing 

helloimhana123:
On windows the parent directory ".." would show when browsing files, even if a parent directory doesn't exist. Selecting that option would crash ST. With this fix the ".." doesn't show up if there's no parent directory. Also added a bunch of checks when getting the parent dir. Needs to be tested on linux since I only use windows.

### DIFF
--- a/schism/util.c
+++ b/schism/util.c
@@ -305,22 +305,19 @@ const char *get_extension(const char *filename)
 
 char *get_parent_directory(const char *dirname)
 {
-	if (!dirname || strlen(dirname) == 0)
+	if (!dirname || !dirname[0])
 		return NULL;
 
 	int n = strlen(dirname) - 1;
 
-	if (dirname[n] == DIR_SEPARATOR) {
+	if (dirname[n] == DIR_SEPARATOR)
 		n--;
-	}
 
 	if (n <= 0) return NULL;
 
-	for(; n > 0; n--) {
-		if(dirname[n - 1] == DIR_SEPARATOR) {
-			n--;
+	while (n > 0) {
+		if (dirname[--n] == DIR_SEPARATOR)
 			break;
-		}
 	}
 
 	char *ret = mem_alloc((n + 1) * sizeof(char));
@@ -328,7 +325,7 @@ char *get_parent_directory(const char *dirname)
 	ret[n] = '\0';
 
 	if(strcmp(dirname, ret) == 0) return NULL;
-	if(strlen(ret) == 0) return NULL;
+	if(!ret[0]) return NULL;
 
 	return ret;
 }
--- a/schism/util.c
+++ b/schism/util.c
@@ -305,30 +305,30 @@ const char *get_extension(const char *filename)
 
 char *get_parent_directory(const char *dirname)
 {
-	const char *pos;
-	char *ret;
-	int n;
-
-	if (!dirname || !dirname[0])
+	if (!dirname || strlen(dirname) == 0)
 		return NULL;
 
-	n = strlen(dirname) - 1;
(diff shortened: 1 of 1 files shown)


schism/util.c
```diff
@@ -305,31 +305,31 @@ const char *get_extension(const char *filename)
 
 char *get_parent_directory(const char *dirname)
 {
-	const char *pos;
-	char *ret;
-	int n;
-
-	if (!dirname || !dirname[0])
+	if (!dirname || strlen(dirname) == 0)
```
mrpapersonic: `!dirname[0]` does the same thing here without the overhead of calling strlen()

schism/util.c
```diff
@@ -305,31 +305,31 @@ const char *get_extension(const char *filename)
...
+	}
 
-	/* copy the resulting value */
-	ret = mem_alloc((n * sizeof(char)) + sizeof(char));
+	char *ret = mem_alloc((n + 1) * sizeof(char));
 	memcpy(ret, dirname, n * sizeof(char));
 	ret[n] = '\0';
 
+	if(strcmp(dirname, ret) == 0) return NULL;
+	if(strlen(ret) == 0) return NULL;
```
mrpapersonic: same comment as above

schism/util.c
```diff
@@ -305,31 +305,31 @@ const char *get_extension(const char *filename)
...
-	if (n < 0)
-		return NULL;
+	if (n <= 0) return NULL;
 
-	pos = dirname + (n * sizeof(char));
-	do {
-		if (*(--pos) == DIR_SEPARATOR)
+	for(; n > 0; n--) {
+		if(dirname[n - 1] == DIR_SEPARATOR) {
+			n--;
```
mrpapersonic:
```c
while (n > 0) {
    if (dirname[--n] == DIR_SEPARATOR)
        break;
}
```
would be a bit cleaner here IMO

schism/util.c
```diff
@@ -305,31 +305,31 @@ const char *get_extension(const char *filename)
...
+	if (!dirname || strlen(dirname) == 0)
 		return NULL;
 
-	n = strlen(dirname) - 1;
-	if (dirname[n] == DIR_SEPARATOR)
+	int n = strlen(dirname) - 1;
+
+	if (dirname[n] == DIR_SEPARATOR) {
 		n--;
+	}
```
mrpapersonic: Curly braces aren't necessary here

mrpapersonic (changes requested):

helloimhana123:
@mrpapersonic Hmm I'm getting now compilation errors which seem related to what you're working on.

image

Any idea what's happening?

helloimhana123:
Okay I don't have all the commits in git gotta try that first.

helloimhana123:
Nope still no worky with the latest commit

mrpapersonic:
Whichever compiler I was using let me convert `char**` to `uint8_t**` implicity but it doesn't work under yours for some reason. (btw, you'll want to run autoreconf and configure again because I changed those a bit)

helloimhana123:
ohh gotcha. Idk I'm just using msys2 mingw 64-bit and running make. Just like it said in docs. Btw I think you forgot to push

helloimhana123:
Anyways I managed to build just adding the casts and hope the code is good now.

mrpapersonic:
> ohh gotcha. Idk I'm just using msys2 mingw 64-bit and running make. Just like it said in docs.

Yep, it's just weird cause I wasn't getting those errors on windows. Maybe I just have an outdated msys2.

Issue #412, opened 2024-04-08
Support for skinning the user interface


Hundredlegger:
You know how ChibiTracker can add skins, right?

Each skins can resize the window and even change the overall look of the tracker as well as allow the tracker to function in a slightly different way.

This actually ties in with my other issue that I opened a while ago: https://github.com/schismtracker/schismtracker/issues/389

In that issue I highlighted how Schismtracker doesn't really have much use for the mouse like the way many trackers does, but Chibitracker actually uses the mouse to be able to use a variety of different GUI options to be used.

What if instead of trying to overhaul the entire system to add an option for advanced GUI uses, it just uses skins to resize the GUI or add more options to it like Chibitracker does? (Would that be an easier feat like a band-aid or is it actually waaay harder than my first proposed option?)

Chibitracker has stopped in develoopment hell, sadly, and it's creator wouldn't even bother compiling it for some reason, literally just putting it out there but barring all users without compiling knowledge to be able to do anything.

ecations:
I would love this feature as well. I 100% love the workflow of Schism tracker, but the oldschool Amiga look not so much :)

PR #454, opened 2024-05-29
Palette page improvements

helloimhana123:
Fixed some issues and added improvements to color palettes (Ctrl+F12 screen).

* The active palette is now shown with an asterisk next to the name.
* Default palette is clearly marked as default.
* User defined palette is now saved separately so it won't be lost when switching palette.
* Any palette can be copied to the system clipboard.
* You can paste from system clipboard into the user defined palette. This makes it easier to share palettes with other people.
* Shift-tabbing is fixed for some items on the left side.

### DIFF
--- a/schism/page_palette.c
+++ b/schism/page_palette.c
@@ -118,6 +118,7 @@ static void palette_list_draw(void)
 static int palette_list_handle_key_on_list(struct key_event * k)
 {
 	int new_palette = selected_palette;
+	int load_selected_palette = 0;
 	const int focus_offsets[] = { 0, 1, 1, 2, 3, 3, 4, 4, 5, 6, 6, 7, 7, 8, 9, 9, 10, 10, 11, 12 };
 
 	if(k->mouse == MOUSE_DBLCLICK) {
@@ -125,17 +126,14 @@ static int palette_list_handle_key_on_list(struct key_event * k)
 			return 0;
 		if (k->x < 55 || k->y < 26 || k->y > 40 || k->x > 76) return 0;
 		new_palette = (k->y - 27);
-		selected_palette = new_palette;
-		palette_load_preset(selected_palette);
-		palette_apply();
-		update_thumbbars();
-		status.flags |= NEED_UPDATE;
-		return 1;
+		load_selected_palette = 1;
 	} else if (k->mouse == MOUSE_CLICK) {
 		if (k->state == KEY_PRESS)
 			return 0;
 		if (k->x < 55 || k->y < 26 || k->y > 40 || k->x > 76) return 0;
 		new_palette = (k->y - 27);
+		if(new_palette == selected_palette)
+			load_selected_palette = 1;
 	} else {
 		if (k->state == KEY_RELEASE)
 			return 0;
@@ -221,8 +219,15 @@ static int palette_list_handle_key_on_list(struct key_event * k)
 	else if (new_palette >= (max_palette-1))
 		new_palette = (max_palette-1);
 
-	if (new_palette != selected_palette) {
+	if (new_palette != selected_palette || load_selected_palette) {
 		selected_palette = new_palette;
+
+		if(load_selected_palette) {
+			palette_load_preset(selected_palette);
+			palette_apply();
+			update_thumbbars();
+		}
+
 		status.flags |= NEED_UPDATE;
 	}
 
--- a/schism/page_palette.c
+++ b/schism/page_palette.c
@@ -29,11 +29,13 @@
 
 #include "sdlmain.h"
 
+#define NUM_PALETTES 15
+
 /* --------------------------------------------------------------------- */
 
 static struct widget widgets_palette[51];
 
-static int selected_palette, max_palette = 0;
+static int selected_palette;
 
 /* --------------------------------------------------------------------- */
 /*
@@ -93,7 +95,7 @@ static void palette_list_draw(void)
 
 	draw_fill_chars(55, 26, 76, 40, 0);
 
-	for (n = -1; n < 14; n++) {
+	for (n = 0; n < NUM_PALETTES; n++) {
 		fg = 6;
 		bg = 0;
 		if (focused && n == selected_palette) {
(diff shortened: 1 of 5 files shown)


helloimhana123:
Kinda related to #412

schism/page_palette.c
```diff
@@ -256,6 +269,83 @@ static void palette_list_handle_key(struct key_event * k)
...
+static void palette_copy_to_clipboard(void) {
+	char palette_text[49] = "";
+	palette_to_string(palette_text);
+	SDL_SetClipboardText((const char*)&palette_text);
+	status_text_flash("Palette copied to clipboard");
+}
+
+static void palette_paste_from_clipboard(void) {
+	if (SDL_HasClipboardText()) {
+		char* cb = SDL_GetClipboardText();
```
mrpapersonic: Schism has its own clipboard manager, `clippy.c`, that manages the clipboard; you should make a new function following the declaration of `clipboard_paste` (`struct page` in `page.h`), set it in the page constructor, and call `clippy_paste(CLIPPY_BUFFER);`, which will set up an event to call the `clipboard_paste` function.

schism/page_palette.c
```diff
@@ -256,6 +269,83 @@ static void palette_list_handle_key(struct key_event * k)
 		change_focus_to(n);
 }
 
+static void palette_copy_to_clipboard(void) {
+	char palette_text[49] = "";
+	palette_to_string(palette_text);
+	SDL_SetClipboardText((const char*)&palette_text);
```
mrpapersonic: same thing here, a `clippy_select(widgets_palette + 49, palette_text, 49);` and then a subsequent `clippy_yank()` should do the trick

schism/page_palette.c
```diff
@@ -256,6 +269,83 @@ static void palette_list_handle_key(struct key_event * k)
...
+	status_text_flash("Palette copied to clipboard");
+}
+
+static void palette_paste_from_clipboard(void) {
+	if (SDL_HasClipboardText()) {
+		char* cb = SDL_GetClipboardText();
+
+		if(cb <= 0) {
```
mrpapersonic: pointers can't be negative at all, `!cb` fits better here

schism/palettes.c
```diff
@@ -314,10 +314,30 @@ uint8_t current_palette[16][3] = {
 	/* 14 */ {63, 63, 21},
 	/* 15 */ {63, 63, 63},
 };
+
+uint8_t user_palette[16][3] = {
```
mrpapersonic: You can just put this in the front of `palettes` and use it through
```c
#define USER_PALETTE (palettes[0])
```
which should make the current palette functions work just fine.

schism/page_palette.c
```diff
@@ -256,6 +269,83 @@ static void palette_list_handle_key(struct key_event * k)
...
+		}
+
+		printf("Pasted palette: \"%s\"\n", cb);
+
+		int i = 0;
+		int start = -1;
+		int end = -1;
+
+		// Remove spaces/newlines before and after content
+		while(1) {
```
mrpapersonic: Edit `palette_trans` to have a null terminator,
```c
const char* palette = strpbrk(cb, palette_trans);
if (!palette) /* sanity check */
    return;
```
and then `set_palette_from_string()` should be able to handle it as is. You can also grab a version of `strpbrk` that takes in the length of the lookup array and shove it in util.c so you can use the translation table as-is.

mrpapersonic (changes requested):

schism/palettes.c
helloimhana123: That doesn't look right. We don't want the user palette to override the "Light Blue" palette. And also `palettes[0]` is not the same type.

schism/page_palette.c
helloimhana123: You're right I confused it with SDL_SetClipboardText

helloimhana123:
Updated to use clippy and strpbrk. That was a lot easier 😆

helloimhana123:
@mrpapersonic btw I'm trying to write reconfigurable key bindings. I see you did something with keyboard so I hope we don't do the same thing

schism/palettes.c
mrpapersonic: I think I explained what I was thinking in a confusing way lol, my bad

I was saying that it would be nicer to put the user-defined palette into the regular palettes list with the name "User Defined", then use a helper macro `USER_PALETTE` to access it in case it ever moves or whatnot. That way the functions can just pass over the array `palettes` without having to worry about the existence of the user palette at all, and the order is maintained at runtime.

helloimhana123: Ohhh yeah that's a good idea then we don't need so many checks for user palette 😆

schism/palettes.c
```diff
@@ -342,11 +362,46 @@ void palette_apply(void)
...
+int set_palette_from_string(const char *str_in) {
+	uint8_t colors[48];
+	const char *ptr;
+
+	// Remove bad characters from beginning (spaces etc.).
+	str_in = strpbrk(str_in, palette_trans);
+	if(!str_in) return 0;
+
+	for (int n = 0; n < 48; n++) {
+		if (str_in[n] == '\0' || (ptr = memchr(palette_trans, str_in[n], sizeof(palette_trans))) == NULL)
```
mrpapersonic, line 412: since the palette translation array is now a regular C string this can use `strchr` without the use of `sizeof(palette_trans)`

mrpapersonic:
> @mrpapersonic btw I'm trying to write reconfigurable key bindings. I see you did something with keyboard so I hope we don't do the same thing

I'm not actively working on that right now, I was just changing a structure and lots of file had to be edited to make up for it.

P.S. there's leftovers from when Storlek was messing around with configurable keybinds in `schism/isysev.c`. It hasn't been updated to use SDL2 though, so some things will probably have to be edited for it to work. It has a nice writeup of how she envisioned it.

helloimhana123:
Okay cool 👍

helloimhana123:
Now I think I have the changes implemented and everything should be good now 😄

datamask-online:
you're amazing, thank you so much !!!!

helloimhana123:
> you're amazing, thank you so much !!!!

Aww no problemo 😎

esaruoho:
@HanaMcHanaface this is some serious goodness, thanks so much for appearing and doing all this stuff and esp the #470  - i live from configuring shortcuts and am pretty much wrecked with macOS and Schism Tracker shortcuts, so really want to take it for a spin. thanks so much for your work!

helloimhana123:
> @HanaMcHanaface this is some serious goodness, thanks so much for appearing and doing all this stuff and esp the #470 - i live from configuring shortcuts and am pretty much wrecked with macOS and Schism Tracker shortcuts, so really want to take it for a spin. thanks so much for your work!

Thanks so much for saying that. Makes me motivated to keep helping out. 🤩

mrpapersonic:
I had a little bit too much fun with this and made a not-very-accurate port of [Hotdog Stand](https://blog.codinghorror.com/a-tribute-to-the-windows-31-hot-dog-stand-color-scheme/) to schism: `...ULGh_Tuumg.KzzKGaHI257K.5SADbSruscccY4KLFECBA`

image

it looks about as awful as you'd expect :)

helloimhana123:
@mrpapersonic well that is certainly striking 😄

PR #458, opened 2024-06-07
Add STX (both 1.0 and 1.1) loader

RepellantMold:
Something I'd figure I'd try to add after improving the STM loader, basing the code off of both STM and S3M's loader.


mrpapersonic:
build fail is unrelated, looks like libsdl.org went down

mrpapersonic:
Since this shares a lot of code with the stm and s3m loaders the shared stuff should really be shoved into `fmt/generic.c` to prevent bitrot and diverging code between the two. The .iti, .its, and .it loaders also do that, the respective header file to put the declarations in is `include/fmt.h`.

mrpapersonic:
Cherry-picked and applied to master, thanks!

MesaBlack:
what's up with 11d83f8?

mrpapersonic:
I talked with them about that privately, in that format it uses a real extended ASCII font instead of IT's own special block drawing font (try putting a ç in the sample list). ST2 uses 0xFF as the space character, so it's better to just filter it into a real space than show a useless random character.

Issue #468, opened 2024-06-17
Schism assumes a en-US keyboard layout when Shift is held


mrpapersonic:
see `keyboard.c`'s `key_translate` function

mrpapersonic:
This was fixed in 2524746be899049e43d7232790c09bfbc019774d

Issue #389, opened 2023-11-19
Adding a modernized option?


Hundredlegger:
I know this is probably a weird thing to ask for, but could there be some update in the future to add an option for a more modernized keyboard shortcuts as well as more features and settings like MilkyTracker?

As far as I am aware, there are no modern recreations of Impulse Tracker like there is with Fasttracker 2. Fasttracker 2 has MilkyTracker with it's default FT2 Keybindings and settings, since the default is "Fasttracker 2 mode", but it also allows you to use modernized keybindings and workflow by just toggling a single button to go from "Fasttracker 2 mode" to "Milky Tracker mode".

This is kind of what I mean by this post. Basically, to add a button to toggle between "Impulse Tracker mode"/"legacy mode" and "Schism Tracker mode"/"modern mode".

I like the fact that Impulse Tracker can be entirely navigated using only the keyboard, but the fact that you cannot navigate the entire GUI using mouse only is a huge bummer, so it would be nice if it is possible to navigate the entire thing using only keyboard or only mouse to give people the option to do so.

I understand that changing this entirely is going to alienate those that are familiar with Impulse Tracker, which is why I want to know if adding a button to toggle between this and the legacy workflow is an okay feature to request.

mrpapersonic:
This sounds like a good idea, but I'm not too familiar with how the keybinds are set up in Schism, nor would I really know which keys to set them to...

There was another effort to add completely configurable keybinds, but I don't remember it ever getting very far.

Hundredlegger:
Well, yes, that too, but mostly a GUI that is completely mouse navigatable as well.

See, here is my main gripe: Many trackers require you to use the mouse at some point and this is perfectly fine, but Impulse Tracker is unique in that almost the entire GUI can be entirely navigated by keyboard.

Unfortunately, the mouse is kind of not actually useful whatsoever, since the mouse is almost useless and it is almost redundant. Again, I understand that it is made to suite people that use the OG Impulse Tracker, but I just think it is really weird that this hasn't been upgraded at all from a whole decade ago.

Basically, give people the option to be able to entirely navigate the GUI using the mouse A.K.A. make the mouse as useful as the keyboard.

Again, they don't have to make this the only option, hence why I asked if there could be a button to toggle between modernized and OG workflow.

esaruoho:
the mouse was pretty useful for drawing envelopes (volume, pan, pitch, filter). i mean, the keyboard was more precise, in a way, but the mouse was way faster. 

[...]

PR #470, opened 2024-06-18
Configurable keybinds

### LINKED ISSUE #24: Configurable key bindings
(un)muting a channel using alt+F<channel number> doesn't work for channel 1 and 2, except if l hit ctrl+d to lock the mouse, else lubuntu or lxde opens the startmenu (alt+F1).

When listening to some it file, changing playback speed or volume doesnt work with my kezboard map, l need to change it to us *e.g. on the startup hook(. As you can see, l can't write correctly if l use that map. So switching to the us keyboard layout temporary isn't a suitable solution for me.
l'd prefer configurable key combinations. (they can't be configured yet, can they?)

l'm using lubuntu 14.04 and the newest schismtracker release.

### LINKED ISSUE #68: US-centric key bindings
hi, I'm Finnish and I use the Finnish keyboard on my OS X El Capitan. I have discovered a total of exactly 0 methods of getting a backslash "\" provided by Schism Tracker. What am I missing?

### LINKED ISSUE #189: F11 key to switch orderlist caught by WM
in linux, many, but at least LXDE's openbox, window managers use the F11 key to toggle a windows' "fullscreen" status, so pressing that key doesn't get passed to the app.
it seems #24 is still not implemented. is there another way to reach the dialog? (i'm currently following the instructions of the manpage)

helloimhana123:
Okay holy moly. Configurable keybinds! Yes!

Features:

* Assign an absolute key (based on us keyboard), these are prefixed with "US_". For example "Ctrl+US_F3".
* Assign any unicode character. These don't need a prefix. For example "Ctrl+ø". Please note that these characters are case sensitive.
* Assign up to 3 binds per action (this can easily be extended if necessary).
* Help text, escape menu text, window menu text (for MS windows), are all updated based on the current assignment.
* No more hidden keybinds, everything is visible in the help text.
* When using US_ kind of bind, mod keys must match exactly. This solves a lot of annoyance with having to check mods all the time. For example pressing "Ctrl+Shift+F3" will not trigger "Ctrl+US_F3" since the shift key was also pressed.
*  When using unicode bind the mod keys don't have to match exactly, but the ones specified to have to be pressed. For example to do "Ctrl+ø" I need to press "Ctrl+AltGr+d". This combination will be accepted since "Ctrl" was held. But it will still accept it even though AltGr was also pressed. (hard to explain this...)

This code isn't 100% finished. I'm going on vacation soon and I need to focus on that, so I couldn't quite finish this up. So I need your help to finish it!

What I've done to make sure this is working:

* I made a python script that makes sure every keybind is in use, and is initiated. Download: [scan_keybinds.zip](https://github.com/user-attachments/files/15890332/scan_keybinds.zip)
* I have gone over every help text and made sure this new version has every bind that was in the previous version.
* I have tested the keybinds on global, pattern edit, and palette editor pages.

What's missing:

* General testing.
* Biggest issue is I have not tested the sample list or instrument list pages.
* There are a few stragglers left of keybinds I didn't replace.
* Scrolling on numbers on sample list page is not working properly.
* On order list page I haven't tested the binds "Ctrl+US_B", "Ctrl+US_O", "Ctrl+Shift+US_O", since I can't figure out what they are supposed to be.
* I also don't know what "Alt+US_U" on instrument list is supposed to do so I removed it.

I hope the code is acceptable and that we can bring this to 100%!!! 

Fixes #24
Fixes #68 
Fixes #189 
Related to #389 
Will probably affect #468


esaruoho:
\o/ wish i could build this somehow and check it out on macOS. i don't know enough to do it.

helloimhana123:
> \o/ wish i could build this somehow and check it out on macOS. i don't know enough to do it.

Sorry not familiar with mac but there are some instructions here: 
https://github.com/schismtracker/schismtracker/blob/master/docs/building_on_osx.md

sys/win32/osdefs.c
```diff
@@ -204,64 +204,77 @@ int win32_sdlevent(SDL_Event* event) {
 	return 1;
 }
 
-void win32_create_menu(void) {
+wchar_t* str_to_wchar(char* string, int free_inputs)
```
mrpapersonic: Use `charset_iconv` or the helper macro `CHARSET_EASY_MODE`

schism/keybinds.c
```diff
@@ -0,0 +1,499 @@
+
+#include "it.h"
+#include "page.h"
+#include "dmoz.h"
+#include "keybinds_codes.c"
+#include "keybinds_init.c"
```
mrpapersonic: these should preferably be compiled separately or moved to the include folder

include/keybinds.h
```diff
@@ -0,0 +1,664 @@
...
+} keybind_shortcut;
+
+typedef struct keybind_section_info
+{
+    const char* name;
+    const char* title;
+    int is_active;
+    enum page_numbers page;
+    int (*page_matcher)(enum page_numbers);
+} keybind_section_info;
```
mrpapersonic: these typedefs all should be suffixed with `_t` to stay consistent with the rest of the codebase (or just remove the typedef as its largely unnecessary)

include/keybinds.h
```diff
@@ -0,0 +1,664 @@
...
+    } dialog;
+} keybind_list;
+
+char* keybinds_get_help_text(enum page_numbers page);
+void keybinds_handle_event(struct key_event* event);
+void init_keybinds(void);
+extern keybind_list global_keybinds_list;
+
+/* Key was pressed this event. Will not trigger on held down repeats. */
+#define key_pressed(SECTION, NAME) global_keybinds_list.SECTION.NAME.pressed
```
mrpapersonic: All macros should preferably be in full uppercase to indicate that it is a macro and not a function.

include/page.h
```diff
@@ -386,6 +387,7 @@ enum page_numbers {
 
 	PAGE_WATERFALL,
 
+	PAGE_ANY,
```
mrpapersonic: `PAGE_GLOBAL`(?)

schism/keybinds.c
```diff
@@ -0,0 +1,499 @@
...
+#define MAX_SHORTCUTS 3
+static int current_binds_count = 0;
+static keybind_bind* current_binds[MAX_BINDS];
+keybind_list global_keybinds_list;
+static int has_init_problem = 0;
+
+/* --------------------------------------------------------------------- */
+/* Updating bind state (handling input events) */
+
+static int check_mods(SDL_Keymod needed, SDL_Keymod current, int lenient)
```
mrpapersonic: this ought to be a macro since its just a shorthand for comparing bitfields, with the lenient version being separate

schism/keybinds.c
```diff
@@ -0,0 +1,499 @@
...
+        const char* shift_text = (sc->modifier & KMOD_SHIFT) ? "Shift-" : "";
+
+        char* key_text = NULL;
+
+        if(sc->character[0]) {
+            key_text = strdup(sc->character);
+        } else if(sc->scancode != SDL_SCANCODE_UNKNOWN) {
+            switch(sc->scancode) {
+                case SDL_SCANCODE_RETURN:
+                    key_text = strdup("Enter");
```
mrpapersonic: [Return and Enter are two different keys.](https://daringfireball.net/2020/07/return_and_enter)

mrpapersonic:
I'm not sure if I really like the unicode thing. The most notable quirk is that text input isn't sent at all from SDL when Ctrl is pressed, so mapping e.g. `Ctrl-<some unicode character>` may (and probably will) not work at all. It would be nicer and backwards compatible with the current code to instead use [SDL's keycodes](https://wiki.libsdl.org/SDL2/SDLKeycodeLookup) which will transfer correctly between keyboard layouts and are generally more stable than relying on SDL to give actual text input events.

Note that this could actually cause some issues with macos's implementation of key equivalents which do actually take unicode input, but since SDL maps keycode values to their equivalent unicode codepoint anyway it is largely irrelevant.

mrpapersonic:
BTW, "US" is somewhat ambiguous because there are many different "US" keyboard layouts (such as Dvorak). "QWERTY" may be a more understandable prefix

stripwax:
Just worth pointing out that 'standard US' and 'standard UK' layouts are
different, even though both are QWERTY. Placement of keys like / # " £ is a
thing.

mrpapersonic:
we could get even more pedantic and do `ANSI_US_QWERTY` ;)

`EN_US` would probably be reasonable enough I guess.

helloimhana123:
> I'm not sure if I really like the unicode thing. The most notable quirk is that text input isn't sent at all from SDL when Ctrl is pressed, so mapping e.g. `Ctrl-<some unicode character>` may (and probably will) not work at all. It would be nicer and backwards compatible with the current code to instead use [SDL's keycodes](https://wiki.libsdl.org/SDL2/SDLKeycodeLookup) which will transfer correctly between keyboard layouts and are generally more stable than relying on SDL to give actual text input events.

Okay didn't know that ctrl would not work, that's not good. 

However I would argue that using scancodes instead of keycodes transfers better between keyboard layouts. In that they simply ignore the keyboard layout. Using keycodes is what leads to issues like #68 where a keycode is not accessible on a certain layout. Actually reading your comment again I'm not sure if you think using scancodes as default is bad or using unicode is bad?

Using keycodes would also be limited to US letters/keys as well. No way to map a key like "ö", which is a normal key on [swedish keyboard](https://upload.wikimedia.org/wikipedia/commons/e/e0/KB_Sweden.svg). In my opinion it seems better to fix the Ctrl issue with unicode, than to use keycodes. 

I should also add that all text of keybinds (help text, menus), convert the key name to the current layout. So if you typed "US_Q", on azerty it should show as "A" in help text.

Anyways, when it comes to the US_ prefix I found this to be the best because:

* It is short, typing EN_US_ in front of every single key seems annoying.
* It should be easier to understand for non-developers since it's not a technical term.
* It's clearly different from the unicode option.
* Sure, us keyboards can have a different layout but... probably like 99.9% of people use qwerty. I think it's a bit silly to disagree with using it because *technically* it could mean something else.
* It's not perfectly descriptive but I think it's the best compromise.

mrpapersonic:
> > I'm not sure if I really like the unicode thing. The most notable quirk is that text input isn't sent at all from SDL when Ctrl is pressed, so mapping e.g. `Ctrl-<some unicode character>` may (and probably will) not work at all. It would be nicer and backwards compatible with the current code to instead use [SDL's keycodes](https://wiki.libsdl.org/SDL2/SDLKeycodeLookup) which will transfer correctly between keyboard layouts and are generally more stable than relying on SDL to give actual text input events.
> 
> 
> 
> Okay didn't know that ctrl would not work, that's not good. 
> 
> 
> 
> However I would argue that using scancodes instead of keycodes transfers better between keyboard layouts. In that they simply ignore the keyboard layout. Using keycodes is what leads to issues like #68 where a keycode is not accessible on a certain layout. Actually reading your comment again I'm not sure if you think using scancodes as default is bad or using unicode is bad?

Both. :) Using scancodes by default means keybinds defaults (not their locations, their values) will change between system to system causing any documentation and current systems to become outdated. Additionally, any current users of Schism on international keyboards will have to completely relearn the keybinds which isn't user friendly at all.

P.S. #68 should have been solved by the SDL2 port which I believe fixed a lot of issues the general keyboard mess SDL 1.2 had.

> Using keycodes would also be limited to US letters/keys as well. No way to map a key like "ö", which is a normal key on [swedish keyboard](https://upload.wikimedia.org/wikipedia/commons/e/e0/KB_Sweden.svg). In my opinion it seems better to fix the Ctrl issue with unicode, than to use keycodes. 

Yes there is. They're called `SDL_GetKeyFromName` and `SDL_GetKeyName`, which have the plus of actually working in UTF-8 so most of the current code won't even need to be changed. so e.g. a truly evil mapping like `Ctrl-AltGr-ö` should work in this scenario. Text input doesn't have to be involved at all, and we can just store the raw keycode plain in memory and print a warning if it couldn't be found.

> I should also add that all text of keybinds (help text, menus), convert the key name to the current layout. So if you typed "US_Q", on azerty it should show as "A" in help text.

That doesn't really help the general problem of having different keybinds depending on the current locale.

> Anyways, when it comes to the US_ prefix I found this to be the best because:
> 
> 
> 
> * It is short, typing EN_US_ in front of every single key seems annoying.
> 
> * It should be easier to understand for non-developers since it's not a technical term.
> 
> * It's clearly different from the unicode option.
> 
> * Sure, us keyboards can have a different layout but... probably like 99.9% of people use qwerty. I think it's a bit silly to disagree with using it because *technically* it could mean something else.
>
> * It's not perfectly descriptive but I think it's the best compromise.

Yea, in hindsight I was being a bit pedantic. `US_` is fine.

I will add that using scancodes for the piano keyboard on the pattern editor is entirely valid (and this was already done as well). However allowing users to remap it to anything they want is always a plus and turns it into their problem if they regularly switch between layouts and use a regular keycode.

include/util.h
```diff
@@ -172,8 +172,33 @@ char *str_escape(const char *source, int space_hack);
...
-char *str_concat(const char *s, ...);
 
+/* Concat strings. Last argument must be NULL */
+// char* str_concat(const char* input_str, ...); // NOT WORKING
+
+/* Count how many times character is in str */
+int str_count_occurences(char character, const char* str);
+
+/* count is size of array. str_array is the array. If free_inputs is true free() will be called on the strings. */
+char* str_concat_array(int count, char** str_array, int free_inputs);
```
mrpapersonic: Use variadic functions here: `str_concat(int count, int free_inputs, ...);`

include/util.h
```diff
@@ -172,8 +172,33 @@ char *str_escape(const char *source, int space_hack);
...
+char* str_concat_three(char* str1, char* str2, char* str3, int free_inputs);
+char* str_concat_four(char* str1, char* str2, char* str3, char* str4, int free_inputs);
+char* str_concat_five(char* str1, char* str2, char* str3, char* str4, char* str5, int free_inputs);
+char* str_concat_six(char* str1, char* str2, char* str3, char* str4, char* str5, char* str6, int free_inputs);
+
+/* count is size of array. str_array is the array. If free_inputs is true free() will be called on the strings. */
+char* str_concat_with_delim(int count, char** str_array, const char* delim, int free_inputs);
+
+/* Get length of utf8 string. Utf8 strings can have a single visible character taking up 2 chars or more. */
+int utf8_length(const char* str);
```
mrpapersonic: I just added a generic charset function for this so it can be replaced with `charset_strlen(string, CHARSET_UTF8)`

schism/keybinds.c
```diff
@@ -0,0 +1,499 @@
...
+
+static int string_to_scancode_length = ARRAY_SIZE(string_to_scancode);
+
+static int parse_shortcut_scancode(keybind_bind* bind, const char* shortcut, SDL_Scancode* code)
+{
+    char* shortcut_dup = strdup(shortcut);
+    str_to_upper(shortcut_dup);
+
+    for (int i = 0; i < string_to_scancode_length; i++) {
+        if (strcmp(shortcut_dup, string_to_scancode[i].name) == 0) {
```
mrpapersonic: your copy of the codebase might not have this (i.e. it's too old), but `charset_strcasecmp` does exactly what you're doing here.

helloimhana123:
>Yes there is. They're called SDL_GetKeyFromName and SDL_GetKeyName, which have the plus of actually working in UTF-8 so most of the current code won't even need to be changed. so e.g. a truly evil mapping like Ctrl-AltGr-ö should work in this scenario. Text input doesn't have to be involved at all, and we can just store the raw keycode plain in memory and print a warning if it couldn't be found.

Okay didn't know that, sounds good!! I can take a look at it tomorrow. 

> Both. :) Using scancodes by default means keybinds defaults (not their locations, their values) will change between system to system causing any documentation and current systems to become outdated. Additionally, any current users of Schism on international keyboards will have to completely relearn the keybinds which isn't user friendly at all.

Yeah okay you're right, I agree that it's bad to change to scancodes cause of current users. I suggest we keep the US_ as an available option, but we change the default binds to use keycodes.

However then I'm not quite sure how to handle modifiers. It's so nice with scancodes because we can require the mods to match exactly. But if we used the code like it is now but made all the binds use the unicode option, then something like "Ctrl+u", would also be triggered by "Ctrl+Shift+u". But we can't just make the unicode version stricter because the whole point with that option is that it allows you to bind keys without having to know exactly what the key combination to reach that key is.

I think probably the best option is to require exact matching modifiers, and the user simple enter the key they want without mods applied. So for example, I can type "ø" by pressing "AltGr+d". I'm thinking I'll remove the ability to simply bind "ø" directly, and instead I would have to enter  into config the keybind "Alt+d". That way we can support things like swedish keyboard while at the same time having nice exact matching modifiers.

Uff anyways. Too tired to think about this properly now gonna have to be in the morning.

schism/util.c
```diff
@@ -834,21 +836,171 @@ char *get_dot_directory(void)
...
+		free(str2);
+	}
+
+	return out;
+}
+
+void str_to_upper(char *s)
+{
+    for (int i = 0; s[i]; i++)
+        s[i] = toupper((unsigned char)s[i]);
```
mrpapersonic: this sucks

mrpapersonic (changes requested):

schism/util.c
helloimhana123: Oookay...? 😆 Doesn't need it anymore anyways cause of charset_strcasecmp

schism/keybinds.c
helloimhana123: Fixed

include/util.h
helloimhana123: Fixed

include/keybinds.h
helloimhana123: Fixed

include/keybinds.h
helloimhana123: Fixed

schism/keybinds.c
helloimhana123: There's no scancode for "enter". [scancodes](https://wiki.libsdl.org/SDL2/SDL_Scancode)

sys/win32/osdefs.c
helloimhana123: Fixed

include/page.h
helloimhana123: Fixed

schism/keybinds.c
helloimhana123: Reason why I didn't do that is because they are not supposed to be standalone modules, they are just a way to split `keybinds.c` into different files, cause it's too long otherwise

schism/keybinds.c
mrpapersonic: Yep because it's usually only present on "extended" 101-key ANSI keyboards on the number pad. (some keyboards even mislabel the return key as an enter key as well, which is just plain wrong)

It is important to note that there really are keyboards with separate physical enter and return keys in the main section, notably old Apple portables. see the the Fn-Return issue for an example.

schism/util.c
mrpapersonic: I should've elaborated more on this lol. This would've failed on any encoding or charset that isn't what the current C locale wants which would be a lot of the time if it's being fed UTF-8.

helloimhana123: Alright gotcha

helloimhana123:
Oke pushed some changes:

* Now using keycodes by default.
* Merged master.
* Fixed some of the issues above.
* Implemented note keybinds.

@mrpapersonic I found an issue with `charset_strcasecmp`, it thinks "F1" == "F11".

Now is break time. 😎

schism/keybinds.c
helloimhana123: Hmm okay so what should I do? I added this code because I wanted to help text to match how it looked before custom keybinds. There it says "Enter" and not "Return".

mrpapersonic: Just change the help text to be correct then :)

mrpapersonic:
> @mrpapersonic I found an issue with `charset_strcasecmp`, it thinks "F1" == "F11".

Fixed with 3dc8c10e76221afbf4f627fbaf91c66155fa0f40

esaruoho:
hi, what happened to this? :)

mrpapersonic:
This hasn't been forgotten, I've just spent the limited time I've had for schism on other things (i.e. better unicode support all around, which will compliment this  PR too)

The implementation details have already been discussed. All of the current code needs to be changed to use keycodes instead of text input, which are better suited for this purpose and avoid the quirks that text input has while also providing simple unicode support out of the box. The current default keycodes in this branch also need to be changed to be keycodes instead of scancodes to
1. keep compatibility with older releases
2. reduce confusion when, say, a German is following an American schism guide and they're wondering why the AdLib sample window is asking them to input a text string
3. be consistent (to the user, not the developer)

These are the only issues that have to be addressed before I feel comfortable merging this.

EDIT: Nevermind, this was already addressed but I never checked it :/

schism/keybinds_codes.c
```diff
@@ -124,3 +124,268 @@ static struct {
...
+static struct {
+    const char *name;
+    SDL_Keycode code;
+} string_to_keycode[] = {
+    {"ENTER", SDLK_RETURN},
+    {"ESCAPE", SDLK_ESCAPE},
+    {"BACKSPACE", SDLK_BACKSPACE},
+    {"TAB", SDLK_TAB},
+    {"SPACE", SDLK_SPACE},
+    {"EXCLAIM", SDLK_EXCLAIM},
```
mrpapersonic: These are represented as raw ASCII (e.g. `SDLK_EXCLAIM` == `'!'`) so these are a bit redundant...

schism/keybinds_codes.c
```diff
@@ -124,3 +124,268 @@ static struct {
...
+    {"F15", SDLK_F15},
+    {"F16", SDLK_F16},
+    {"F17", SDLK_F17},
+    {"F18", SDLK_F18},
+    {"F19", SDLK_F19},
+    {"F20", SDLK_F20},
+    {"F21", SDLK_F21},
+    {"F22", SDLK_F22},
+    {"F23", SDLK_F23},
+    {"F23", SDLK_F24},
```
mrpapersonic: :)

helloimhana123:
> hi, what happened to this? :)

Sorry I've been gone on a trip for several weeks and now that I'm back I've needed time to recover (mental illness is fun! XD). Anyways very good that you asked so we can get this show on the road. 

@mrpapersonic okay good that you made it clear. Like you mentioned in edit I think already this is done. But if you have a specific problem I can fix it of course.

There were two conversations that were unresolved. The first was about using "return" instead of "enter" in help text. In my opinion "enter" is a more common way of saying it and also it was used in previous versions, so I prefer "enter".

The second was about variadic functions. I had a lot of issues making variadic functions work and researching it more made it seem like kind of a hack, so I avoided it.

sys/win32/osdefs.c
```diff
@@ -280,10 +281,18 @@ int win32_sdlevent(SDL_Event* event)
...
 {
-	int wchars_num = MultiByteToWideChar(CP_UTF8, 0, string, -1, NULL, 0);
-	wchar_t* wstr = malloc(wchars_num * sizeof(wchar_t));
-	MultiByteToWideChar(CP_UTF8, 0, string, -1, wstr, wchars_num);
-	return wstr;
+	wchar_t* out = NULL;
+	charset_error_t result = charset_iconv(string, (uint8_t**)&out, CHARSET_UTF8, CHARSET_WCHAR_T);
+
+	if (result != CHARSET_ERROR_SUCCESS) {
+		printf("Failed converting \"%s\" to wchar. Error: %i.\n", string, result);
```
mrpapersonic: `charset_iconv_error_lookup()` gives string representations.

mrpapersonic (changes requested):

schism/keybinds_codes.c
helloimhana123: Sure, I just wanted to make sure I didn't miss anything so I just copied the entire list of keycodes and put it in the list.

schism/keybinds_codes.c
helloimhana123: Same answer as above.

sys/win32/osdefs.c
helloimhana123: Fixed.

mrpapersonic:
> mental illness is fun! XD

real

> There were two conversations that were unresolved. The first was about using "return" instead of "enter" in help text. In my opinion "enter" is a more common way of saying it and also it was used in previous versions, so I prefer "enter".

They were used interchangeably; it would be more accurate to say "return" especially on mac os x where these are actually treated as different keys in many applications.

> The second was about variadic functions. I had a lot of issues making variadic functions work and researching it more made it seem like kind of a hack, so I avoided it.

The way variadic functions work is quite simple:
```c
char *str_concat(int n, ...) {
    int i;
    va_list ap;
    va_start(ap, n);

    int len = 0;

    for (i = 0; i < n; i++) {
        const char *str = va_arg(ap, const char *);
        len += strlen(str);
    }

    va_end(ap);

    char *result = malloc(len + 1);
    result[len] = '\0'; 

    va_start(ap, n);

    len = 0;

    for (i = 0; i < n; i++) {
        const char *str = va_arg(ap, const char *);
        int str_len = strlen(str);
        memcpy(result + len, str, str_len);
        len += str_len;
    }

    va_end(ap);

    return result;
}
```

An example call would be `char *concat = str_concat(2, "first", "second");`, as in `n` is the number of arguments.

helloimhana123:
I'll take a look when I'm able. There's a lot of merging that has to be done first

helloimhana123:
@mrpapersonic oke merged everything in now and changed a bit.

> They were used interchangeably; it would be more accurate to say "return" especially on mac os x where these are actually treated as different keys in many applications.

I made it so it should show "return" on osx now.

> The way variadic functions work is quite simple:

I don't like it cause it can't check for types or correct number of args. But I removed the use of `str_concat_array` and changed name of functions. Instead of using variadic like `str_concat(2, "first", "second");` it's `str_concat_2("first", "second");`. There is up to `str_concat_6`.  Also there is `str_concat_2_free` for freeing the inputs which I think is cleaner than sending in an argument for freeing.

mrpapersonic:
Something to keep in mind is that while windows, mac, etc. display a plus (`+`) to separate keys in keybinds, Impulse Tracker (and Schism by extension) have always used a minus (`-`):

image

The only exception in Schism is for the win32 menu bar (and probably the mac one too, but we don't create the text for that one).

schism/keybinds.c
```diff
@@ -0,0 +1,531 @@
...
+
+static int check_mods(SDL_Keymod needed, SDL_Keymod current, int any)
+{
+	int ralt_needed = needed & KMOD_RALT;
+	int ctrl_needed = needed & KMOD_CTRL;
+	int shift_needed = needed & KMOD_SHIFT;
+	int alt_needed = needed & KMOD_LALT;
+
+	int has_ralt  = current & KMOD_RALT;
+	int has_ctrl  = !has_ralt && (current & KMOD_CTRL);
```
mrpapersonic: why is this checking for `!has_ralt`?

actually, why is `KMOD_RALT` being treated differently from regular alt? they aren't separate keys. however, AltGr is, but that's mapped to `KMOD_MODE`, not `KMOD_RALT`.

include/keybinds.h
```diff
@@ -0,0 +1,722 @@
...
+		keybind_bind_t note_row3_d;
+		keybind_bind_t note_row3_d_sharp;
+		keybind_bind_t note_row3_e;
+	} notes;
+} keybind_list_t;
+
+char* keybinds_get_help_text(enum page_numbers page);
+void keybinds_handle_event(struct key_event* event);
+void init_keybinds(void);
+extern keybind_list_t global_keybinds_list;
```
mrpapersonic, line 735: Having this be a globally mutable variable is kind of scary imo, but i don't see any real way around it for now

helloimhana123:
> Something to keep in mind is that while windows, mac, etc. display a plus (+) to separate keys in keybinds, Impulse Tracker (and Schism by extension) have always used a minus (-):

And this code will also use `-` for help texts and menus (not win menu). Although `+` is used in the keybinding file cause writing `-` just seemed so confusing and wrong.

mrpapersonic:
> And this code will also use `-` for help texts and menus (not win menu). Although `+` is used in the keybinding file cause writing `-` just seemed so confusing and wrong.

all I really care about is if it's somewhat consistent (as in, using `-` everywhere except windows menus OR using `+` everywhere except the main application OR using `-` everywhere unconditionally)

schism/keybinds.c
mrpapersonic: `audio_playback.c` has more lines of code itself than these three files combined. separating them into different sections within the same file, which is what a lot of the other files do as well, would work fine here.

helloimhana123:
> actually, why is KMOD_RALT being treated differently from regular alt? they aren't separate keys. however, AltGr is, but that's mapped to KMOD_MODE, not KMOD_RALT.

This is because ralt isn't handled correctly by SDL. When you press ralt SDL will report both ralt and also lctrl. This is a known issue that wasn't fixed until SDL 3. So this basically means there is no way to detect if only ralt was pressed without lctrl, or they were both pressed together. 

So because ralt is completely messed up by SDL, I decided that it would make more sense for most cases that ralt will cancel out ctrl. Although I realized now that it should only cancel lctrl so I'll fix that.

> audio_playback.c has more lines of code itself than these three files combined. separating them into different sections within the same file, which is what a lot of the other files do as well, would work fine here.

I think that files that are too long add a lot of frustration and unnecessary time hunting for the part that you are trying to edit. It's much better to split things into multiple files and folder for good organization. Having it all in one file just adds so much hassle that can be completely avoided by using multiple files.

helloimhana123:
> actually, why is KMOD_RALT being treated differently from regular alt? they aren't separate keys. however, AltGr is, but that's mapped to KMOD_MODE, not KMOD_RALT.

Also, `KMOD_MODE` never actually works. I have a keyboard with an AltGr key and it never triggers. Doesn't matter if I use US or swedish layout either.

mrpapersonic:
> > actually, why is KMOD_RALT being treated differently from regular alt? they aren't separate keys. however, AltGr is, but that's mapped to KMOD_MODE, not KMOD_RALT.
> 
> Also, `KMOD_MODE` never actually works. I have a keyboard with an AltGr key and it never triggers. Doesn't matter if I use US or swedish layout either.

what...

image

WTF is KMOD_MODE for then??? ughhhhhh

mrpapersonic:
> > actually, why is KMOD_RALT being treated differently from regular alt? they aren't separate keys. however, AltGr is, but that's mapped to KMOD_MODE, not KMOD_RALT.
> 
> This is because ralt isn't handled correctly by SDL. When you press ralt SDL will report both ralt and also lctrl. This is a known issue that wasn't fixed until SDL 3. So this basically means there is no way to detect if only ralt was pressed without lctrl, or they were both pressed together.
>
> So because ralt is completely messed up by SDL, I decided that it would make more sense for most cases that ralt will cancel out ctrl. Although I realized now that it should only cancel lctrl so I'll fix that.

I can't replicate this at all (using a traditional US QWERTY keyboard layout):

image

is that a Windows thing?

helloimhana123:
> WTF is KMOD_MODE for then??? ughhhhhh

I know right XD

> is that a Windows thing?

Hmm actually it [seems it is](https://github.com/libsdl-org/SDL/issues/5685). Ok will only add the fix on windows then.

helloimhana123:
I pushed a new change now. I realized left/right modkeys weren't supported properly so added that so now you can use LAlt, RAlt, RCtrl etc. By default keybinds will accept either side. I also added the RAlt fix only on windows. 

> all I really care about is if it's somewhat consistent (as in, using - everywhere except windows menus OR using + everywhere except the main application OR using - everywhere unconditionally)

So what's the actual problem with the current code?

mrpapersonic:
Nothing, it was really just a comment. so currently the idea would be to use `+` everywhere except for the main application...

mrpapersonic:
Doesn't that `BYTE_TO_BINARY` macro actually get the *last* 8 bits and not the first?

helloimhana123:
> Nothing, it was really just a comment. so currently the idea would be to use `+` everywhere except for the main application...

oh okay. Yeah so `+` is used in the config file and `-` is used everywhere in the actual program. That should be how it is now. Somehow windows menu is using `+` even though I think `-` is sent to it, weird...

> Doesn't that BYTE_TO_BINARY macro actually get the last 8 bits and not the first?

Yes my bad XD

helloimhana123:
@mrpapersonic any remaining issues now?

mrpapersonic:
Haven't really had the time (or really the mental health) to look over it right now, hopefully soon though :)

helloimhana123:
@mrpapersonic no problem! I have the same thing often. Take your time.

mrpapersonic:
I did a bit of research and found a simple way to implement a type-safe `str_concat`, through variadic macros! 

Something like:
```
char* str_concat(size_t count, const char **arr);

#define STR_CONCAT(count, ...) \
    str_concat((count), (const char *[]){ __VA_ARGS__ })
```

This'll make sure that every variable passed is a `(const) char *` while also having the simplicity advantage of var args.

I've implemented this in my local copy but I'm also editing lots else, like for example renaming `str_concat_with_delim` to `str_implode` (name taken from the similar PHP function) and making a similar variadic macro for it as well.

There were other issues I had with the existing functions like using `strcat` when the offset was already known etc etc. Once I get all this done I'll do some general testing to make sure the keybinds still work fine and review the code for memory leaks since there's a lot of dynamic allocation having to happen here.

helloimhana123:
> I did a bit of research and found a simple way to implement a type-safe str_concat, through variadic macros!

Ohh very cool!

> Once I get all this done I'll do some general testing to make sure the keybinds still work fine and review the code for memory leaks since there's a lot of dynamic allocation having to happen here.

Oke sounds good

mrpapersonic:
Current commit breaks the current keybinds; I've gotta get the utf8proc branch merged before it's viable (need `charset_case_fold` which only exists on that branch)

mrpapersonic:
I found quite the weird bug when porting back the Time Information page; if a keybind is defined as needing *both* left alt and right alt, the current code will ignore that and say modifiers are both down, even if only one is pressed.

I'm working on a fix for it right now...

mrpapersonic:
I'm almost comfortable merging this now, it still needs a good bit of testing though. A weird quirk on macos is that only one of the keybinds is actually visible in the menu bar (which is not a fault of ours, it's a limitation of NSMenuItem). Another limitation there is that left/right variants cannot be displayed. Right now it is hardcoded to the first keybind, but ideally it would search for a keybind that is able to be represented in the menu bar and fall back to the first if none could be found.

helloimhana123:
> I'm almost comfortable merging this now

Very exciting 🤩

mrpapersonic:
looking at it now, there are more issues (not counting the new merge conflicts) that have to be addressed, such as:
* having a keybind of, for example, Ctrl-O will override another binding of, say, Ctrl-Alt-O if it comes later
  * this can probably be addressed by doing something like this:
```c
int keybind_mod_held(schism_keymod_t km, enum keybind_modifier mask)
{
    /* we only care about ~held~ modifiers
     * when incorporated into the codebase this should rather be
     * put into keyboard.h as a SCHISM_KEYMOD_TOGGLE mask or something
     * like that */
    schism_keymod_t tmp = km & ~(SCHISM_KEYMOD_NUM | SCHISM_KEYMOD_CAPS | SCHISM_KEYMOD_SCROLL);

    static const struct {
        enum keybind_modifier kbm;
        schism_keymod_t km;
    } km_lookup[] = {
        /* map of keybind modifiers to their keymod mask equivalents */
    };

    for (int i = 0; i < ARRAY_SIZE(km_lookup); i++) {
        if (!(mask & km_lookup[i].kbm))
            continue;

        tmp &= ~(km_lookup[i].km);
    }

    /* punt if there are any unexpected modifiers */
    if (tmp)
       return 0;

    /* (original code of check_mods here) */
}
```
* some parts of the code still check for keymods when they shouldn't (in particular for places where the alt/shift/ctrl keys were separated across different functions)
* probably more I'm forgetting right now

getting the code updated to the *very different* master branch doesn't seem like it'll be too difficult, I think its mostly just the changed names because of removing the explicit SDL 2 dependency

esaruoho:
hi, are we still looking at this?
will the end result be cmd-d for instance settable on macOS to replace alt-d?
would it be possible to export + import keybindings? further down the line?

mrpapersonic:
The keybinds are all stored next to the config file (in `keybinds.ini`) so they are exported by default :) and pretty much any keybind will be able to be changed AFAIK (unless hana missed one, I guess). Additionally there is no UI yet for setting the keybinds, so they have to be configured in the ini file itself.

asbjornu:
Great job @helloimhana123! 🎉 As mentioned in https://github.com/schismtracker/schismtracker/issues/24#issuecomment-2777162821, I would love to see this feature land in Schism Tracker. It's outright impossible to use properly without a separate keyboard and locale for Schism Tracker.

esaruoho:
is there something that can be done to help this PR get landed? like additional testing, or..?
@mrpapersonic

mrpapersonic:
> is there something that can be done to help this PR get landed? like additional testing, or..?

some enthusiasm on my part would've probably made this get merged much sooner :)

anyway I actually have *some* update: the mysterious `MODE` keymod may infact be AltGr after all. see this line in `keyboard.c`:
https://github.com/schismtracker/schismtracker/blob/7c44377647ce310f8dd2c37b8c7c2cf10d480063/schism/keyboard.c#L199

and then, these lines in `config.c`:

https://github.com/schismtracker/schismtracker/blob/master/schism/config.c#L318-L321

Schism automatically, by default, replaces the `MODE` keymod (supposedly AltGr) with a regular `ALT` keymod.

sigh.

anyway, a lot has changed between this branch and master (though it is mostly the names of keybind constants). I suppose it wouldn't take a significant effort to update it. the comments I made above also still apply as well.. :/

schism/keybinds.c
```diff
@@ -0,0 +1,308 @@
...
+
+			char *old = out;
+			out = STR_CONCAT(4, old, (current_title) ? "\n \n  " : "\n  ", (char *)bind_title, "\n");
+			free(old);
+
+			current_title = bind_title;
+		}
+
+		char *old = out;
+		out = STR_CONCAT(3, old, (char*)bind->help_text, "\n");
```
logiclrd, line 303: This approach is inefficient and hard on the heap. Every single row results in the entire help text being copied to a newly-allocated buffer in order append that row, and then the previous one being freed. If the help text is 100 lines long, then that first line gets `strcpy`d 100 times, the second one 99 times, the third one 98 times, the fourth one 97 times, etc. Computers are fast, but this is not tight code. Also, the paradigm is copy/pasted 3 times right here. I kind of feel like it ought to be abstracted.

Since the problem here is bounded -- only the baked-in help text is ever going to be generated anyway -- my gut feeling is that it probably makes sense to just figure out what the upper bound is and do a single allocation of a buffer that'll never run out. This allocation could even just be persistent across the entire process. Some help screens will only use part of it, but that's okay.

logiclrd:
If the keyboard layout is Dvorak, then the key for inputting a period is the one that is marked E on the QWERTY keyboard. In the Pattern Editor view, this key _should_ input an E note in the upper scale, but the way the code detects that the "clear" key has been pressed takes precedence.

Without knowing that this behemoth existed, I created a PR that takes a different approach to sorting out keyboard handling for the Pattern Editor view. My PR (#645) takes the approach of making unmodified keys in the Pattern Editor view be handled by SCANCODE instead of KEYSYM. The help text tracks this accordingly. The key to clear the current entry is such a key, so it is no longer logically PERIOD, but rather "the key that on a QWERTY keyboard is marked with `.`". With this code, on a Dvorak keyboard, the Pattern Editor key to clear the current input is `V`, which is physically placed where the period key is on a QWERTY keyboard.

With configurable key bindings, it could be _changed_ to this, but I don't see any obvious/clean way to make the defaults avoid such collisions.

schism/keybinds.c
mrpapersonic: Probably a good way to go about this would be to just allocate an array of all the help texts, then use the underlying `str_concat` function to concatenate them all together, and free the array.

Also FYI - I've been incorporating many of the changes in this PR to a new `keybinds` branch in this repository, which I'm hoping will end up less abandoned :)

logiclrd:
I see that upstream has diverged considerably since these changes were made, and a change introduced there that causes some controversy for my PR also affects this PR: upstream now supports SDL 1.2 explicitly, and SDL 1.2 lacks keyboard layout abstractions. `SDL_GetKeyFromScancode` doesn't exist. :-(

PR #478, opened 2024-06-24
remove all uses of PATH_MAX

mrpapersonic:
This has a couple of bugs (notably in the variables page where I'm pretty sure there's a heap overflow somewhere) but at least its something

PR #493, opened 2024-07-13
Drop use of PATH_MAX on non-Windows platforms.

petterreinholdtsen:
There is no need to hardcode a maximum size limit for home directory path nor backup file name paths, and dropping the artificial limit get the source building on GNU Hurd.


mrpapersonic:
This is actually in progress at #478 which will additionally remove the size limit for the modules, samples, and instruments directories.

PR #498, opened 2024-07-24
Fix for the user defined colors not getting saved bug

hotgar:
This fixes a bug which prevents user defined colors from being saved to the "User Defined" palette or copied to clipboard.

Behavior before the fix:
https://github.com/user-attachments/assets/937db1ad-edbc-4dd4-806b-66b88a1c4358

After:
https://github.com/user-attachments/assets/3c469ed3-c972-4411-8ddb-67c6a22f0859

### DIFF
--- a/schism/page_palette.c
+++ b/schism/page_palette.c
@@ -343,7 +343,7 @@ static void update_palette(void)
 		current_palette[n][2] = widgets_palette[3 * n + 2].d.thumbbar.value;
 	}
 	selected_palette = current_palette_index = 0;
-	memcpy(palettes[selected_palette].colors, current_palette, sizeof(current_palette));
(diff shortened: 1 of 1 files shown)


schism/page_palette.c
```diff
@@ -343,6 +343,7 @@ static void update_palette(void)
 		current_palette[n][2] = widgets_palette[3 * n + 2].d.thumbbar.value;
 	}
 	selected_palette = current_palette_index = 0;
+	memcpy(palettes[selected_palette].colors, current_palette, sizeof(current_palette));
```
mrpapersonic: should explicitly request `palettes[0].colors` to avoid confusion

hotgar: Done

PR #500, opened 2024-07-25
Fix for exit prompt stacking

### LINKED ISSUE #487: Exit confirmation dialogue stacking
Repeated presses of the close window button or Alt + F4 while the exit confirmation dialogue is on-screen will require extra presses of 'cancel' to close the confirmation dialogue.
https://github.com/schismtracker/schismtracker/assets/151794548/ffa0160d-65dd-4830-9f37-9855bf98046d

hotgar:
Fixes #487

It looks like the code handling exiting via Ctrl+q already has a check for presence of other dialogs 
https://github.com/schismtracker/schismtracker/blob/e444afa0da227c4b114768a779e7c4cc3992a6d0/schism/page.c#L558C2-L560C13
so i added a similar check to the handling of the `SDL_QUIT` event.

Maybe the check should be moved inside the  `show_exit_prompt()` function since it already checks the current page status, I'm not sure.

### DIFF
--- a/schism/main.c
+++ b/schism/main.c
@@ -612,8 +612,7 @@ static void event_loop(void)
 				}
 				break;
 			case SDL_QUIT:
-				if (status.dialog_type == DIALOG_NONE)
-					show_exit_prompt();
+				show_exit_prompt();
 				break;
 			case SDL_WINDOWEVENT:
 				/* reset this... */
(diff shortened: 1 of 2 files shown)


hotgar:
Also placing the check for a dialog outside the `show_exit_prompt()` makes it impossible to quit from the "about" page so just copying the "ctrl-q" behavior isn't ideal here.

Issue #484, opened 2024-06-27
Mouse cursor outline and envelope cursor


0xHexaDecimator:
In IT the mouse cursor has a thin outline, and a smaller cursor for click-dragging instrument envelopes. Neither are present in Schism, but would help cursor visibility. 
https://github.com/schismtracker/schismtracker/assets/151794548/ceb6b070-e895-4f10-8160-8eb2f50a9aac

Lumping these two together because they're about cursor appearance.

hotgar:
I've managed to make the cursor look more similar to the IT one. In the IT case it seems like the body of the pointer always has the foreground color and the outline has the background one rather than getting colors from XORing. But its rendering over some parts of the UI is still a bit weird. Not sure how much better than the current one this is.

https://github.com/user-attachments/assets/2f15a9c0-f382-4d84-bd8a-1b254773ebc8


Haven't touched the envelope editor crosshair yet.

mrpapersonic:
It's possible (likely even) that some foreground colors are completely wrong. See `draw_text_len` in `schism/vgamem.c`, more particularly the `draw_fill_chars` call which just shoves the same color in both slots. This isn't the full story though, because the scanner checks for the character value set with `draw_fill_chars` and sets the foreground to 3, which may or may not also just be plain wrong ;)

hotgar:
Yes, it looks like the rendering oddities are caused by 

https://github.com/schismtracker/schismtracker/blob/e444afa0da227c4b114768a779e7c4cc3992a6d0/schism/vgamem.c#L285-L286

which seems to be there exactly to deal with the `draw_fill_chars` taking only one color and setting both fg and bg to it.

So i guess adding a parameter to the `draw_fill_chars` and maybe setting something like  `DEFAULT_FG 3` in `vgamem.h` would be the solution here.

I've also noticed that this issue doesn't happen with strings that are slurped from an xm module like instrument or sample names because they seem to be padded. So padding the strings could also work but that seems more like a crutch to me.

mrpapersonic:
The PR attached to this issue was merged, but github didn't see that :/

PR #503, opened 2024-07-28
Cursor rendering rework

hotgar:
Addresses a part of #484 
Makes the mouse cursor more similar to the IT one and fixes a bunch of small bugs affecting mouse visibility.

https://github.com/user-attachments/assets/c5553622-450b-42ec-9562-27930fc83f4d

### DIFF
--- a/schism/page_info.c
+++ b/schism/page_info.c
@@ -255,15 +255,18 @@ static void info_draw_samples(int base, int height, int active, int first_channe
 
 	for (pos = base + 1; pos < base + height - 1; pos++, c++) {
 		song_voice_t *voice = current_song->voices + c - 1;
-
 		/* always draw the channel number */
-		if (c == selected_channel)
+		memcpy(buf, numtostr(2, c, buf), 8*sizeof(char));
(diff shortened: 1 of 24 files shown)


schism/page_info.c
```diff
@@ -255,15 +255,18 @@ static void info_draw_samples(int base, int height, int active, int first_channe
 
 	for (pos = base + 1; pos < base + height - 1; pos++, c++) {
 		song_voice_t *voice = current_song->voices + c - 1;
-
 		/* always draw the channel number */
-		if (c == selected_channel)
+		memcpy(buf, numtostr(2, c, buf), 8*sizeof(char));
```
mrpapersonic: `numtostr` modifies `buf` inplace and returns the same pointer. This code is *undefined behavior*:

> If copying takes place between objects that overlap, the behavior is undefined.
>
> https://pubs.opengroup.org/onlinepubs/007904975/functions/memcpy.html

mrpapersonic (changes requested):

PR #516, opened 2024-08-10
Depend on utf8proc for Unicode

mrpapersonic:
This has multiple advantages:
* Since the utf8proc library has a stable ABI, it's trivial to upgrade just the library file itself with no side effects for schism.
* Real string normalization
* Full case folding as opposed to schism's simple case folding
* Likely most important: making Unicode a "not our problem" deal
* Probably many, many others

The build is failing because cmake is odd and I can't figure out how to get it to cross compile :/


mrpapersonic:
squashed into master with a0c007b4269f93fa0a69c76f60d8274b3dc08dd1

PR #524, opened 2024-08-25
fmt: edit the loader functions to use slurp

### LINKED ISSUE #521: [longterm] Need to replace memory mapping for files
Memory mapping files kind of... sucks. The best way I can think of to do this is to retrofit the slurp code into being a glorified `FILE *`. In addition, I'd like to refactor the init functions to modify a passed pointer instead of forcing heap allocation.

mrpapersonic:
Closes #521


mrpapersonic:
Oddly enough, WAV and AIFF info loading is completely broken under mmap, but works fine with the stdio driver. (probably something bugged)

mrpapersonic:
Merged into master with the media foundation commits rebased into one commit.

Issue #528, opened 2024-09-25
Load Module: file search is broken


drradio:
File search in Load module page does not work in version 20240909, but works in 20240809.
How to reproduce:
1) Open Load module page (F9);
2) Enter part of file name;
3) Nothing...

OS: Debian GNU/Linux 12.7 (stable)
Tested build: Schismtracker 20240909 (built from source)

0xHexaDecimator:
Broken with 20240909 on Windows 10 as well 

EDIT: Apparently it works if I type the _entire_ filename (including extension) into the searchbar

drradio:
Found a bug in charset_strncasecmp. Preparing patch.

drradio:
Created pull request #529.

PR #529, opened 2024-09-26
Fix codepoint comparisions in *cmp* functions

drradio:
Fix for issue #528.
Possibly fixes everything related to file search.


mrpapersonic:
There's a bigger underlying issue with `charset_strncasecmp` in that it completely ignores the count argument. I'm not entirely sure why I changed it to that (likely because the size of the string can change, I was going to fix that, and got distracted)

Unfortunately implementing a Unicode strncasecmp in a sane way is actually surprisingly difficult. It can probably be done with something like:
```
for (size_t i = 0; i < count; i++) {
    /* decode next char */
    uint32_t cp1[2] = {decoder1.codepoint, 0}, cp2[2] = {decoder2.codepoint, 0};

    uint32_t *ucs4_folded1 = charset_case_fold_to_set(cp1, CHARSET_UCS4, CHARSET_UCS4), *ucs4_folded2 = charset_case_fold_to_set(cp2, CHARSET_UCS4, CHARSET_UCS4);

    int cmp = charset_strcmp(ucs4_folded1, CHARSET_UCS4, ucs4_folded2, CHARSET_UCS4);
        
    free(ucs4_folded1);
    free(ucs4_folded2);

    if (!cmp)
        break;
}
```
note this doesn't have any proper castings and is not tested at all!!

EDIT: implemented with 0a3d9d70c7b8c2b29541fb4489a2d91bfb5ddadb

mrpapersonic:
P.S. this doesn't really fix anything, and it even breaks `charset_strcmp` as well to where it will never return zero on a non-empty string because it doesn't reset the result value to zero before returning. This same thing happens in `charset_strcasecmp` and `charset_strncasecmp` but only in the latter when the strings are the same length.

PR #534, opened 2024-10-03
Fixed resizing sample with interpolation.

### LINKED ISSUE #533: Resize sample with interpolation function doesn't work correctly.
Hello, developers.

Resize sample without interpolation(Alt-F) is work well but, Resize sample with interpolation(Alt-E) doesn't work correctly. When I use the resizing with Alt-E, the samples turn into flat wave and the sound becomes weird.

It is not urgent because it is fine to use Alt-F but, I compared with original impulse tracker on DOS Box, but I think this is not work well.

 * Before Resizing
image

 * After Resizing (I only changed 1 length)
image

yoonsangmin:
It fixed #533.

I found this bug and cheked this issue myself. I found the typo in resize with aa code at this 2c64682 commit.

I tested it on linux environment.
I couldn't find the contribution guidelines, so I made a fix on my own and submitted a pull request. I apologize for that. Also, I’m not very good at English, so I ask for your understanding.


mrpapersonic:
thanks!!

PR #535, opened 2024-10-03
Fixed inverted key repeat setting

yoonsangmin:
I found checking key_repeat_enabled variable in event handler loop is inverted and fixed it.


mrpapersonic:
I assume you already figured this out, but that check is there so we use the system key repeat setting when one isn't configured in Schism.

yoonsangmin:
@mrpapersonic I misunderstood it at first, but I found that it works as intended. However, I found note playback repeats when using the system repeat speed, so I'm trying to figure out the cause.

PR #536, opened 2024-10-03
Mention pkg-config in building_on_osx.md

lukas2:
Wouldn't configure on my system without png-config. Possibly this can be solved by adjusting paths, but.. using png-config just worked..


mrpapersonic:
odd, pkgconf works for me. Maybe Debian has a patch that Homebrew doesn't have.

Thanks!

PR #546, opened 2024-10-21
build: remove redundant AC_SUBST calls from configure.ac

jengelh:
PKG_CHECK_MODULES uses AC_ARG_VAR, which in turn is documented to imply AC_SUBST.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -163,9 +163,6 @@ if test "x$utf8proc_found" = "xno"; then
 	AC_MSG_ERROR([*** Failed to find utf8proc])
 fi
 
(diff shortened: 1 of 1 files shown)


mrpapersonic:
This was intentional, see [Autotools Mythbuster](https://autotools.info/pkgconfig/pkg_check_modules.html):
> On older versions of pkg-config, the macro will not call AC_SUBST on these variables; modern versions (at least version 0.24) will take care of that already.
> Running it twice, though, will not cause problems, so if you have doubts, you should add a snippet like the following.
> ```
> AC_SUBST([FOO_CFLAGS])
> AC_SUBST([FOO_LIBS])
> ```

PR #547, opened 2024-10-21
build: resolve -Wreturn-type warnings

jengelh:
Functions without return can produce unpredictable behavior, and it goes without saying that is bad.
openSUSE has added -Werror=return-type distro-wide, so it is perhaps fair to do the same here.

```
fmt/iti.c: In function ‘load_it_instrument_old’:
fmt/iti.c:157:1: error: control reaches end of non-void function [-Werror=return-type]
  157 | }
fmt/iti.c: In function ‘load_it_instrument’:
fmt/iti.c:207:1: error: control reaches end of non-void function [-Werror=return-type]
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -452,6 +452,10 @@ if test x$ADD_WARN \!= xno; then
 	CFLAGS="$ADD_WARN $CFLAGS"
 fi
 
+dnl Unlike a grand -Werror, this one could be rather important:
+dnl functions returning random values are no good under any circumstances.
+AS_IF([test ${ac_cv_c_compiler_gnu+y}], [CFLAGS="$CFLAGS -Werror=return-type"])
+
 if test x$ADD_PROFILING \!= xno || test x$ADD_DEBUG \!= xno; then
 	CFLAGS="$CFLAGS -g -O0"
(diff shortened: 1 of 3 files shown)


mrpapersonic:
Can this check if the compile flag works before trying to use it? I'm pretty sure `-Werror` and friends are gcc-isms.

jengelh:
That much is true, but are there any platforms which utilize both ./configure and a non-gcc-non-clang compiler?

mrpapersonic:
I use tinycc very often and AFAIK it doesn't have those error flags ;)

Much much more niche but I'm sure solaris still uses its own C compiler.

jengelh:
> I use tinycc very often and AFAIK it doesn't have those error flags ;)

Fair point; since configure tests for GNU C more or less implicitly anyway it seems, I've made -Werror=return-type conditional upon that.

> Much much more niche but I'm sure solaris still uses its own C compiler.

With Solaris being pretty much dead _and_ C++ projects reaching for higher standard levels (like C++20), I doubt that SunStudio/Oracle Developer Studio has much relevancy left either. (Quite unlike the BSDs and e.g. OSF/1, Solaris did not come with a free compiler TTBOMK.)

PR #555, opened 2024-11-07
Allow non-SDL backends

mrpapersonic:
This largely moves everything that was SDL-specific into `sys/sdl2`. Currently as a test, I added a backend for SDL 1.2 that pretty much just reused a lot of the code from before the SDL2 port.


mrpapersonic:
merged into master as of 197f4f7e2da2633c2bf82a19cdab5b7104a8f91e

PR #571, opened 2024-12-17
* Make the original Time Information shortcut actually work on Window…

a11cf0:
…s. Since Ctrl-Pause is in fact Ctrl-Scroll Lock, catch it as well.

* Fix octave switching with KP / and KP *.

### DIFF
--- a/schism/page.c
+++ b/schism/page.c
@@ -872,7 +872,11 @@ static int handle_key_global(struct key_event * k)
 		if (status.dialog_type != DIALOG_NONE)
 			return 0;
 		_mp_finish(NULL);
-		if (k->mod & SCHISM_KEYMOD_ALT) {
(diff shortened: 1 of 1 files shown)


mrpapersonic:
IMO this should rather fix the keysym in `win32_event()`. A possible fix for now would be to use `GetKeyboardState` and check for `VK_PAUSE` but this may not work on older, slower systems.

Additionally, if this problem only exists in SDL2 it should be moved into just `sys/sdl2/events.c` as a platform-specific hack. (SDL2 seems to need a lot of these for windows in particular...)

a11cf0:
OK. The Pause hack should indeed be rewritten. What about octave switching with numpad keys? I'm probably missing something, but it seems that kbd_key_translate makes it impossible to know if a numpad key was pressed without checking orig_sym no matter what backend or system is used.

mrpapersonic:
That's intentional, and its made ambiguous so users without a numpad can use those keys. I don't know of a case where these keys overlap anyway (if they do, the original keysym does need to be checked).

and P.S. I am one of the users without a numpad ;)

PR #600, opened 2025-01-23
Add --headless CLI flag to enable headless exporting to disk

trh-x:
Adds the `--headless` flag. Requires the `--diskwrite` flag, will export to disk without initialising the GUI.

```
Usage: ./build/schismtracker [OPTIONS] [DIRECTORY] [FILE]
  -a, --audio-driver=DRIVER
  -v, --video-driver=DRIVER
      --classic (--no-classic)
  -f, --fullscreen (-F, --no-fullscreen)
  -p, --play (-P, --no-play)
      --diskwrite=FILENAME
      --font-editor (--no-font-editor)
      --hooks (--no-hooks)
      --headless
      --version
  -h, --help
```


mrpapersonic:
whoops, clicked the wrong button :)

trh-x:
Hey @mrpapersonic, thanks for taking an interest (even if accidentally 😆 ).

I've just pushed an update to the man page, so this is ready to review.

schism/main.c
```diff
@@ -955,6 +962,47 @@ int schism_main(int argc, char** argv)
...
+			fprintf(stderr, "Error: --headless requires --diskwrite\n");
+			return 1;
+		}
+		if (!initial_song) {
+			fprintf(stderr, "Error: --headless requires an input song file\n");
+			return 1;
+		}
+
+		// Initialize minimal systems
+		video_startup();
```
mrpapersonic, line 978: is there any reason for video to be initialized?

trh-x: Fair question! I had to go back and check, but yes it's needed because it does some fundamental setup that the song export relies on. It might be nice to refactor/rename it at some point, but it seems OK for now...

PR #610, opened 2025-02-09
Clearer error messages when saving nameless samples

mawkish:
Currently the sample list's alt+o, alt+w, and alt-t operations fail with "Sample <n> NOT saved! (File Error)" when the sample hasn't been given a filename. With alt+w and alt+o the sample can have no filename if the "filename" field has been left blank; with alt+t it happens when the dialog text box is left blank.

Here I check for each of these operations that the filename is blank and if so bail with a more descriptive error message.

### DIFF
--- a/schism/page_samples.c
+++ b/schism/page_samples.c
@@ -1117,6 +1117,11 @@ static void do_export_sample(SCHISM_UNUSED void *data)
 		if (sample_save_formats[i].enabled && !sample_save_formats[i].enabled())
 			exp++;
 
+	if (!*export_sample_filename) {
+		status_text_flash("Empty filename; can't save sample");
+		return;
+	}
+
 	sample_save(export_sample_filename, sample_save_formats[exp].label);
 }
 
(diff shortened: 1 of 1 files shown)


mrpapersonic:
In Impulse Tracker I'm fairly sure that the proper error for this is `Sample NOT saved! (No Filename?)`. I've had to fix this multiple times because I keep breaking it somehow.

Nevertheless, all of these functions end up just calling `sample_save` anyway so this logic ought to be moved there, probably something like
```c
if (!filename || !*filename || !*sample->filename) {
    status_text_flash("Sample NOT saved! (No Filename?)");
    return;
}
```

mawkish:
This seems reasonable to me. I'll update my pull request now.

mawkish:
Sorry, never mind that last commit. I didn't properly check it over. I'll do another one now.

mawkish:
OK, that should be it now.

mrpapersonic:
couldn't it just be `*(filename ? filename : sample->filename)` ?

mawkish:
I've made that change. It seems to work well.

mrpapersonic:
Thanks!

PR #625, opened 2025-03-13
XI instrument import/export fixes

Spectere:
Greetings! I ended up stumbling upon a few regressions when I was trying to import some old XIs and went down a bit of a rabbit hole. Here goes:

- Fixed a regression introduced in 5912f37 that caused XI to read sample data from the wrong place (`header, data, header, data` rather than `header, header, data, data`).
- Corrected an oddity in the XI save function, likely stemming from its ITI roots, which would cause it to initially use 1-based indexing instead of XI's 0-based indexes.
- The default panning flags were not taken into account when exporting FT2 samples, which would usually cause samples to panned hard left. Panning is assigned using the following precedence: instrument panning, sample panning, centered.
- The sample save function was incorrectly doubling the size (in bytes) of 16-bit samples. At best this would cause them to take up double the space and at worst (particularly when exporting instruments) it would cause a segfault or, in the case of XIs, corrupted samples.

I've only tested these changes on Linux at this time, and have verified the exported XIs against Schism and Milky Tracker.

### DIFF
--- a/fmt/xi.c
+++ b/fmt/xi.c
@@ -538,7 +538,13 @@ int fmt_xi_save_instrument(disko_t *fp, song_t *song, song_instrument_t *ins)
 		xmss.looplen = bswapLE32(xmss.looplen);
 
 		xmss.vol = smp->volume >> 2;
-		xmss.pan = smp->panning;
+
+		if(smp->flags & CHN_PANNING) {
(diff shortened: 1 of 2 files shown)


mrpapersonic:
Thanks!!

vcaputo:
+1

PR #639, opened 2025-05-20
Add a simple "test suite" of sorts

mrpapersonic:
This just plugs into `schism_main`. It's enabled by passing `--enable-test` to `configure`. Currently we only test the config functions, bit shift, and byteswap, but more is probably due to be implemented (for example, sample read/writing functions).


mrpapersonic:
closing in favor of #715

Issue #642, opened 2025-05-26
BUG: Keyboard input is sometimes translated by scan code, sometimes by logical key


logiclrd:
The key input handling in Schism Tracker uses logical key values in some places (e.g. -/+ for switching patterns, ;/' for switching instruments), and raw scan codes in others (`kbd_get_note`). If you are using a QWERTY keyboard, these overlap in a sensible way and create the consistent keyboard input arrangement that users of Impulse Tracker/Schism Tracker are accustomed to.

But, if the keyboard's current layout is not QWERTY, then this disconnect creates problems and makes keyboard input impractical in various ways.

For instance, if the keyboard layout is set to Dvorak, then the locations of the quote key and semicolon character are completely different. They are where Q and Z are on a QWERTY keyboard. The result is that if you press the lower-left key on the keyboard, marked Z with QWERTY, expecting to get a low C, the keystroke never makes it to `kbd_get_note` because the _logical_ key that was pressed was a semicolon. So then you press the top-left key, expecting a C one octave up. Nope, that key's _logical_ value is the quote character. So, you are unable to input the note C in the classical way. Instead, you're just scrolling through the instruments.

I believe the most sensible thing would be for _all_ input that isn't to text fields to go based on `SCHISM_SCANCODE_` constants, completely ignoring the key's logical value. That way, the keyboard's function is identical no matter what layout is imposed at the system level, and the actual values of the keys from that perspective is only relevant when you're typing in a new name for e.g. the song, an instrument or a file.

mrpapersonic:
This is because the semicolon character is being eaten by the `pattern_editor_handle_key_default` function. I'm not really sure what the logic in there is doing (probably accounting for *another* edge case)

> I believe the most sensible thing would be for all input that isn't to text fields to go based on SCHISM_SCANCODE_ constants, completely ignoring the key's logical value. That way, the keyboard's function is identical no matter what layout is imposed at the system level, and the actual values of the keys from that perspective is only relevant when you're typing in a new name for e.g. the song, an instrument or a file.

This came up in an earlier discussion: https://github.com/schismtracker/schismtracker/pull/470#issuecomment-2179664187 -- this would result in breaking behavior for international users as well as being different from most every other application. It would also mean having to dynamically create the help text at runtime to adjust for the keyboard differences. If anything, it's just a band-aid over schism expecting that nothing but letters will be in the piano keyboard's place, which is true for most, but not all keyboard layouts.

[...]

PR #643, opened 2025-05-26
Changed the paradigm for input processing: KEYSYM if it's textual input, SCANCODE otherwise

logiclrd:
This PR replaces all locations that inspect `key_event`'s `sym` member for a purpose that is not text input to instead inspect the `scancode` member, switching the values from `SCHISM_KEYSYM_*` to `SCHISM_SCANCODE_*` accordingly.

One point in particular, the method `fontedit_handle_key` splits out the handling of digits from the main `switch` block so as to not have to worry about keypad digits having different scan codes.

With these changes, I have not tested every possible key input in every dialog, but with my Dvorak-layout keyboard:
- In the pattern editor, the letter keys on the keyboard act like a piano again.
- In the effect data column, because `kbd_char_to_hex` has not been switched to scan codes (being for textual input), the letters A to F are where they are with the Dvorak layout.
- In every text input field I could find (song name, sample name/filename, instrument name/filename, file/dir list quick search, filename entry, etc.), text input is natural for the keyboard layout and uses the Dvorak keys.
- All accelerator key combos are forced to the QWERTY layout.

(For instance: selecting the entire column/pattern is Alt and the last key on the home row, even though with Dvorak, L is the last key on the top row instead, marked P on QWERTY keyboards. Pressing Alt and the last key on the top row gives me Alt-P even though that's the letter L with Dvorak.)

The net result of these changes is that except for text input, the current keyboard layout being imposed at the system level is irrelevant. When you _are_ entering text, though, you just type naturally.

### DIFF
--- a/include/it.h
+++ b/include/it.h
@@ -150,7 +150,7 @@ struct tracker_status {
 	int flags;
 	enum tracker_time_display time_display;
 	enum tracker_vis_style vis_style;
-	schism_keysym_t last_keysym;
+	schism_scancode_t last_scancode;
 
 	schism_keymod_t keymod;
 
--- a/schism/itf.c
+++ b/schism/itf.c
@@ -850,18 +850,18 @@ static int fontedit_handle_key(struct key_event * k)
 	}
 
 	/* kp is special */
-	switch (k->sym) {
-	case SCHISM_KEYSYM_KP_0:
+	switch (k->scancode) {
+	case SCHISM_SCANCODE_KP_0:
 		if (k->state == KEY_RELEASE)
 			return 1;
 		k->sym += 10;
 		/* fall through */
-	case SCHISM_KEYSYM_KP_1: case SCHISM_KEYSYM_KP_2: case SCHISM_KEYSYM_KP_3:
-	case SCHISM_KEYSYM_KP_4: case SCHISM_KEYSYM_KP_5: case SCHISM_KEYSYM_KP_6:
-	case SCHISM_KEYSYM_KP_7: case SCHISM_KEYSYM_KP_8: case SCHISM_KEYSYM_KP_9:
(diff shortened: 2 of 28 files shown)


sagamusix:
This change doesn't seem quite right?

As a German keyboard layout user, I expect that Ctrl+Z undoes the last action in most applications using standard shortcuts, even though the Z and Y letters are swapped compared to US keyboard layout. Both Impulse Tracker and Schism Tracker currently get this right: Pressing Alt+Z cust a pattern block, just like the F1 page says. Going by scan codes, I would have to press Alt+Y, which is not the expected behaviour.

Going by scan codes only really makes sense for note input, but not for shortcuts. However, Schism Tracker already seems to be doing the right thing at least with a German keyboard layout when entering notes? e.g. pressing Z gives me an A-5 in the pattern, just like pressing Y on a QWERTY layout would.

logiclrd:
Mm, okay, so the focus of the change is wrong. It needs to enforce scancodes specifically for non-modified keys in the pattern view _only_. And, as you mentioned in the issue #642, it would be ideal for the documentation to reverse-map the scancodes so it could accurately tell the user which buttons to press (so e.g. the documentation that says to press `[` and `]` would instead be saying to press `/` and `=`, as those are the same physical keys with the Dvorak layout).

You mention that SDL 1.2 doesn't have any mechanism for mapping scancodes to logical keys. My first response to that is, is it important today to still support SDL 1.2? SDL 2 came out 12 years ago now. But maybe there are some platforms where only SDL 1.2 is available on which people are actively using Schism Tracker?? If not, though, then ditching SDL 1.2 support would resolve the tension about not being able to map scancodes. Just a thought.

logiclrd:
#645

PR #644, opened 2025-05-26
itf: highlight all matching characters

logiclrd:
In the font editor, if you select a character that appears more than once in the map, only the first match is highlighted. This PR updates `draw_itfmap` to highlight all matching characters so that the user can see that their edit is going to affect multiple places simultaneously. This is only done when the `selected_item` is not `ITFMAP`; it would be confusing if characters other than the cursor were highlighted when interact with the map directly.

### DIFF
--- a/schism/itf.c
+++ b/schism/itf.c
@@ -300,6 +300,7 @@ static inline void draw_itfmap(void)
 {
 	int n, fg, bg;
 	uint8_t *ptr;
+	uint8_t highlight_char = 0xFF;
 
 	if (itfmap_pos < 0 || itfmap_chars[itfmap_pos] != current_char) {
 		ptr = (unsigned char *) memchr((char *) itfmap_chars, current_char, sizeof(itfmap_chars));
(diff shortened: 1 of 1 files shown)


logiclrd:
@mrpapersonic This one? :-) I think it significantly improves the font editing experience.

mrpapersonic:
to be honest I'm not entirely sure what this is supposed to do... I've never used the font editor :)

logiclrd:
Okay :-) So, the font editor is a pretty straightforward bitmap editor for the array of characters in a font. It has 3 panes:

* The first pane is the bitmap for the current character being edited. This is an 8x8 grid of on/off pixel values.
* The second pane is the raw grid of all characters in the font. This is a 16x16 grid where each character is just (uint8_t)(row * 16 + col).
* The third pane is something referred only as "ITFMAP" internally. I've taken to interpreting this as "Impulse Tracker Font Map". It's a little picture that presents a preview of the font in various use cases. It includes some basic drawing, VU meters, etc.

When the focus is in the ITFMAP pane, you can move the cursor around to a particular character, and that character will then be selected and you can then go and edit it.

When the focus is in the other panes, the code makes a basic attempt to highlight the corresponding character in the ITFMAP pane, to show you what you're editing. But the key thing is, _there are repeats in the ITFMAP pane_. The existing code only highlights the first match. With this PR, all matches are highlighted, since they're all the same character and are all being edited simultaneously.

Before:

image

After:

image

And I just noticed I typoed the branch name and didn't notice all the way through to creating the PR. "Editer". Oof. :-P

logiclrd:
(And in case you're looking at your keyboard and wondering how you could typo an 'o' as 'e', the keys are adjacent with the Dvorak layout. :grin: )

mrpapersonic:
> (And in case you're looking at your keyboard and wondering how you could typo an 'o' as 'e', the keys are adjacent with the Dvorak layout. 😁 )

English suffixes -er and -or sound the exact same 99% of the time so I just assumed it was that :)

PR #645, opened 2025-05-27
page_patedit: keyboard layout agnostic & graphical piano in helptext

logiclrd:
This PR makes three main changes:

- The Pattern Editor's handling of unmodified keys is made to be agnostic of keyboard layout. The keys are processed based on their location (scan code), rather than their value. This only applies to keys unmodified by Ctrl or Alt, so for instance Alt-L to select column/pattern uses the L that is idiomatic to the current keyboard layout, but `<`/`>` to switch instrument is always Shift and the keys above the right Alt key, which are marked `,` and `.` on a QWERTY keyboard but `W` and `V` on a Dvorak keyboard.
- Rendering of the help screen for the Pattern Editor is updated to map the keys accordingly and inject the mapped values into the help text. So, `<` and `>` to switch instrument will render as `Shift-W/Shift-V` if the keyboard layout is Dvorak. I split this into two lines, since the inclusion of "Shift-" makes the key names too long to fit all four of them in the left column.

image

- The Pattern Editor help screen is also given a graphical depiction of a piano keyboard showing how keyboard keys map to piano keys.

image

image

This mapping of keystrokes based on the current layout depends on `SDL_GetKeyFromScancode`, which is only present in SDL 2 and SDL 3. The SDL 1.2 implementation has a fallback that doesn't take keyboard layout into account and just renders keys as they would be for a QWERTY keyboard.

This PR resolves the issue of the pattern editor not accepting input treating the letters like a piano without making the help text wrong.


logiclrd:
I've just pushed a commit that makes the Pattern Editor help screen show a graphical depiction of the piano keyboard that the keyboard is representing. It would really be better as its own PR, but it builds on the changes in this PR, so at this point it could only be a PR that merges into this PR's source branch.

If you like this but don't want to mix purposes in the PR, and you are amenable to this PR's main changes being merged in, I'd be happy to move it off of this branch and into a separate PR after this one is merged.

image

image

logiclrd:
Rebased. This was a fun one, with the VGAMem format change from 141b87428 :-)

logiclrd:
I put a screenshot of the QWERTY piano visualization into the PR description. :-)

@mrpapersonic Ping

logiclrd:
Rebased.

logiclrd:
Ruh roh :slightly_smiling_face: Trouble in DevKitPro land, looks like? :stuck_out_tongue:

mrpapersonic:
yea, their servers keep going down. Nothing we can really do about

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
This one would be quite useful. At present, if I'm not building off this branch, I literally can't enter either of the C notes on the left end of the keyboard, because the low one is `;` and the high one is `'`, and the current code steals those for next/prev sample/instrument.

logiclrd:
Rebased. Please, sir, may I have some keyboard layout? 😛

logiclrd:
Hello, it's me, PR that wants to be merged :-) Is there a problem with this PR that's preventing it from being merged??

mrpapersonic:
> This one would be quite useful. At present, if I'm not building off this branch, I literally can't enter either of the C notes on the left end of the keyboard, because the low one is `;` and the high one is `'`, and the current code steals those for next/prev sample/instrument.

I can input those just fine on Linux when I set my layout to Dvorak since commit 15b3375be3c928c87170338f46914d09d7792bf2.

logiclrd:
I don't think it fully covers what this PR handles.

* Can you change instruments with `<` / `>`? When I try with the actual `<` and `>` characters, it just enters notes, and when I try with the keys that are `<` and `>` on QWERTY (`W` and `V` on Dvorak), nothing happens.

* Can you toggle the note field in the edit mask? When I press `,` on the instrument or volume fields, the corresponding edit mask field toggles, but if the cursor is in the note field, it just enters a `D-5` (or what have you).

* Jumping to the next `Shift-F` is awkward. `F` on Dvorak is the key that with the QWERTY layout is `Y`. The functionality works, but in the QWERTY layout, prev and next are `Shift-A` and `Shift-F` because those keys fall naturally under the little and index fingers, respectively.

Plus, this PR makes the help text dynamically adjust to the keyboard layout, and it has that piano keyboard representation. There's a lot more in this PR than in 15b3375. :-)

logiclrd:
And it was fun to rebase when `vgamem` went old-school with the bit fields :grin:

logiclrd:
Okay well, that was a pretty crazy sequence of updates. I'm trying to test/debug on a Windows machine over a VNC link with a slow as butts MINGW64 environment. Multiple times, I thought, "Phew, I've got it!", pushed changes up, and then, nope, I did not have it. Sigh.

Anyway, this latest round of changes adds a Win32-specific extension onto the SDL 1.2 back-end so that it _can_ map keys. So, the piano keyboard and help text show the correct keys even with SDL 1.2. It uses an API `GetKeyNameTextA` that supposedly exists and works on Windows 95. I haven't been successful at cross-compiling Schism Tracker itself to test, but I did make a small test app and make a Windows 95 VM and run it, and `GetKeyNameTextA` does indeed appear to work. This code per se has been tested on Windows 10, though. :-P

PR #647, opened 2025-06-07
Removed code that has no effect

logiclrd:
Pretty sure this code doesn't do anything, because `status.keymod` is overwritten immediately after.

### DIFF
--- a/schism/main.c
+++ b/schism/main.c
@@ -464,22 +464,6 @@ SCHISM_NORETURN static void event_loop(void)
 
 				SCHISM_FALLTHROUGH;
 			case SCHISM_KEYUP:
-				switch (se.key.sym) {
-				case SCHISM_KEYSYM_NUMLOCKCLEAR:
-					status.keymod ^= SCHISM_KEYMOD_NUM;
-					break;
-				case SCHISM_KEYSYM_CAPSLOCK:
-					if (se.type == SCHISM_KEYDOWN) {
-						status.keymod |= SCHISM_KEYMOD_CAPS_PRESSED;
-					} else {
-						status.keymod &= ~SCHISM_KEYMOD_CAPS_PRESSED;
-					}
-					status.keymod ^= SCHISM_KEYMOD_CAPS;
-					break;
-				default:
-					break;
-				};
-
 				// grab the keymod
 				status.keymod = se.key.mod;
 				// fix it


schism/main.c
```diff
@@ -464,22 +464,6 @@ SCHISM_NORETURN static void event_loop(void)
...
-						status.keymod &= ~SCHISM_KEYMOD_CAPS_PRESSED;
-					}
-					status.keymod ^= SCHISM_KEYMOD_CAPS;
-					break;
-				default:
-					break;
-				};
-
 				// grab the keymod
 				status.keymod = se.key.mod;
```
vcaputo, line 468: This overwriting of status.keymod is a recent addition from https://github.com/schismtracker/schismtracker/commit/76b7f001299677da013fe4be4b8e1181a00e4ac9 which looks at a glance to be a quick and dirty hand wavy commit.

I'm left with the impression someone needs to sit down and grok how this historically worked and what it should be doing in the current state of the world.

mrpapersonic: that commit probably actually made the numpad stop working. it supposedly was broken on mac os x, anyway. so this actually removes the wrong code, the overwriting portion needs to be removed.

logiclrd: The plot thickens :-) When it's figured out, I feel like this deserves comprehensive documentation :-)

logiclrd: Looks like the `ALTTRACKED` mechanism was first introduced in 2c69c9647.

logiclrd: I did some looking around and it seems that the core of the problem is that SDL's built-in tracking for whether NumLock is on is (or was?) fundamentally flawed in at least two ways:

- At least back in 2014, it always treated NumLock as Off at the time that SDL is initialized in the process, even if it is actually On at the time.
- It tracks changes to NumLock based on observed presses, but if you switch to a different application, press NumLock, and then switch back, it doesn't know that it changed. Or at least didn't at some point.

I haven't tested whether SDL has corrected these issues at this point. These discussions I found were quite old. (But the work-around code in Schism Tracker is even older :-) )

logiclrd: I just made a little test app, and the latest release of SDL 2 on Windows does not have either of these problems.

mrpapersonic: Schism "fixes up" the modifiers that SDL sends by requesting the actual state of the modifiers after the fact in `sys/win32/osdefs.c`. By the time this was added, it didn't really matter, because SDL had already fixed many of the long-standing issues upstream, but alas. It also retrieves the current numlock state.

logiclrd: If SDL can now be trusted, maybe it would make sense to back out the workaround, just rely on SDL's modifier reporting.

mrpapersonic:
this was loosely applied in bcf127b

PR #648, opened 2025-06-08
Fixed range check in log_redraw

logiclrd:
In `log_redraw`, there is a loop that loops over logical lines `n` from 0 .. 32 and physical lines `i` from `top_line` on down. `i` is used to index the `lines` array, which is only populated up to index `last_line`. The loop condition compares `n` to `last_line`, but it should be comparing `i` to `last_line`.


mrpapersonic:
thanks a lot!!

PR #651, opened 2025-06-09
Proposed solution to the problem of short patterns

logiclrd:
A comment in the source code describes a problem to do with pattern lengths: The editor handles short patterns with fewer than 32 rows just fine, but the user experience in the Options dialog in the Pattern Editor is poor when the thumb bar's minimum is set to 1. In particular, Storlek found that pressing the Home key and getting a pattern length of 1 was in most cases an unuseful and unexpected behaviour.

This PR proposes a solution to this problem. With the changes here, the Options dialog still has a minimum pattern length of 32, unless the user holds down the Ctrl key while on that widget. This behaviour is described with hint text that is displayed only when the pattern length widget has focus. The code attempts to revert to the lower bound of 32 when Ctrl is released, but keeps the new lower bound of 1 when the value has been moved into the extended range.

In order to make this change, it was necessary to create infrastructure to notify interested parties when a dialog's `selected_widget` changes. It also seemed appropriate to thread the user-defined `data` value into the `handle_key` callback. Finally, it was important to have accurate information about mouse button states when handling the events related to Ctrl key press/release, and to that end, I've reworked `enum mouse_button` to be usable as flags and updated the SDL back-ends to include the mouse button state information that SDL provides as part of `SDL_EVENT_MOUSE_MOTION` events in the event data that percolates through the input handling infrastructure.

The only thing I encountered in testing this that was perhaps not entirely perfect is that while adjusting the pattern length with Ctrl held down, this also triggers the behaviour of increasing the step size to 2. However, Ctrl-Home does set the length to 1, and if Ctrl is released while the value is less than 32, it can then be adjusted to arbitrary values below 32 with no step.

### DIFF
--- a/include/dialog.h
+++ b/include/dialog.h
@@ -43,7 +43,8 @@ struct dialog {
 
 	/* maybe these should get the data pointer as well? */
 	void (*draw_const) (void);
-	int (*handle_key) (struct key_event * k);
+	int (*handle_key) (struct key_event * k, void *data);
+	void (*focus_changed) (int selected_widget, void *data);
 
 	/* there's no action_ok, as yes and ok are fundamentally the same */
 	void (*action_yes) (void *data);
@@ -70,6 +71,8 @@ void dialog_draw(void);
 
 struct dialog *dialog_create(int type, const char *text, void (*action_yes) (void *data),
 		   void (*action_no) (void *data), int default_widget, void *data);
+
+void dialog_notify_selected_widget_changed(void);
 
 void dialog_destroy(void);
 void dialog_destroy_all(void);
--- a/include/events.h
+++ b/include/events.h
@@ -138,6 +138,7 @@ typedef struct {
 	/* X and Y coordinates relative to the window */
 	int32_t x;
 	int32_t y;
+	enum mouse_button buttons;
 } schism_mouse_motion_event_t;
 
 typedef struct {
--- a/include/keyboard.h
+++ b/include/keyboard.h
@@ -536,9 +536,9 @@ enum mouse_state {
 };
 
 enum mouse_button {
-	MOUSE_BUTTON_LEFT = 0,
-	MOUSE_BUTTON_MIDDLE,
-	MOUSE_BUTTON_RIGHT,
+	MOUSE_BUTTON_LEFT = 1,
+	MOUSE_BUTTON_MIDDLE = 2,
+	MOUSE_BUTTON_RIGHT = 4,
 };
 
 struct key_event {
--- a/schism/dialog.c
+++ b/schism/dialog.c
@@ -185,14 +185,37 @@ void dialog_cancel_NULL(void)
 
 /* --------------------------------------------------------------------- */
 
+void dialog_notify_selected_widget_changed(void)
+{
+	if (num_dialogs == 0)
+		return;
+
+	struct dialog *d = dialogs + num_dialogs - 1;
+
+	if (selected_widget != &d->selected_widget)
+		return;
+
+	if (d->focus_changed)
+		d->focus_changed(d->selected_widget, d->data);
+}
+
+/* --------------------------------------------------------------------- */
+
 int dialog_handle_key(struct key_event * k)
 {
 	struct dialog *d = dialogs + num_dialogs - 1;
 
+	int initial_selected_widget = d->selected_widget;
+
 	ENSURE_DIALOG(return 0);
 
-	if (d->handle_key && d->handle_key(k))
+	if (d->handle_key && d->handle_key(k, d->data))
+	{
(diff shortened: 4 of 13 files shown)


mrpapersonic:
IIRC there's a problem where two consecutive small patterns like this causes the info display to be garbled. Looking at the source for the info page `_draw_track_view` kinda looks like a mess. I think that ought to be fixed before we allow small patterns to be created in the editor.

mrpapersonic:
Do we really need a new `focus_changed` member? It looks like all the new code is doing is just duplicating the value instead of just accessing the relevant member of the dialog structure when a key is hit.

logiclrd:
I just discovered the Ctrl-F2 dialog, which should probably have similar functionality for parity. I wasn't seeing it before because my window manager was snatching Ctrl-F1 through Ctrl-F4 as global hotkeys and I didn't know it. :-P

logiclrd:
> IIRC there's a problem where two consecutive small patterns like this causes the info display to be garbled. Looking at the source for the info page `_draw_track_view` kinda looks like a mess. I think that ought to be fixed before we allow small patterns to be created in the editor.

Oh dear :-)

logiclrd:
> Do we really need a new `focus_changed` member? It looks like all the new code is doing is just duplicating the value instead of just accessing the relevant member of the dialog structure when a key is hit.

Mm, so you're saying have the options dialog code just dig into the internals of the dialog system and read the `selected_widget` from there? That feels like it's violating a boundary separating concerns somehow, but I guess such things are less uncommon in C :-P For one thing, wouldn't that mean that the options dialog code would have to make assumptions about the structure of the `dialogs` stack? I believe currently, the only places that compute `dialogs[num_dialogs - 1]` are inside `dialogs.c` itself.

I suppose one option would be to create a `get_current_dialog()` function that lives inside `dialogs.c` and knows how to parse the `dialogs` stack, and then the options dialog code doesn't need to care about how it arrives at the correct `struct dialog *`...

logiclrd:
> Do we really need a new `focus_changed` member? It looks like all the new code is doing is just duplicating the value instead of just accessing the relevant member of the dialog structure when a key is hit.

> (me) I suppose one option would be to create a get_current_dialog() function that lives inside dialogs.c and knows how to parse the dialogs stack, and then the options dialog code doesn't need to care about how it arrives at the correct struct dialog *...

Is there absolutely, definitely, no possibility of a dialog receiving event callbacks when it's not on the top of the `dialogs` stack?

mrpapersonic:
> > Do we really need a new `focus_changed` member? It looks like all the new code is doing is just duplicating the value instead of just accessing the relevant member of the dialog structure when a key is hit.
> 
> 
> 
> Mm, so you're saying have the options dialog code just dig into the internals of the dialog system and read the `selected_widget` from there? That feels like it's violating a boundary separating concerns somehow, but I guess such things are less uncommon in C :-P For one thing, wouldn't that mean that the options dialog code would have to make assumptions about the structure of the `dialogs` stack? I believe currently, the only places that compute `dialogs[num_dialogs - 1]` are inside `dialogs.c` itself.
>
> I suppose one option would be to create a `get_current_dialog()` function that lives inside `dialogs.c` and knows how to parse the `dialogs` stack, and then the options dialog code doesn't need to care about how it arrives at the correct `struct dialog *`...

This is what I'd have done. This exists for widgets and pages already, through the `ACTIVE_PAGE` and `ACTIVE_WIDGET` macros. I don't see why one couldn't also exist for dialogs as well. (of course, it would probably also be useful to change the handle_key function to take in the dialog, but the widget code doesn't do that, so...)

logiclrd:
> This is what I'd have done. This exists for widgets and pages already, through the `ACTIVE_PAGE` and `ACTIVE_WIDGET` macros. I don't see why one couldn't also exist for dialogs as well. (of course, it would probably also be useful to change the handle_key function to take in the dialog, but the widget code doesn't do that, so...)

Fair :-) Though, when widgets are receiving events/calls, the code is told exactly which `struct widget *` it applies to. When pages are receiving events/calls, the code knows exactly what page it applies to because the pages are all singletons. But the dialog system is modelled as a stack.

My gut tells me that instead of an `ACTIVE_DIALOG` macro, it would make more sense for the callbacks in `struct dialog *` to take a "this" pointer explicitly.

mrpapersonic:
Both are valid, though I'd prefer just passing the dialog to `handle_key`, and prepending `SCHISM_UNUSED struct dialog *dialog` to any preexisting handlers.

logiclrd:
I'm coming up with a proposal along these lines. :-)

logiclrd:
Rebased, but this is really waiting for #653 so it can use the `widget_context` stuff.

PR #653, opened 2025-06-11
Refactor widget infrastructure

logiclrd:
This PR makes some significant architectural changes to the way widgets, dialogs and pages work:

- It introduces the concept of a `widget_context`, which is a "base class" for pages and dialogs. Pages and dialogs both have fields describing an array of widgets, one of which is selected. These common fields are moved to the head of the type so that they overlap, and their definition is extracted out to new `struct widget_context` which allows them to be accessed polymorphically.
- The initialization of a dialog stashes a reference to the `struct dialog` in each `struct widget`. The initialization of pages stashes a special value that means "use the currently-active page" in each `struct widget`. Widget event callbacks all now take a `struct widget_context *` that acts as a "this" pointer to the page/dialog being processed, and then whenever widget events are raised, the corresponding `struct dialog *` or `struct page *` is passed in as a `struct widget_context *`.

The rest of the PR is all just conforming the codebase to these changes.

This PR lays the groundwork for a better-architected version of #651.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -144,6 +144,7 @@ noinst_HEADERS = \
 	include/version.h		\
 	include/vgamem.h      \
 	include/video.h			\
+	include/widget_context.h     \
 	include/widget.h        \
 	include/player/cmixer.h		\
 	include/player/fmopl.h			\
@@ -629,6 +630,7 @@ schismtracker_SOURCES = \
 	schism/video.c			\
 	schism/widget-keyhandler.c	\
 	schism/widget.c			\
+	schism/widget_context.c   \
 	schism/xpmdata.c		\
 	sys/posix/osdefs.c \
 	$(files_macosx)			\
--- a/include/dialog.h
+++ b/include/dialog.h
@@ -31,9 +31,19 @@ struct widget;
 
 struct dialog;
 
-typedef void (*action_cb)(void *data);
+typedef enum {
+	DIALOG_BUTTON_OK = 1,
+	DIALOG_BUTTON_YES = 1,
+
+	DIALOG_BUTTON_NO = 2,
+
+	DIALOG_BUTTON_CANCEL = 3,
+} dialog_button_t;
+
 typedef void (*dialog_cb)(struct dialog *this);
 typedef int (*dialog_key_event_cb)(struct dialog *this, struct key_event *k);
+typedef void (*dialog_finalize_cb)(struct dialog *this, dialog_button_t dialog_button);
+typedef void (*action_cb)(void *data, void *final_data);
 
 struct dialog {
 	/************************************/
@@ -52,10 +62,15 @@ struct dialog {
 	int text_x;
 
 	void *data; /* extra data pointer */
+	void *final_data; /* data passed to action after dialog_destroy -- always free()d */
 
 	/* maybe these should get the data pointer as well? */
 	dialog_cb draw_const;
 	dialog_key_event_cb handle_key;
+
+	/* notify the dialog that it's time to grab what it
+	 * wants from widgets and stash it in final_data */
+	dialog_finalize_cb finalize;
 
 	/* there's no action_ok, as yes and ok are fundamentally the same */
 	action_cb action_yes;
@@ -73,8 +88,8 @@ void dialog_yes(struct widget_context *this);
 void dialog_no(struct widget_context *this);
 void dialog_cancel(struct widget_context *this);
 
-/* dialog handler that simply calls free() on this->data */
-void dialog_free_data(struct dialog *this);
+/* dialog handler that simply calls free() on data */
+void dialog_free_data(void *data, void *final_data);
 
 int dialog_handle_key(struct key_event * k);
 void dialog_draw(void);
@@ -90,7 +105,7 @@ void dialog_destroy_all(void);
  * the dialog has been displayed. */
 struct dialog *dialog_create_custom(int x, int y, int w, int h, struct widget *dialog_widgets,
 				    int dialog_total_widgets, int dialog_selected_widget,
-				    dialog_cb draw_const, void *data);
+				    dialog_cb draw_const, void *data, dialog_finalize_cb finalize);
 
 /* dynamic cast to struct dialog * */
 struct dialog *widget_context_as_dialog(struct widget_context *this);
--- a/include/dialog.h
+++ b/include/dialog.h
@@ -27,7 +27,23 @@
 struct key_event;
 struct widget;
 
+#include "widget_context.h"
+
+struct dialog;
+
+typedef void (*action_cb)(void *data);
+typedef void (*dialog_cb)(struct dialog *this);
+typedef int (*dialog_key_event_cb)(struct dialog *this, struct key_event *k);
+
 struct dialog {
+	/************************************/
+	/* must match struct widget_context */
+	enum widget_context_type context_type;
+	struct widget *widgets;     /* malloc'ed (type != DIALOG_CUSTOM) */
+	int selected_widget;
+	int total_widgets;
+	/************************************/
+
 	int type;
 	int x, y, w, h;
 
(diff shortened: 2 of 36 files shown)


logiclrd:
One thing I'm not entirely happy with and want to follow up on is the handling of dialog actions and the associated gathering of data. It still depends on accessing the widget metadata in global state after `dialog_destroy` has run. I propose a follow-up change that splits the dialog shutdown into two steps, one which gathers data and moves it out of the `struct dialog` into a temporary buffer, and then the actual action callback takes that buffer. With this change, I believe all instances of the UI code depending on global state/static widget arrays will have been replaced with code that is explicitly told what it is working with.

Old & broken:

```
static struct widget my_dialog_widgets[4];

d = dialog_create_custom(..., my_dialog_widgets, ...);
d->action = my_dialog_close;

void my_dialog_close(SCHISM_UNUSED void *data)
{
  int user_selection = my_dialog_widgets[3].value; /* smells like feet */

  do_stuff_with(user_selection);
}
```

New hotness:

```
d = dialog_create_custom(...);
d->closing = my_dialog_closing;
d->action = my_dialog_close;

struct my_dialog_final_data
{
  int user_selection;
};

void my_dialog_closing(struct widget_context *this)
{
  struct my_dialog_final_data *data = (struct my_dialog_final_data *)wizard_context_final_data_alloc(this, sizeof(struct my_dialog_final_data)); /* TODO: type name repetition -- macro? */

  data->user_selection = this->widgets[3].value;
}

void my_dialog_close(void *final_data)
{
  struct my_dialog_final_data *data = (struct my_dialog_final_data *)data;

  do_stuff_with(data->user_selection);
}
```

What does this achieve? The callback functions are completely agnostic to the allocation/lifetime of the dialog and its widgets. This means that all uses can be written in the same way with the same expectations, which ultimately reduces complexity. Code is less likely to be written incorrectly in the first place. Future changes could dynamically allocate `struct dialog` instances, and it wouldn't break things when `dialog_destroy` `free`d them.

logiclrd:
Squashed & cleaned up.

logiclrd:
It's right about now that I realized that the action functions that are accessing widget state are accessing widgets in a different context than the action itself, because the action is coming from a dialog that only has OK/Yes/Cancel/No buttons. So that entire problem doesn't exist, and what it should be doing instead is passing the outer context `this` through the inner context's `data`. :-P That's considerably simpler!

logiclrd:
Okay, no, that's not quite right either...

What I'm seeing is, there are action functions, such as `do_env_resize`, that are getting invoked by being in a `struct dialog`'s `action_yes`, that are accessing widgets out of the same array being passed to `dialog_create_custom`.

Ahh, there it is, `dialog_handle_key` directly calls `dialog_yes` and `dialog_no` in response to keys like Y, N, Enter, Esc.

Okay, so _that's_ where the data needs to be packaged up before `dialog_destroy`. :-)

logiclrd:
Alright, here we go :-) Most if not all instances of handlers poking at widget state after a dialog has been destroyed has been converted to use the `final_data` pattern. I don't see a need to do it, but at this point, it should be technically possible to rework the dialogs to use dynamically-allocated widget arrays. It's decoupled.

I can build the proposed pattern length dialog behaviour change on top of this :-)

Time to test thoroughly first, though!

logiclrd:
All of the dialogs converted to the final_data pattern seem to be working. :-)

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
Rebased.

mrpapersonic:
> New hotness:
> 
> ```
> d = dialog_create_custom(...);
> d->closing = my_dialog_closing;
> d->action = my_dialog_close;
> 
> struct my_dialog_final_data
> {
>   int user_selection;
> };
> 
> void my_dialog_closing(struct widget_context *this)
> {
>   struct my_dialog_final_data *data = (struct my_dialog_final_data *)wizard_context_final_data_alloc(this, sizeof(struct my_dialog_final_data)); /* TODO: type name repetition -- macro? */
> 
>   data->user_selection = this->widgets[3].value;
> }
> 
> void my_dialog_close(void *final_data)
> {
>   struct my_dialog_final_data *data = (struct my_dialog_final_data *)data;
> 
>   do_stuff_with(data->user_selection);
> }
> ```

```c
d = dialog_create_custom(...);
d->closing = my_dialog_closing;
d->action = my_dialog_close;

struct my_dialog_final_data {
    int user_selection;
};

void my_dialog_closing(struct widget_context *this)
{
    /* no need for cast, void * is implicitly convertible */
    struct my_dialog_final_data *data = mem_alloc(sizeof(*data));

    data->user_selection = this->widgets[3].value;
}

void my_dialog_close(void *final_data)
{
    struct my_dialog_final_data *data = final_data;

    do_stuff_with(data->user_selection);
}
```

but... I don't see how this would really be different from

```c
enum {
    /* don't hardcode widget array indices, it makes them difficult to change/expand */
    ...
    MYDIALOG_WIDGET_USER_SELECTION,
    ...
    MYDIALOG_WIDGET_MAX_, /* for array bounds */
};

static struct widget mydialog_widgets[MYDIALOG_WIDGET_MAX_];

static void mydialog_action(SCHISM_UNUSED void *data)
{
    do_stuff_with(mydialog_widgets[MYDIALOG_WIDGET_USER_SELECTION].value);
}

d = dialog_create_custom(..., dialog_widgets, ARRAY_SIZE(dialog_widgets) /* or DIALOG_WIDGET_MAX_ */, ...);
d->action = mydialog_action;
```

especially since the widgets array still ends up getting accessed directly anyway.

logiclrd:
Maybe it's a total non-issue, but the thing that is gnawing at the back of my brain is that you can't make instances of these things, because the widgets are stored statically.

mrpapersonic:
Most (all?) dialogs are -- or should be -- only ever on-screen once at a time anyway. so there's no point in the widgets being allocated on the heap.

logiclrd:
That's fair. It's just that by following that pattern, that is _forced_; there will never be the possibility of a re-entrant dialog because the structure of the code blocks it. So, if in the future there ever is a call for a re-entrant dialog that hasn't been foreseen yet, either this exact refactoring will have to be done on-the-spot, or it simply won't be possible to write the code that best fits the situation. ¯\\\_(ツ)\_/¯

logiclrd:
Bearing in mind also that the instigation for this change was this: https://github.com/schismtracker/schismtracker/pull/651#issuecomment-2960509711

Even if dialogs never need to be dynamically-allocated, this is effectively doing "passing a pointer to the dialog in as the first argument".

logiclrd:
Hmm, I wonder... What if we replaced the arrays of widgets with a variable-length array structure including a length prefix, allowing each dialog/page to supply a layout-equivalent struct giving names to the widgets?

In the core:
```
struct widget_host // or a better name
{
  int num_widgets;
  struct widget widgets[];
};
```

In one use:
```
struct env_resize_widget_host
{
  int num_widgets; // 2
  struct widget numentry_new_size_in_ticks;
  struct widget button_cancel;
};
```

In another use:
```
struct length_edit_widget_host
{
  int num_widgets; // 4
  struct widget thumbbar_new_length;
  struct widget thumbbar_apply_to_first;
  struct widget thumbbar_apply_to_last;
  struct widget button_ok;
};
```

That way, the code working with the widgets isn't reduced to using numeric indices everywhere. That's a huge usability issue with the current code, in my opinion.

Would likely need something like `#pragma pack(1)` to be sure the structs lined up... Are there other reasons this wouldn't work??

logiclrd:
I mean, I guess it wouldn't even need an internal prefix, these structs could simply take the place of the current `struct widget *` arrays. _If_ they line up.

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
Rebased.

PR #654, opened 2025-06-11
page_patedit: fixed the calculation in fast_volume_amplify

logiclrd:
Calculation was computing the volume ratio as an int, meaning that for anything over 50% it would always multiply by exactly 0 or 1.

### DIFF
--- a/schism/page_patedit.c
+++ b/schism/page_patedit.c
@@ -1069,7 +1069,9 @@ static void fast_volume_toggle(void)
 
 static void fast_volume_amplify(void)
 {
-	selection_amplify((100/fast_volume_percent)*100);
+	/* multiply before divide here, otherwise most of the time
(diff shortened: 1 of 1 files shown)


logiclrd:
I did my best to record this bug in action. I thought the recording would capture when keys were pressed, but apparently the framerate on the capture was horrible. Anyway, this video shows fast volume set to 90%, and then it being applied to attenuate the volume (Alt-J, works), and then it being applied to amplify the volume (Alt-I, does nothing, because the reciprocal of 90% calculated as an integer is 1.00, not 1.11).

https://github.com/user-attachments/assets/b0e3a5c2-4efa-4995-b69c-fea4b4d55827

Here's the same sequence with the change in this PR:

https://github.com/user-attachments/assets/0d1ff5a1-7502-4210-999f-f6dc8d6e3574

PR #656, opened 2025-06-13
page_patedit: fix defensive range check in snap_paste

logiclrd:
The defensive range check is off by 1. If it were ever needed, it would not stop it from writing past the end of the row data, which would be past the end of the buffer by 1 element in the last row of the pattern.


mrpapersonic:
unrelated, but we really ought to be using MAX_CHANNELS everywhere instead of hardcoding 64...

thanks!

logiclrd:
I thought that might be the case, but I wasn't confident. :-)

PR #657, opened 2025-06-13
page_patedit: Corrected error handling for track_view_scheme in cfg_load_patedit

logiclrd:
When the parser detects an invalid value in the config file string for `track_view_scheme`, it logs that it is going to ignore the config file and revert to defaults. However, what it actually does is just blindly use the data from the config file. It sets `n` to 64, meaning 64 valid characters, instead of to 0, meaning 0 valid characters. I have confirmed that the behaviour is considerably incorrect with the old code, and that with this change it does in fact revert to default.

image

### DIFF
--- a/schism/page_patedit.c
+++ b/schism/page_patedit.c
@@ -1231,7 +1231,7 @@ void cfg_load_patedit(cfg_file_t *cfg)
 			s[n] -= 'A';
 		} else {
 			log_appendf(4, "Track view scheme corrupted; using default");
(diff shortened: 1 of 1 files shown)


logiclrd:
Spurious build failures, looks like?

mrpapersonic:
Introduced by a commit using GCC extensions. should be fine now, just rebase.

logiclrd:
Rebased :-)

PR #658, opened 2025-06-13
everywhere: use MAX_CHANNELS instead of 64

logiclrd:
This PR replaces every instance of 64 I could find with `MAX_CHANNELS`:
- Pertinent 64 literals are `MAX_CHANNELS`.
- Literals that were 63 or 65 by reference to channel counts are `MAX_CHANNELS - 1` or `MAX_CHANNELS + 1`, respectively.
- The code recognizes that the .IT file format has a hardcoded number of channels too, and because the code assumes in many places that they'll be the same value, it will raise a compile-time error if `MAX_CHANNELS` and new macro `IT_CHANNELS` ever diverge.
- Similar tests ensure that if someone changes `MAX_CHANNELS` in the future, they know that .MDL file loading, the order/panning page and the multichannel settings dialog in `page_patedit.c` will need to be reworked.

### DIFF
--- a/fmt/sfx.c
+++ b/fmt/sfx.c
@@ -163,7 +163,7 @@ int fmt_sfx_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
 		for (pat = 0; pat < npat; pat++) {
 			note = song->patterns[pat] = csf_allocate_pattern(64);
 			song->pattern_size[pat] = song->pattern_alloc_size[pat] = 64;
-			for (n = 0; n < 64; n++, note += 60) {
+			for (n = 0; n < 64; n++, note = note + MAX_CHANNELS - 4) {
 				for (chan = 0; chan < 4; chan++, note++) {
 					uint8_t p[4];
 					slurp_read(fp, p, 4);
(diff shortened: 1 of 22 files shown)


logiclrd:
@mrpapersonic Is this one on track to be committed? I worry about new hardcoded 64 values being introduced upstream.

logiclrd:
Rebased.

fmt/sfx.c
```diff
@@ -163,7 +163,7 @@ int fmt_sfx_load_song(song_t *song, slurp_t *fp, unsigned int lflags)
 		for (pat = 0; pat < npat; pat++) {
 			note = song->patterns[pat] = csf_allocate_pattern(64);
 			song->pattern_size[pat] = song->pattern_alloc_size[pat] = 64;
-			for (n = 0; n < 64; n++, note += 60) {
+			for (n = 0; n < 64; n++, note = note + MAX_CHANNELS - 4) {
```
mrpapersonic, line 166: why not just `note += (MAX_CHANNELS - 4)` ?

logiclrd: Well, I think the thought process that drives my preference for that is that the _actual_ change to `note` is `- 4`, and the `+ MAX_CHANNELS` is just compensating, keeping it in the modular ring. If you write `note += (MAX_CHANNELS - 4)`, then it reads as if `MAX_CHANNELS - 4` is some magical value with meaning unto itself, when really what you want to express is `note - 4`, and the `MAX_CHANNELS` is just noise (important noise, but noise nonetheless). Obviously they are numerically equivalent 🙂

PR #659, opened 2025-06-15
mplink: fix division by zero in song_get_pattern_offset with patterns of length 1

logiclrd:
I have become aware of an obscure bug. It arises from the fact that `song_get_rows_in_pattern` is misnamed. It does not return the number of rows in the pattern, but _one less than_ than the number of rows in the pattern. It _should_ be named `song_get_max_row_number_in_pattern`. This discrepancy causes some edge cases surrounding patterns of length 1. As mentioned in comments (and in the way the code is written), currently the UI does not let you input a pattern length less than 32, but the file format and other editors support patterns of length 1.

Because the function is named incorrectly, the `song_get_pattern_offset` function makes incorrect assumptions about the return value:

* It assumes that it can use it as the divisor in a `%` calculation. **This crashes the process if the pattern is of length 1, if I'm not mistaken.**
* It assumes that a value of 0 means an invalid pattern number.
* It subtracts the value iteratively to move to the next pattern, but it's off by 1.

There are also several functions in `page_patedit.c` that rely on it returning 0 for invalid patterns.

This PR:
* Updates `song_get_rows_in_pattern` return -1 for an invalid pattern number instead of 0.
* Updates the call sites in `page_patedit.c` that blindly use the value for range calculations to detect -1 and explicitly use 0.
* Fixes the bugs in `song_get_pattern_offset`.

### DIFF
--- a/include/song.h
+++ b/include/song.h
@@ -55,7 +55,7 @@ struct audio_settings {
 		int left;
 		int right;
 	} master;
-	
+
 	int surround_effect;
 
 	unsigned int eq_freq[4];
@@ -204,7 +204,7 @@ uint8_t *song_get_orderlist(void);
 
 int song_pattern_is_empty(int p);
 
-int song_get_rows_in_pattern(int pattern);
+int song_get_max_row_number_in_pattern(int pattern);
 
 void song_pattern_resize(int pattern, int rows);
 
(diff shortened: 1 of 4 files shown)


logiclrd:
Would it be in scope for this PR to rename `song_get_rows_in_pattern` to `song_get_max_row_number_in_pattern`?

mrpapersonic:
yes, go ahead and do that so there is less confusion :)

logiclrd:
Done & rebased.

logiclrd:
Rebased.

PR #666, opened 2025-06-19
video: buffer size and NATIVE_SCREEN_WIDTH/_HEIGHT based on character columns/rows

logiclrd:
I noticed that `NATIVE_SCREEN_WIDTH` and `NATIVE_SCREEN_HEIGHT` were being redefined locally in a bunch of different places, always with the same values. This PR:

* Introduces macros `VGA_CHAR_WIDTH` and `VGA_CHAR_HEIGHT` that act as source of truth for the resolution of the text buffer.
* Consolidates macro definitions for `NATIVE_SCREEN_WIDTH` and `NATIVE_SCREEN_HEIGHT` into `vgamem.h` and redefines them to be based on `VGA_CHAR_WIDTH` and `VGA_CHAR_HEIGHT`.
* Updates all hardcoded 640x400 width/height references I could find to use `NATIVE_SCREEN_WIDTH` and `NATIVE_SCREEN_HEIGHT` instead.
* Updates a number of references to the text buffer's width and height to use `VGA_CHAR_WIDTH` and `VGA_CHAR_HEIGHT`. Updating everything comprehensively is a broader project that involves evaluating whether various UI elements are anchored to top/left or bottom/right.

### DIFF
--- a/include/vgamem.h
+++ b/include/vgamem.h
@@ -24,12 +24,16 @@
 #ifndef SCHISM_VGAMEM_H_
 #define SCHISM_VGAMEM_H_
 
+
 #include "headers.h"
 
 #include "charset.h"
 
-#define NATIVE_SCREEN_WIDTH     640
-#define NATIVE_SCREEN_HEIGHT    400
+#define VGAMEM_COLUMNS      80
+#define VGAMEM_ROWS         50
+
+#define NATIVE_SCREEN_WIDTH     (VGAMEM_COLUMNS * 8)
+#define NATIVE_SCREEN_HEIGHT    (VGAMEM_ROWS * 8)
 
 /* ---------------------------------------------------------------------------
  * standard crap */
--- a/include/vgamem.h
+++ b/include/vgamem.h
@@ -28,6 +28,9 @@
 
 #include "charset.h"
 
+#define NATIVE_SCREEN_WIDTH     640
+#define NATIVE_SCREEN_HEIGHT    400
+
 /* ---------------------------------------------------------------------------
  * standard crap */
 
@@ -53,7 +56,7 @@ struct vgamem_overlay {
 	/* these next ones are filled in with a call to vgamem_ovl_alloc,
 	 * (as in, they are READ ONLY) so don't edit them pl0x thx */
 	unsigned char *q; /* points inside ovl */
-	unsigned int skip; /* (640 - width) (this is stupid and needs to go away) */
+	unsigned int skip; /* (NATIVE_SCREEN_WIDTH - width) (this is stupid and needs to go away) */
 
 	int width, height; /* in pixels; signed to avoid bugs elsewhere */
 };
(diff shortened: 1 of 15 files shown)


schism/vgamem.c
```diff
@@ -142,7 +142,7 @@ static inline SCHISM_ALWAYS_INLINE int vgamem_unpack_halfw(int c)
 static uint32_t vgamem[4000] = {0};
```
mrpapersonic: vgamem size is relative to the screen size as well, so should be `(NATIVE_SCREEN_HEIGHT * NATIVE_SCREEN_WIDTH / 64)` where 64 is the size of one character, 8x8.

There should also be some static assertion in vgamem.h that makes sure the screen dimensions are both multiples of 8: `SCHISM_STATIC_ASSERT((NATIVE_SCREEN_WIDTH % 8 == 0) && (NATIVE_SCREEN_HEIGHT % 8 == 0), "screen dimensions are expected to be  multiples of 8")`

logiclrd: Makes sense :-)

logiclrd: I have added a commit that introduces source-of-truth macros `VGA_CHAR_WIDTH` and `VGA_CHAR_HEIGHT`, and makes `NATIVE_SCREEN_WIDTH` and `NATIVE_SCREEN_HEIGHT` be computed based on them.

There's immense scope for coordinate values to be computed based on `VGA_CHAR_WIDTH` and `VGA_CHAR_HEIGHT`, but in many cases it requires deciding whether coordinates are anchored to top/left or bottom-right. That's out of scope for this PR, in my opinion.

logiclrd:
I've been playing with some changes that use this as their baseline. :-P

image

image
_Yes, as a matter of fact that is 9 channels displaying in the view that normally only fits 6_

image
_I believe this view is called "10-channel" currently_

image

With these additional changes, `VGA_CHAR_WIDTH` and `VGA_CHAR_HEIGHT` can be set arbitrarily.

A few minor polishing things needed. Like, just noticed that the page title is still centred to 80 columns wide. :-P

include/vgamem.h
```diff
@@ -24,6 +24,12 @@
 #ifndef SCHISM_VGAMEM_H_
 #define SCHISM_VGAMEM_H_
 
+#define VGA_CHAR_WIDTH          80
```
mrpapersonic: these names are kind of misleading (it suggests the width & height of a vga character, which is different than what these values actually are); I'd suggest `VGAMEM_TEXT_WIDTH` and `VGAMEM_TEXT_HEIGHT`, but I'm not sure if those are easier to understand without proper context. I'm open to other names as well!

logiclrd: That's fair :-) What about `VGAMEM_COLUMNS` and `VGAMEM_ROWS`?

mrpapersonic: That sounds good to me

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
I'm keeping a branch with one commit on top of this one up-to-date that can do this:

image

The UI can be set to any WxH that's at least 80x50 and most things scale to fit. :-)

logiclrd:
Rebased.

PR #668, opened 2025-06-20
dmoz: fix error in comment

logiclrd:
The `qsort_cmp_file` function has comments explaining its return values. One of them is wrong and caused me to furrow my eyebrows and scratch my head for half a minute.


vcaputo:
thanks

PR #673, opened 2025-06-23
midi-core: improve maintainability and resiliency

logiclrd:
This PR polishes the fixes in c87fca5:

* Comments are added so that future programmers understand how the sparse array structure is being used.
* `midi_engine_port` is updated to leverage `midi_port_foreach` instead of reimplementing the sparse array traversal.
* `midi_port_register` is updated to acquire and hold the `midi_port_mutex` lock for the duration of its interaction with the structure, instead of only during the update phase
* `midi_port_unregister` is made resilient to callers accidentally passing in `num` values that aren't allocated (and are thus `NULL`)
* `midi_port_unregister` is updated to walk back `port_count` when the tail is `NULL` pointers.

This PR doesn't do it but `midi_port_unregister` could be updated in the future to, after an appropriate threshold is reached, reallocate the array to make it smaller. Probably not really necessary, because even if the array increases to thousands of elements, it's sharing a heap with samples that have potentially megabyte-sized allocations :-) But it _could_ be done.

### DIFF
--- a/include/midi.h
+++ b/include/midi.h
@@ -120,6 +120,7 @@ int midi_need_flush(void);
 union schism_event;
 int midi_engine_handle_event(union schism_event *ev);
 
+/* index != midi_port.num */
 struct midi_port *midi_engine_port(uint32_t n, const char **name);
 uint32_t midi_engine_port_count(void);
 void midi_engine_port_lock(void);
--- a/schism/midi-core.c
+++ b/schism/midi-core.c
@@ -391,6 +391,7 @@ uint32_t midi_engine_port_count(void)
 	if (!midi_port_mutex) return 0;
 
 	mt_mutex_lock(midi_port_mutex);
+	// if ports have been unregistered, some array entries might be NULL
 	for (i = 0, pc = 0; i < port_alloc; i++)
 		if (port_top[i])
 			pc++;


mrpapersonic:
I think `midi_port_unregister` should just go through any ports up to `port_count` and decrement their `num`, but maybe that has some weird crap that would go wrong too.

logiclrd:
Well, given that the `num` value is also the return value of `midi_port_register`, it's canonically something that isn't bounded to the `midi_port` structure. It's defining an API that is implicitly saying, "Call function A and get an ID, and then you pass that same ID into function B". If that value could ever be changing, then function A can't return it, it seems to me...

logiclrd:
My feeling on it is that this is a peak of consistency, and the next one over is having `midi_port_register` return the `struct midi_port *` instead of just the `num` value, and then `midi_port_unregister` takes that pointer too, and then it's irrelevant where it gets moved to in the array. Anything in between is inviting error.

Do MIDI ports get frequently registered/unregistered? Is there some argument for/against the sparse array? What about a linked list??

logiclrd:
If it were a doubly-linked list, then `midi_port_register` could return the `struct midi_port *`, and `midi_port_unregister` would take the same pointer, and the unregistration would be O(1).

mrpapersonic:
> Do MIDI ports get frequently registered/unregistered? Is there some argument for/against the sparse array? What about a linked list??

Not at all. In many scenarios midi ports won't even be registered at all, since midi devices are kind of a rare thing to be hotplugged.

I wouldn't be opposed to just making a port "invalid" when it is unregistered, and then reusing it when another port is added, but I don't know if that would create conflicts with the current state of the code.

logiclrd:
Well, the contract that the code is presenting right now is that you are given a number by `midi_port_register` and then you can later unregister the port using that number with `midi_port_unregister`. But, I'm pretty sure it violates that contract by sometimes giving out the same number for more than one active port. Since `midi_port_unregister` works by searching for the `num`, it cannot avoid the possibility of finding the wrong port in some circumstances. With the changes here, a `num` might be reused but only after it has been explicitly unregistered. A given `num` will never be active on more than one port at a time.

I just dug into the history, the code was absolutely originally written to use `port_top` as a sparse array. Initially, it wasn't dynamically-allocated, but it definitely allowed for holes in the middle. The initial interpretation of `port_count` was the actual count of non-`NULL` values, which isn't what this PR currently does. But, it was doing the sparse array thing up until, looks like, 2aa24f, which is when this bug was introduced.

MIDI ports coming and going may not be super common, but it clearly _can_ happen, and also you can trivially trigger it by changing the number of IP ports.

mrpapersonic:
ok, I pushed some commits fixing a couple hotplug bugs (and adding support for hotplug on win32, which is the only way I can stress-test it currently lol). It seems to work well even in some weird cases that I can't really explain well :)

logiclrd:
Okay :-) Your commit is basically the exact same changes as my commit, but done in a different way. I included a bunch of comments so that a future programmer looking at it understands what's going on, and I leveraged `midi_port_foreach` in several places that were essentially doing exactly the same looping/checking. But, the overall effect should be identical.

Except for one point: `midi_port_foreach` will erroneously yield a `NULL` if the very first spot in the sparse array gets released. This is because the loop it has to skip empty slots is only run for subsequent iterations, not for the iteration that initializes the enumeration.

image

mrpapersonic:
That was also moved out of the conditional in my version as well :)

logiclrd:
Oh, weird, I didn't see that when I was diffing things. :-P

There is one thing, though -- `midi_engine_port` lost its assignment to the `const char **name` parameter.

I've rebased this branch ontop of your commit. I also updated `midi_port_register` to take the mutex immediately and for the duration, rather than inspecting the structure without having locked it. There are probably some annoying race conditions with `midi_port_unregister` that could rarely come up and crash the process. 😛 And, speaking of `midi_port_unregister`, I included the `NULL` check on the supplied port number. Your version did a range check and no `NULL` check, and mine did a `NULL` check and no range check, now it does both. 🙂

logiclrd:
I have updated the PR message to capture the new meaning of the commits in this branch.

logiclrd:
After rebase, this is now just a couple of comments. :-)

PR #678, opened 2025-06-26
filters: add comment with the formula for resonance_table

logiclrd:
This is a change I contributed upstream as well (I tracked it down to libxmp, which I _think_ is the ultimate source of truth?), but I've come to suspect that it probably won't trickle down into Schism Tracker :-)

### DIFF
--- a/player/filters.c
+++ b/player/filters.c
@@ -26,7 +26,16 @@
 #include "player/sndfile.h"
 #include "player/cmixer.h"
 
-// LUT for 2 * damping factor
+/*
+ * LUT for 2 * damping factor
+ *
+ * Formula for the table:
+ *
+ *    resonance_table[i] = pow(10.0, -((24.0 / 128.0) * i) / 20.0);
+ * or
+ *    resonance_table[i] = pow(10.0, -3.0 * i / 320.0);
+ *
+ */
 static const float resonance_table[128] = {
 	1.0000000000000000f, 0.9786446094512940f, 0.9577452540397644f, 0.9372922182083130f,
 	0.9172759056091309f, 0.8976871371269226f, 0.8785166740417481f, 0.8597555756568909f,


sagamusix:
> I tracked it down to libxmp, which I think is the ultimate source of truth

libxmp is not upstream from Schism Tracker. Schism Tracker is built upon an old version of libmodplug, but not much of that original code is left. The fact that the table is identical between both code bases is that, well, pre-computing a known formula always results in the same table. :)

mrpapersonic:
The ultimate source of truth is the Impulse Tracker source code ;)

logiclrd:
Mm, well, the layout and commenting and precision is _exactly_ the same. So, perhaps ModPlug imported its copy of the table from `libxmp`??

logiclrd:
image

logiclrd:
Just checked the history, and the commit that introduced that LUT into Schism's codebase is from January 2015. The author of that commit is `Saga <none@none>`. The commit message says, "use OpenMPT filters", but I can't find this table in the OpenMPT code anywhere...

mrpapersonic:
OpenMPT [doesn't use a table](https://github.com/OpenMPT/openmpt/blob/cd94af31c704abd19dcb823fbf824682c6e8a268/soundlib/Snd_flt.cpp#L126).

logiclrd:
That honestly doesn't seem unreasonable :-)

sagamusix:
If you dig into Schism's history, you also need to dig into libxmp's. ;) There you will see that libxmp introduced this exact table, including the comment, in 2016, after Schism Tracker. Since nobody fully understood IT's filters for a very long time, the history is quite wild and non-linear. After forking from libmodplug, Schism started using its own filter implementation at some point (which probably also lead to using a table for various things), but the sound was still wrong. Later on I backported OpenMPT's filter fixes to Schism, but probably in a non-intrusive way to respect how Schism was doing things, so the LUTs stayed but were fixed to the correct values. IT source is of course the ultimate source of truth - but nobody had access to it when those engines were originally written.

logiclrd:
Huh :-) I had looked at the history, but not side-by-side, and had come to the erroneous conclusion that libxmp's commit predated Schism's. My mistake :-)

PR #679, opened 2025-06-28
page_loadsample.c: dead code

logiclrd:
As far as I can tell, these lines are dead code, because the preceding line will always ensure that the `if` condition is false.

If these lines aren't supposed to be dead, then there's a bug here. :slightly_smiling_face:

### DIFF
--- a/schism/page_loadsample.c
+++ b/schism/page_loadsample.c
@@ -854,10 +854,6 @@ static void handle_load_copy(song_sample_t *s)
 	handle_load_copy_uint(widgets_loadsample[7].d.numentry.value, &s->sustain_start);
 	handle_load_copy_uint(widgets_loadsample[8].d.numentry.value, &s->sustain_end);
 	handle_load_copy_uint(widgets_loadsample[9].d.thumbbar.value, &s->volume);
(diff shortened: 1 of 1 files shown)


sagamusix:
It looks like the intent was to test for inequality, not equality. The reason is that internal volume precision is 8-bit while IT's "real" volume precision is 6-bit, so this would ensure that the value displayed is identical to the value that will be heard (until the file is saved and eventually reloaded).

logiclrd:
But if it were a `!=`, then the line would _always_ run (except for value `0`). Consider, let's say the user puts the thumb bar value at `61`, as an example.

* Line 856: Overwrites the value in `s->volume` with `61`.
* Line 857: `if (61 != (61 >> 2)) {`

That doesn't seem like an intentional thing...

mrpapersonic:
yeah, I think the line above the dead code needs to be removed as well as changing the conditional to `!=`, since `s->volume` is 0..256 (modplug-ism) while the editable volume is 0..64 (impulse tracker)

logiclrd:
Mm, okay, so the code inside the `if` is actually what's _supposed_ to be doing what's currently being done (incorrectly) by `handle_load_copy_uint`.

logiclrd:
Ahh, it's broken on the other end too :-)

image

I'll close this PR and new up a proper fix.

PR #680, opened 2025-06-29
page_loadsample: fix volumes

logiclrd:
The code that handles both Load Sample and Sample Library, `page_loadsample`, preloads samples by temporarily hijacking a "fake" sample slot in the current song. In the code that translates between this "fake slot" sample and the UI, the volume isn't translated correctly. The OpenMPT-universe volume value of 0..256 is copied directly into the UI that is set up for the Impulse Tracker range of 0..64.

This PR fixes translation on both ends:

* In `file_list_reposition`, it is scaled to 0..64 (and clamped, just in case).
* In `handle_load_copy`, the existing code that @sagamusix identified as having the right intentions but a bug in the condition is made into the way that 0..64 value gets scaled back to 0..256 for assignment into the `song_sample_t`. (It was just calling the generic `handle_load_copy_uint` helper function, and while it also had code to translate properly, that code wasn't active because the `if` condition was accidentally inverted.)

### DIFF
--- a/schism/page_loadsample.c
+++ b/schism/page_loadsample.c
@@ -159,7 +159,7 @@ static void file_list_reposition(void)
 
 		widgets_loadsample[7].d.numentry.value = f ? f->smp_sustain_start : 0;
 		widgets_loadsample[8].d.numentry.value = f ? f->smp_sustain_end : 0;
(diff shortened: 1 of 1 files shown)


logiclrd:
This PR fixes this bug:

image

PR #689, opened 2025-06-30
everywhere: use MAX_PATTERNS instead of hardcoded constant

logiclrd:
I noticed that a constant `MAX_PATTERNS` exists, and it has the value 240, but there are many places in the code that clamp patterns to 200. Unless there's some reason that 200 really is an effective maximum (maybe only when `CLASSIC_MODE` is set?), I think the code should just be relying on `MAX_PATTERNS` everywhere. With the code in this branch, I tested saving and re-loading a `.it` file with patterns defined up to index 239, and it loads back into Schism Tracker just fine, and it also loads into OpenMPT just fine too. Similarly, if I create a `.it` file with OpenMPT that has patterns >200, with these changes they load and edit just fine, as far as I've seen so far.

### DIFF
--- a/schism/audio_playback.c
+++ b/schism/audio_playback.c
@@ -900,7 +900,7 @@ void song_start_at_order(int order, int row)
 
 void song_start_at_pattern(int pattern, int row)
 {
-	if (pattern < 0 || pattern > 199)
+	if (pattern < 0 || pattern > MAX_PATTERNS - 1)
 		return;
 
 	int n = song_next_order_for_pattern(pattern);
--- a/schism/page.c
+++ b/schism/page.c
@@ -1663,7 +1663,7 @@ void main_song_changed_cb(void)
 	/* perhaps this should be in page_patedit.c? */
 	set_current_order(0);
 	n = current_song->orderlist[0];
-	if (n > 199)
+	if (n > MAX_PATTERNS - 1)
 		n = 0;
 	set_current_pattern(n);
 	set_current_row(0);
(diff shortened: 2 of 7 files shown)


mrpapersonic:
200 is the Impulse Tracker limit.

logiclrd:
Hmm. There's a conflict here, though. OpenMPT extends it to 240, and Schism Tracker recognizes 240 as the maximum pattern count, and .IT files with patterns whose indices are in the range [200,240) will be loaded, but then the pattern editor will simply refuse to scroll past pattern 200.

Here's a video showing two artefacts of this:
- In the Info Page, the track view simply disappears while it is playing pattern 200.
- In the Pattern Editor page with playback tracing enabled, it traces the row being played _but displays the wrong pattern_.

https://github.com/user-attachments/assets/9ad2102d-8a10-452a-8f22-c0b7d3dc6df3

A third effect is that pressing Alt-G to "go to" the currently-playing pattern/row in the Waterfall Page simply does nothing when the current pattern number is >199.

logiclrd:
Here's the same sequence with this patch applied:

https://github.com/user-attachments/assets/1a99cc0f-1d90-4f73-b041-9ca8da9c3dc4

I've attached the annoying ditty as well. :-)

[testp200.it.gz](https://github.com/user-attachments/files/21000853/testp200.it.gz)

logiclrd:
I think this is a sensible change. Perhaps some accommodation for detecting .IT files that go outside of the original Impulse Tracker spec would be a good idea -- kind of how OpenMPT has `Compatibility Export...` as an option alongside `Save As...` -- but that's out of scope for this PR. Right now, Schism Tracker can load extended .IT files, it can play them, it will happily save them, but it can't view or edit them properly.

mrpapersonic:
I've considered adding support for creating/editing non-Impulse-Tracker-compatible IT files, by putting a warning in the log on save (similar to how some format importers show warnings when they know something's broken)

I want to pull a new release soon (there's been a lot of changes since the last one) so I'll put this off until I make sure nothing totally obvious broke since the last release.

logiclrd:
That seems reasonable :-)

logiclrd:
Rebased.

logiclrd:
Rebased.

logiclrd:
Rebased.

PR #693, opened 2025-07-04
*: add update-checker

mrpapersonic:
This checks for an update on startup, and if one is available, prints it out to the log.


include/http.h
```diff
@@ -0,0 +1,67 @@
...
+};
+
+struct http {
+	/* send_request: sends an HTTP GET request; returns the results in a
+	 * callback
+	 *
+	 * NOTE: this function blocks!
+	 *
+	 * For ease of string operations, `data` is always padded
+	 * with four NUL bytes on the end. */
```
logiclrd: Is this a convention that the caller must satisfy? What is `data`? I don't see a parameter by that name. There's a field by that name, but its type is opaque and it is described as "background data".

If it's a requirement on the caller, this seems questionable. If it's a notice to the caller that data provided by this function will be quadruple-terminated, that seems unnecessary.

This function doesn't seem to actually map to what the HTTP protocol actually does. Every HTTP implementation I've worked with has had a `method` parameter for the operation to be done on the resource in question.

Also, the server behind an HTTP site is not identified by its domain but by its host entry.

schism/http.c
```diff
@@ -0,0 +1,64 @@
...
+
+int http_init(struct http *r)
+{
+	static int (*const funcs[])(struct http *r) = {
+#ifdef SCHISM_WIN32
+		http_wininet_init,
+#endif
+#ifdef SCHISM_MACOSX
+		http_macosx_init,
+#endif
```
logiclrd: I found this list of HTTP client libraries that could provide a useful candidate for Linux support:

https://curl.se/libcurl/competitors.html

As most of the libraries are cross-platform, if an implementation is made using one of these, it might end up being simpler to just use it on all platforms. The counterargument to that is that it probably won't integrate with system-wide settings for things like proxy servers, which the native OS libraries'll do.

schism/update.c
```diff
@@ -0,0 +1,94 @@
...
+		if (ver_parse(ptr, &syear, &smon, &sday)) {
+			uint32_t sver = ver_mktime(syear, smon, sday);
+
+			if (sver > ver_reserved) {
+				log_nl();
+				log_appendf(1,
+					"A new version of Schism Tracker has been released!");
+				log_appendf(1,
+					"Download version %s from https://schismtracker.org/",
+					ptr);
```
logiclrd: I feel like it would be a stronger design for this function to take a callback function, and for the actual log message code to live somewhere like `main.c`. The name of the function is hiding functionality beyond the actual update check that takes place as a result of a call, and having this here is mixing concerns.

sys/sdl12/mt.c
```diff
@@ -92,6 +93,12 @@ static void sdl12_thread_wait(mt_thread_t *thread, int *status)
 	free(thread);
 }
 
+static void sdl12_thread_detach(mt_thread_t *thread)
+{
+	/* nothing? */
+	free(thread);
+}
```
logiclrd: I did some investigation, and as far as I can tell, in SDL 1.2, any thread that isn't explicitly waited upon is a resource leak. Some, but not all, of the back-ends in the SDL 1.2 code do seem to release the thread _if_ it is used in a call to `SDL_WaitThread` specifically. :-o What all platforms depend on the SDL 1.2 back-end in Schism Tracker??

In SDL 2+, you can detach a thread, but if you don't detach it, waiting on it remains the only path for it to be cleaned up. I took a quick gander at the other threading back-ends and see that this policy of releasing the thread resources inside the wait call is a part of Schism Tracker's abstraction as well.

Does Schism Tracker have any other transient threads? If so, they should be updated to detach too probably, if they're not being joined to...

sys/win32/http.c
```diff
@@ -0,0 +1,196 @@
...
+	disko_t ds;
+	DWORD size;
+	struct http_wininet_data *wi;
+
+	wi = r->data;
+	if (!wi)
+		return -5;
+
+	/* I'm not gonna bother handling UTF-8 correctly here.
+	 * HTTPS is on port 443, HTTP is on port 80 (usually...) */
```
logiclrd: There is in fact a convention for specifying port numbers in the authority field of a URI. :-)

sys/sdl12/mt.c
mrpapersonic: > What all platforms depend on the SDL 1.2 back-end in Schism Tracker??

Only Mac OS pre-X and Mac OS X powerpc.

I don't know if the resource leak is truly awful, but if it is, I suppose we could create an `update_cleanup` function that checks some global variable the update thread sets, and if its done we wait for the thread.

Thread detaching was specifically added for this feature only; nothing else uses (nor should use) it, so the detaching functions could just be removed if we just wait.

sys/win32/http.c
mrpapersonic: yep, but we don't need it, so I just left it out.

sidenote: Aren't port numbers unsigned 16-bit in TCP?

schism/http.c
mrpapersonic: libcurl would be the easiest for me since I've written code targeting it before :p 

The main appeal for not using something like libcurl everywhere was to just reduce the amount of binary bloat (libcurl is a fairly big library). this means SSL is broken if the system's certificates are outdated. but if you're gonna be using something old like that you might as well update those root certs :)

include/http.h
mrpapersonic: this is hardcoded to ONLY do GET requests; the `data` referred to is the data given to the callback, see the callback spec defined above (that comment ought to be moved up there too). There is also nothing allowing changing the headers. Of course all of this is meant to simplify adding new backends.

The guarantee is primarily useful when downloading and parsing text files, which is exactly what we are doing. Perhaps it would make more sense to just pad with one trailing NUL since most of the internet is UTF-8 now.

> Also, the server behind an HTTP site is not identified by its domain but by its host entry

I should have named it `host` but I was concerned more with getting everything working ;)

sys/win32/http.c
logiclrd: Yep, that's right :-)

sys/sdl12/mt.c
logiclrd: I mean, if this is the only transient thread, and only one check is done at startup, then the potential scope of the leak is one single thread handle. I don't think that would break anything :-) Just a little icky.

logiclrd: For what it's worth, the detaching was actually _more correct_ -- it just didn't do anything on SDL 1.2 targets.

What about having a dedicated transient thread clean-up thread (for back-ends that can't detach only), and implementing `Detach` by throwing the current thread into a set of handles that it is monitoring? It wouldn't be deterministic, because who knows what order the threads will actually terminate in and the act of waiting on any one of them will block the clean-up thread for an unknown amount of time, but as long as all the threads that get into that set _are in fact transient_, then _technically speaking_ there's no leak any more :-)

And if there is only one transient thread anyway then the pattern is actually the perfect expression of cleaning up the resource as soon as it is done.

PR #696, opened 2025-07-07
key handling: avoid latching modifier keys themselves when tracking last key

logiclrd:
This PR resolves an issue I ran into with the repeated keypress detection used in a few screens.

As an example, pressing Alt-N in the Pattern Editor toggles the Multichannel state of the current channel. Immediately pressing Alt-N again is supposed to pop up the Multichannel Dialog that allows rapidly updating the Multichannel state of all channels. But, in HEAD, it doesn't work this way. The reason is:

https://github.com/schismtracker/schismtracker/blob/75b87f980ceb310359025a5df37a53e01f363669/schism/main.c#L506-L507

An intermediate variable `last_key` is used, resulting in `status.last_keysym` being delayed and representing the press that occurred _two_ keypresses ago.

Because of this, to get the Multichannel Dialog, you need to press Alt-N _three_ times. Unless you release Alt and then press it again, in which case the second press of Alt fills that intermediate slot. This is inconsistent and undesirable, in my opinion. :-)

The issue is resolved here by updating the code in `main.c` that latches the last key event to only latch non-modifier keys. Then, the last key latched will always be e.g. the N in Alt-N.

There's another issue with the existing code. It doesn't have any provision for checking whether that last keysym was actually part of a modifier combo. For instance, if you go to the Pattern Editor and then you press N followed by Alt-N, the Alt-N handler will see that previous _unmodified_ N and think that you have pressed Alt-N twice.

This PR addresses this by also latching the value of `kk.mod`, which it places into new field `last_keymod` added alongside `last_keysym` in `struct tracker_status`. This makes it possible to tell whether that preceding N was in fact an Alt-N.

This test requires inspecting both `last_keysym` and `last_keymod` simultaneously, which is a bit of an unwieldy expression, so a macro `LAST_KEY_IS` is added that does the right check given `SCHISM_KEYSYM_` and `SCHISM_KEYMOD_` values.

All sites that were doing a naive check on `last_keysym` have been updated to use this new macro.

With these changes:

* Alt-N twice in the Pattern Editor reliably always brings up the Multichannel Dialog, regardless of whether Alt was released in the middle.
* N, Alt-N does not bring up the Multichannel Dialog.

These semantics apply to all double keypress detections.

### DIFF
--- a/schism/main.c
+++ b/schism/main.c
@@ -374,23 +374,6 @@ static void key_event_reset(struct key_event *kk, int start_x, int start_y)
 
 /* -------------------------------------------- */
 
-static int is_modifier_key(schism_keysym_t sym)
-{
-	switch (sym) {
-	case SCHISM_KEYSYM_LCTRL:
-	case SCHISM_KEYSYM_LSHIFT:
-	case SCHISM_KEYSYM_LALT:
-	case SCHISM_KEYSYM_LGUI:
-	case SCHISM_KEYSYM_RCTRL:
-	case SCHISM_KEYSYM_RSHIFT:
-	case SCHISM_KEYSYM_RALT:
-	case SCHISM_KEYSYM_RGUI:
-		return 1;
-	default:
-		return 0;
-	}
-}
-
 SCHISM_NORETURN static void event_loop(void)
 {
 	unsigned int lx = 0, ly = 0; /* last x and y position (character) */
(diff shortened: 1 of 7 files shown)


mrpapersonic:
looks good to me. a little sidenote though: Schism currently handles modifier keys incorrectly, because it only checks whether a specific set of keybinds are down, and not that they are the ONLY ones down. I wrote an algorithm to handle this properly somewhere in the `keybinds` branch.

schism/main.c
```diff
@@ -374,6 +374,23 @@ static void key_event_reset(struct key_event *kk, int start_x, int start_y)
 
 /* -------------------------------------------- */
 
+static int is_modifier_key(schism_keysym_t sym)
```
mrpapersonic: IMO this would be better moved into keyboard.c as `kbd_key_is_modifier`

logiclrd: That's a good idea :-)

logiclrd:
Rebased.

PR #702, opened 2025-07-16
it: accept midnight history timestamps

logiclrd:
The loader sanity checks the date/time stamp values it reads by ensuring that they aren't zeroes, but `fat_time` is legitimately 0 at midnight, which will cause the corresponding history entries to be erroneously discarded.


vcaputo (approved):
Thanks for all the great contributions Jonathan!

PR #708, opened 2025-07-17
liq: unflip title and artist

logiclrd:
An oddity in liq.c: The marshalling of the title and artist fields into `dmoz_file_t` swaps their places. I have it on good authority that this isn't intentional. :-P


mrpapersonic:
The format docs for liquid tracker do say the title and artist are in the places we read them from. (I'm not sure why we even read the artist name, since we don't actually USE it anywhere, but I suppose at some point we could display it in the file browsers.)

I'm 99% sure the flip-flop happened when I refactored everything to use `slurp` instead of hardcoding for memory. there were a lot of mistakes made in that transition iirc that probably would've been fixed earlier if i had made it a pull request and had someone else review the changes, but oh well. :)

PR #714, opened 2025-07-26
slurp: fix SF2 block boundary handling

logiclrd:
I was looking at the slurp code for SF2's disjoint sample channel handling and saw something that didn't make sense:
```
int64_t left = sf2_slurp_tell(s) + count;

if (left < 0)
    return 0; /* ??? */

if ((size_t)left < count) {
    ...
```

So... we take a nonnegative value and add it to positive value `count`. Then, we make sure the result of adding two nonnegative numbers isn't negative. Finally, we check whether this number we got by adding a number to `count` is less than `count`. :-P

I wrote up a simple ad hoc test case and sure enough, it's completely nonfunctional code.

This PR fixes the code in three ways:

1. The calculation of `left` is fixed. From context, it's clearly meant to capture how many bytes are left to be read in the `current` data region.

2. If the caller requests more data than is captured in the data regions, the existing code doesn't clamp the read down to the available range and will read bytes that aren't part of the logical SF2 stream.

3. I noticed that most parts of the SF2 slurp implementation are generalized. In practice, presently, there are only ever two data regions (because they're left and right channels :slightly_smiling_face:), but the code is written to be agnostic of that. But, this code only checks `left` against `count` once, so if a read was large enough to span more than 2 blocks, this would read past the end of the second of those blocks without seeking properly to the third, and the wrong data would be returned. So, I changed the `if` to a `while`. If this code is ever adapted to support more than 2 data regions, it should do the right thing.

This code needs unit tests :-) Did I see that you were creating some sort of automated testing infrastructure recently?

### DIFF
--- a/schism/slurp.c
+++ b/schism/slurp.c
@@ -473,23 +473,29 @@ static size_t sf2_slurp_read(slurp_t *s, void *data, size_t count)
 {
 	size_t read = 0;
 
-	if (s->internal.sf2.current < (ARRAY_SIZE(s->internal.sf2.data) - 1)) {
-		int64_t left = sf2_slurp_tell(s) + count;
+	while (s->internal.sf2.current < (ARRAY_SIZE(s->internal.sf2.data) - 1)) {
+		int64_t off_current = slurp_tell(s->internal.sf2.src) - s->internal.sf2.data[s->internal.sf2.current].off;
+		int64_t left = s->internal.sf2.data[s->internal.sf2.current].len - off_current;
+
 		if (left < 0)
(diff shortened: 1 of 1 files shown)


logiclrd:
If #715 is accepted, then I could add automated testing of the SF2 slurp back-end :-)

mrpapersonic:
Thanks! I think this was overlooked because, realistically, it's not something that will come up ever in the current source (because the split stereo importer only ever reads one audio sample at a time, so the code path for the boundary-reading is never taken in the first place)

PR #715, opened 2025-07-26
Added automating testing infrastructure

logiclrd:
I saw the existing PR #639 `Add a simple "test suite" of sorts` and noted that it has some drawbacks. Most notably, tests are all run in-process, which means if a test fails because the code crashed, then the entire test run aborts.

I decided to have half a go at making a bit more solid test suite.

This PR adds an automated testing infrastructure that operates off of a different build target. To build for testing, you run the `testbuild` target:
```
make clean testbuild
```
In this code, `main` is redirected to a test harness.

Tests are defined in methods with a specific signature and collected up in a file `index.c`. To add a new test, you add the test method, and then you add it to the list in `index.c`. Tests use a set of macros to perform assertions and return results, and can also return ad hoc output which is gathered and displayed independently at the end of the test.

```
testresult_t foobs()
{
  test_printf("Hello from test %s!", "foobs");
  ASSERT((1 == 1) == 1);
  RETURN_SKIP;
}
```

To run tests, you simply invoke `schismtracker` produced by `make testbuild`. You can also run a specific test by passing it on the command-line:
```
./schismtracker test_bshift_arithmetic
```

(This mechanism is used internally for running the full suite; a separate child process is spawned for each individual test.)

There is certainly some polishing that can be done to this:

* Display captured test output even when crashing?? Could possibly handle this by writing it to a file and having the harness check the file...
* Colour coding.
* Needs to be tested on architectures other than Linux and Win32.
* Other stuff I'm sure.

I also am not super familiar with autoconf, so I'm not sure if I did all the right things there. Empirically, though, it works on my end. :-)
```
(base) logiclrd@visor:/code/schismtracker$ ./schismtracker 
TEST: test_bshift_arithmetic ............................... PASS (0 ms)
TEST: test_bshift_right_shift_negative ..................... PASS (0 ms)
TEST: test_bshift_left_shift_overflow ...................... PASS (0 ms)
Results: 3 passed, 0 failed
(base) logiclrd@visor:/code/schismtracker$ 
```


logiclrd:
Incidentally, the PR defines a new platform abstraction `os_shell`, and I've verified that the implementations for Linux and Win32 work. If the platform doesn't provide the abstraction, then it falls back to running the tests in-process, so it's not a zero at least. :-P

automated-testing/testresult.c
```diff
@@ -0,0 +1,36 @@
...
+
+#include "automated-testing.h"
+
+const char *testresult_str(testresult_t result)
+{
+	switch (result) {
+	case SCHISM_TESTRESULT_NOT_RUN: return "NOT RUN";
+	case SCHISM_TESTRESULT_PASS: return "PASS";
+	case SCHISM_TESTRESULT_FAIL: return "FAIL";
+	case SCHISM_TESTRESULT_INCONCLUSIVE: return "INC.";
```
mrpapersonic: Can this say "INCLUSIVE" instead of incorporated?

include/osdefs.h
```diff
@@ -84,6 +84,12 @@ A return value of 0 indicates that the event should NOT be processed by the main
 # define os_fopen xbox_fopen
 #endif
 
+#if defined(SCHISM_WIN32)
+# define os_shell win32_shell
```
mrpapersonic: Isn't this virtually equivalent to
```c
char *cwd = dmoz_get_current_directory();
os_run_hook(cwd, name, arg);
free(cwd);
```
? It would be nice if it could just be reduced to that for all platforms :)

Makefile.am
```diff
@@ -506,6 +508,11 @@ endif
 
 ## aaaaaaaaahhhhhhhhhhhhhhhhhhh!!!!!!!1
 schismtracker_SOURCES = \
+	automated-testing/harness.c        \
```
mrpapersonic: hm, this shouldn't be apart of the normal build. It should have an `--enable-tests` that explicitly enables it.

automated-testing/testresult.c
logiclrd: It can :-) I'll do that in a bit. The only thing is that it means that a column of text containing these values is suddenly potentially much longer.

include/osdefs.h
logiclrd: It isn't, because that doesn't wait on the child process and return its exit code.

Makefile.am
logiclrd: I can see arguments both ways. It was intentional that it was included, because that means that it _must build_, even when it isn't in use. If this ends up being part of PR hooks, then that isn't an issue because test builds will be done all the time anyway :-P

logiclrd: Maybe there's a best-of-both-worlds option. What if it always built the code, but it left the objects out of the linking phase for non-test builds? Come to think of it, in full-optimizing mode, wouldn't it _effectively_ do that anyway? If you define a symbol, but nothing references it, wouldn't it get stripped out during linking??

logiclrd: I just gave it a quick test, and it seems it does not omit unused objects automatically, but the build could probably be configured so that it builds everything but then just doesn't link in the testing-related objects in the default target.

logiclrd: This seemingly-simple idea appears to be rather tricky to fit into the paradigm of autoconf. :-P

logiclrd: Okay this looks like it ought to do it.

I've defined a new target in Makefile.am called `schismtrackertest` (on Windows, `schismtrackertest.exe`). A default build builds both `schismtracker` and `schismtrackertest`. The `schismtrackertest` source files contain all of the test harness & test case files that we don't want built into `schismtracker`. They do _not_ contain the rest of the code; instead, the object files built by the `schismtracker` build are pulled into the link of `schismtrackertest` directly using `schismtrackertest_LDADD`.

With this change, all files are built only once, except for one file, `automated-testing/entrypoint.c`, which defines a thunk to call either `schism_main` or `schism_test_main`, depending on the build type. When `schismtrackertest_LDADD` is defined to include `schismtracker_OBJECTS`, this particular object file is explicitly filtered out (otherwise there'd be two definitions of the thunk when linking).

There's only one thing I haven't figured out that I am aware of: As far as I can tell, `make install` is going to install `schismtrackertest` in the `bin` directory alongside `schismtracker`. I spend a good amount of time trying to figure out how to exclude it from `install` while still taking advantage of the build pipeline, but I didn't find a way. That's a problem for another day. :-)

```
$ autoreconf && ./configure
...
$ make
...
$ ./schismtrackertest
TEST: test_bshift_arithmetic ..................................... PASS (0 ms)
TEST: test_bshift_right_shift_negative ........................... PASS (0 ms)
TEST: test_bshift_left_shift_overflow ............................ PASS (0 ms)
Results: 3 passed, 0 failed
$ ./schismtracker
𝄞 yey music ♯ ♫
```

With these changes, every `make` builds everything. If a change causes a test to stop compiling, you find out about it right away. But the testing infrastructure and test cases do not clutter up the binary.

logiclrd:
GitHub Workflows will need to be updated to include `schismtrackertest$(EXEEXT)` in their artifact packages.

Some workflows can also be updated to execute the tests and fail if the tests fail. :-)

Makefile.am
```diff
@@ -672,3 +675,24 @@ schismtracker_LDADD = $(LIBM) $(libs_sdl3) \
...
+	automated-testing/index.c          \
+	automated-testing/cases/bshift.c
+
+# Enable test paths -- overrides the entrypoint
+schismtrackertest_CFLAGS = $(schismtracker_CFLAGS) -DSCHISM_TEST_BUILD
+
+schismtrackertest_CPPFLAGS = $(schismtracker_CPPFLAGS)
+schismtrackertest_OBJCFLAGS = $(schimtracker_OBJCFLAGS)
+schismtrackertest_DEPENDENCIES = $(schismtracker_DEPENDENCIES)
+schismtrackertest_LDADD = $(filter-out automated-testing/schismtracker-entrypoint.o,$(schismtracker_OBJECTS)) $(schismtracker_LDADD)
```
mrpapersonic: Isn't this a GNU make extension? (though i see we're also using it above...??????!!!)

anyway, we shouldn't really be using GNU make extensions since the point of automake is to be able to create portable makefiles that work everywhere (tho most everyone's just using gnu tools anyway now, so it isn't really that big of a deal.)

include/osdefs.h
mrpapersonic: Maybe we could separate out the common gory bits of `os_run_hook` and `os_shell` and put them into a separate function, maybe `os_exec`, that does all the dirty work, and then the former two functions can just be implemented as stubs that call `os_exec`. (this hypothetical function should probably also store the program return value in a passed pointer, because I think on some OSes it isn't limited to just 0..255 (?) like it is on linux)

automated-testing/testresult.c
mrpapersonic: as long as it can fit in 120 columns or so, it's fine.

Makefile.am
logiclrd: It looks like that's the case. What's the alternative?

mrpapersonic: no idea, I think i'll just put it off until an issue actually pops up with it :^)

mrpapersonic:
hm, if anything the creation of `schismtrackertest` shouldn't happen at all if that path was not explicitly requested (i.e. through `make schismtrackertest`)

From what I've seen, other projects "solve" this by putting all the main program logic within a static library, and the real `main` function just calls into the routines in that static library, and the tests can link with the program logic. This is probably the most clean solution (but I don't know how to do it without introducing libtool complexity)

On the other hand:
```make
schismtrackertest_SOURCES = \
    $(schismtracker_SOURCES) \
    # extra test files

schismtrackertest_LDADD = $(schismtracker_LDADD)
schismtrackertest_CPPFLAGS = -DSCHISM_TEST $(schismtracker_CPPFLAGS)
# ...
```
i THINK this might work properly? It would probably require rebuilding everything, but that should likely be happening anyway if there are any new headers or whatnot.

Makefile.am
logiclrd: Okeydoke :-)

include/osdefs.h
logiclrd: Huh, I didn't realize that POSIX limited it to 8 bits. It's `int` in all the APIs. But sure enough, Linux only gives you 8 bits. The POSIX API is limited thusly, because it packs a bunch of stuff into the `status` parameter to `waitpid` and only gives 8 bits to the exit code. I just checked FreeBSD, though, and if you use its `wait6` API function, which is considerably more capable than `waitpid`, you can pass in a structure that receives the full 32-bit exit code. Windows as well uses `DWORD` values for process exit codes (thread exit codes too). OS/2 takes a `ULONG` to `DosExit` which is in turn returned via `DosWaitChild`. Linux does have an API call `waitid` that _could_ return the full value, but I tested it and it does not. So, it seems Linux is the wimpy one here 😁 

(An AI summary told me that POSIX's 2016 update actually specifies that `waitid` _not_ limit the exit status to 8 bits, but that Linux is dragging its heels on actually implementing it.)

The code I wrote here actually already breaks on Linux because I just assumed I had an `int` to work with and encode some test statuses to values larger than 255. 😮 

As for making a generalized abstraction, it should probably just return an `int`. That might not be 32-bit on all platforms, but it should be _more than enough_ on any platform. 🙂 If we use `waitid` on Linux, then it might in the future magically start returning the full 32 bits.

logiclrd: Looking at the implementations, it looks like `run_hook` is a straight superset of `shell`, and could be implemented in terms of calling it instead of doing its own machinations.

logiclrd: Incidentally, `MAX_PATH` is not the maximum length of the path on Windows. It is the maximum length of a _path component_. This is _widely widely_ misunderstood, but in the Win32 model, a path is a null-terminated `TSTR[32767]`.

logiclrd:
It has been my observation on many other projects that if the tests aren't a fundamental part of every build, they _will_ get neglected and fall out of date. The worst thing is a test suite where half the tests are broken and nobody wants to do anything with it. I would strongly advise having it build the test suite on every build, unconditionally. :-P

include/osdefs.h
logiclrd: Okay, I've reworked the Win32 implementation. I made the following changes:

* It just straight-up always uses the wide character API. In this day and age, I don't see a compelling reason to support the ANSI character API, and in extreme cases, an ANSI build may encounter files and paths it can't access because the characters can't be mapped into the current code page. Unicode will just work.
* This means that the nasty undebuggable macros for the variants aren't needed. Instead, there's a pithy little function `win32_create_process_wait`, and `win32_shell` and `win32_run_hook` both use it to actually invoke the process.
* The process is created using `CreateProcessW`, which supports passing in the initial CWD directly. No more `chdir`-and-back nonsense. :-)

Now, on the other hand, the Linux implementations are _not_ so easily-collapsed.

* `posix_shell` uses the function `posix_spawn`, which simplifies the process of running a child process. Under the hood, I'm sure it's implemented as `fork`/`exec`, but it's much simpler to use at the top end.
* `posix_run_hook` can't use this function, though, because POSIX _doesn't_ provide anything like `CreateProcessW`'s `lpCurrentDirectory` parameter. Instead, the paradigm is to `chdir`, which makes sense, because the paradigm is also `fork`/`exec`, so you're your own process anyway, with your own CWD.

So, `posix_shell` is tight and compact because it leverages `posix_shell`, and `posix_run_hook` _can't_ leverage `posix_shell`. I think they just have to stand as they are.

logiclrd:
> On the other hand:
> 
> ```makefile
> schismtrackertest_SOURCES = \
>     $(schismtracker_SOURCES) \
>     # extra test files
> 
> schismtrackertest_LDADD = $(schismtracker_LDADD)
> schismtrackertest_CPPFLAGS = -DSCHISM_TEST $(schismtracker_CPPFLAGS)
> # ...
> ```
> 
> i THINK this might work properly? It would probably require rebuilding everything, but that should likely be happening anyway if there are any new headers or whatnot.

It does work, but it results in everything being recompiled. The objects can't be shared, because Autoconf gives them names that incorporate the final target. For instance, `schism/main.c` builds to `schism/schismtracker-main.o` for `schismtracker`, but with the above model, it builds to `schism/schismtrackertest-main.o` for `schismtrackertest`. I quickly tired of this because, at least for me, here, now, MINGW64 builds seem take about 10 times as long as Linux builds. It's really tedious having it repeat the exact same compilation when every single compiler invocation runs for 2-3 seconds. 🙁 

There's also a "test purity" argument to be made that recompiling to a different object file for the test build means that the test _isn't actually testing the project's output_.

include/osdefs.h
logiclrd: > It isn't, because that doesn't wait on the child process and return its exit code.

When I wrote this, incidentally, I had the strange idea in my head that `os_run_hook` was fire-and-forget. I don't know where that came from.

`os_run_hook` is pretty much a superset of `os_shell`, but the preceding comment explains why the POSIX versions are independent. On Windows, while you could technically write one in terms of the other, it is simpler, with fewer dynamic buffer allocations (I think), to use a common implementation function per 884b306.

logiclrd: Hmm, also, this:

https://github.com/schismtracker/schismtracker/blob/6306e7a337077ae428d5b7188834ee5da5338d00/sys/posix/osdefs.c#L76

...isn't threadsafe. That is likely completely irrelevant with current uses; they won't ever overlap. But as a generalized function, this is going to have issues if two threads run it concurrently, since `wait` will return on _any_ child process exiting.

logiclrd: I just committed a change that should address this. However, it is untested, because I'm not at my Linux workstation right now. It's a pretty simple change, I expect it to work, but I'll give it a test once I'm on `visor` again. 26d7f03

mrpapersonic: > It just straight-up always uses the wide character API. In this day and age, I don't see a compelling reason to support the ANSI character API, and in extreme cases, an ANSI build may encounter files and paths it can't access because the characters can't be mapped into the current code page. Unicode will just work.

The Unicode APIs don't work on Win9x, but I'm 99% sure I'm the only one who even cares about that (or even routinely tests it). So I'll be the one to fix up Win9x when it breaks.

For the `posix_spawn` stuff, I mostly suggested that it be something like

```c
/* launches a subprocess and waits for it */
int posix_exec(/* possibly NULL */ const char *dir, const char *exe, /* possibly NULL */ const char *maybe_arg, int *status)
{
    int st;

    switch (fork()) {
    /* mostly just cruft from posix_run_hook with additional checks for NULL */
    }

    while (wait(&st) == -1);

    if (WIFEXITED(st) && status)
        *status = WEXITSTATUS(st);

    return 1;
}
```

where then `posix_run_hook` can be
```c
int posix_run_hook(const char *dir, const char *exe, const char *maybe_arg)
{
    int st;
    char *s;

    if (asprintf(&s, "./%s", exe) < 0)
        return 0;

    if (!posix_exec(dir, s, maybe_arg, &st)) {
        free(s);
        return 0;
    }

    free(s);

    return (st == 0);
}
```

and posix_shell:
```c
int posix_shell(const char *exe, const char *arg)
{
    int st;

    if (!posix_exec(NULL, exe, arg, &st))
        return -1; /* ? */

    return st;
}
```

unless im just totally missing something. Then we could (hopefully) remove the platform-specific `os_run_hook` entirely and just base it all off of `os_exec`.

logiclrd: Mm, okay. I'll shuffle stuff around. :-)

logiclrd: Oh, but. The Win32 implementation has all the batch file-specific logic in it. So, `os_exec` will have to be able to handle that contextually...

mrpapersonic: It can just do what the `posix_run_hook` implementation i put above does, just putting `.bat` on the end instead of a `./` at the front :)

logiclrd: Right, but if we're going to have a consistent `platform_exec` interface, then I'm thinking we don't actually need `platform_run_hook` and `platform_shell` at all. They can be generically implemented. Commit coming up in a bit.

logiclrd:
I'm still not at `visor`, so the Linux build isn't end-to-end tested, but I did copy/paste the `posix_exec` implementation into a Linux shell and it seemed to work.

logiclrd:
Broken OS X build?? The log seems to show a `./configure --verbose` or something with no actual `make` afterward...

logiclrd:
Flarbs :-) Just got back to the Linux workstation. It builds, sure, but it does not run properly :-D Fixing...

UPDATE: That was an easy fix. :-)

logiclrd:
noice

image

mrpapersonic:
Merged, but I combined some of the extra commits I made trying to get the win32 build to work properly, and github didn't detect it properly.

PR #723, opened 2025-07-29
page: removed unused field widget_numentry.handle_unknown_key

logiclrd:
Doesn't appear to actually be used at all.

### DIFF
--- a/include/page.h
+++ b/include/page.h
@@ -130,7 +130,6 @@ struct widget_numentry {
 	int32_t max;
 	int32_t value;
 	int *cursor_pos; /* XXX why is this a pointer */
-	int (*handle_unknown_key)(struct key_event *k);
 	int reverse; /* boolean */
 };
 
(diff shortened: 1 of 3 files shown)


mrpapersonic:
It looks like this was used historically for handling a key that wasn't handled by a key event, and was ripped out when I went through and edited everything to use text input instead. https://github.com/schismtracker/schismtracker/blob/8e54c90748ac946f2bee1c2f18c014552e623e53/schism/widget.c#L376

So this probably shouldn't be deleted and has a legit reason to be there. I'm just gonna add back the case for where the text wasn't handled, but really this is just a hack :(

mrpapersonic:
> I'm just gonna add back the case for where the text wasn't handled

done in d2c17fcfc17160aafb1a65ff600d75ab010c3d88

logiclrd:
Cool cool :-)

PR #725, opened 2025-07-29
Waveform visualization

logiclrd:
This PR adds waveform visualization to the Info page.

In order to do this, the mixer code had to be updated to gather recent sample data. This needed to be as performant as possible, as I'm sure this is the hottest hot path in the codebase :slightly_smiling_face: I did my best in that regard. The buffers are statically-allocated, and the calculation and assignment of the values fits into the same spot as the VU calculation.

Also, in recognition of the possibility that some of the target architectures might yet have difficulty with the added CPU load (maybe? I don't know :slightly_smiling_face:), the feature can be turned off with `./configure --disable-waveformvis`. It is enabled by default as presently written.

I also tried making it join the waveform lines, but I actually like it more unjoined. I left that experiment in but `#ifdef`ed out, you can find it inside `page_info.c` under `#ifdef WAVEFORMVIS_JOINED`.

image

### DIFF
--- a/include/player/sndfile.h
+++ b/include/player/sndfile.h
@@ -757,7 +757,7 @@ typedef struct song {
 	uint32_t flags;                                 // Song flags SONG_XXXX
 	uint32_t pan_separation;
 	uint32_t num_voices; // how many are currently playing. (POTENTIALLY larger than global max_voices)
-	uint32_t mix_stat; // number of channels being mixed (not really used)
+	uint32_t mix_stat; // number of channels being mixed
 	uint32_t buffer_count; // number of samples to mix per tick
 	uint32_t tick_count;
 	uint32_t frame_delay;
--- a/include/player/sndfile.h
+++ b/include/player/sndfile.h
@@ -549,6 +549,9 @@ typedef struct song_voice {
 	int32_t rofs, lofs; // ?
 	int32_t ramp_length;
 	uint32_t vu_meter; // moved this up -paper
+#ifdef ENABLE_WAVEFORMVIS
+	int oldest_recent_sample;
+#endif
 	// Information not used in the mixer
 	int32_t right_volume_new, left_volume_new; // ?
 	int32_t final_volume; // range 0-16384 (?), accounting for sample+channel+global+etc. volumes
@@ -862,6 +865,25 @@ typedef struct song {
 	// multi-write stuff -- NULL if no multi-write is in progress, else array of one struct per channel
 	struct multi_write *multi_write;
 } song_t;
+
+#ifndef NATIVE_SCREEN_WIDTH
+#define NATIVE_SCREEN_WIDTH 640
+#endif
+#ifndef NATIVE_SCREEN_HEIGHT
+#define NATIVE_SCREEN_HEIGHT 400
+#endif
+
+#ifdef ENABLE_WAVEFORMVIS
+#define RECENT_SAMPLE_BUFFER_SIZE NATIVE_SCREEN_WIDTH
+
+/* one for each voice, plus one for each output channel */
+extern int8_t recent_sample_buffer[(MAX_VOICES + 2) * RECENT_SAMPLE_BUFFER_SIZE];
+
+#define RECENT_SAMPLE_BUFFER(voice) (&recent_sample_buffer[(voice) * RECENT_SAMPLE_BUFFER_SIZE])
+
+int csf_get_oldest_recent_sample_output(void);
+void csf_set_oldest_recent_sample_output(int new_value);
+#endif
 
 song_note_t *csf_allocate_pattern(uint32_t rows);
 void csf_free_pattern(void *pat);
--- a/include/player/sndfile.h
+++ b/include/player/sndfile.h
@@ -776,7 +776,8 @@ typedef struct song {
 	uint32_t freq_factor; // not used -- for tweaking the song speed LP-style (interesting!)
 	uint32_t tempo_factor; // ditto
 	int32_t repeat_count; // 0 = first playback, etc. (note: set to -1 to stop instead of looping)
-	int8_t *recent_sample_buffer; // one buffer per voice, stored outside of song_voice_t because we regularly obliterate that
+	int8_t *recent_sample_buffer; // one buffer per voice, stored outside of song_voice_t because we regularly obliterate that, plus two for output LR
+	int oldest_recent_output_sample; // pointer into the ringbuffer for the output buffers
 
 	uint8_t row_highlight_major;
 	uint8_t row_highlight_minor;
@@ -882,9 +883,6 @@ typedef struct song {
 #define RECENT_SAMPLE_BUFFER_SIZE NATIVE_SCREEN_WIDTH
 
 #define RECENT_SAMPLE_BUFFER(csf, voice) (&csf->recent_sample_buffer[(voice) * RECENT_SAMPLE_BUFFER_SIZE])
-
-int csf_get_oldest_recent_sample_output(void);
-void csf_set_oldest_recent_sample_output(int new_value);
 #endif
 
 song_note_t *csf_allocate_pattern(uint32_t rows);
--- a/include/player/sndfile.h
+++ b/include/player/sndfile.h
@@ -821,6 +821,10 @@ typedef struct song {
 
 	int32_t opl_to_chan[OPL_CHANNELS];
 	int32_t opl_from_chan[MAX_VOICES];
+
+	// up to MIXBUFFERSIZE stereo samples on each of OPL_CHANNELS channels
+	// allocated on-demand in snd_fm.c
+	int32_t *opl_buffer_data; // up to MIXBUFFERSIZE samples on each of OPL_CHANNELS channels
 	// -----------------------------------------------------------------------
 
 	// MIDI stuff ------------------------------------------------------------
--- a/include/player/sndfile.h
+++ b/include/player/sndfile.h
@@ -776,6 +776,7 @@ typedef struct song {
 	uint32_t freq_factor; // not used -- for tweaking the song speed LP-style (interesting!)
 	uint32_t tempo_factor; // ditto
 	int32_t repeat_count; // 0 = first playback, etc. (note: set to -1 to stop instead of looping)
+	int8_t *recent_sample_buffer; // one buffer per voice, stored outside of song_voice_t because we regularly obliterate that
 
 	uint8_t row_highlight_major;
 	uint8_t row_highlight_minor;
@@ -876,10 +877,7 @@ typedef struct song {
 #ifdef ENABLE_WAVEFORMVIS
 #define RECENT_SAMPLE_BUFFER_SIZE NATIVE_SCREEN_WIDTH
 
-/* one for each voice, plus one for each output channel */
-extern int8_t recent_sample_buffer[(MAX_VOICES + 2) * RECENT_SAMPLE_BUFFER_SIZE];
-
-#define RECENT_SAMPLE_BUFFER(voice) (&recent_sample_buffer[(voice) * RECENT_SAMPLE_BUFFER_SIZE])
+#define RECENT_SAMPLE_BUFFER(csf, voice) (&csf->recent_sample_buffer[(voice) * RECENT_SAMPLE_BUFFER_SIZE])
 
 int csf_get_oldest_recent_sample_output(void);
 void csf_set_oldest_recent_sample_output(int new_value);
--- a/player/csndfile.c
+++ b/player/csndfile.c
@@ -34,6 +34,20 @@
 #include "fmt.h" // for it_decompress8 / it_decompress16
 #include "mem.h"
 
+#ifdef ENABLE_WAVEFORMVIS
+int8_t recent_sample_buffer[(MAX_VOICES + 2) * RECENT_SAMPLE_BUFFER_SIZE];
+static int oldest_recent_output_sample = 0;
+
+int csf_get_oldest_recent_sample_output(void)
+{
+	return oldest_recent_output_sample;
+}
+
+void csf_set_oldest_recent_sample_output(int new_value)
+{
+	oldest_recent_output_sample = new_value;
+}
+#endif
 
 static void _csf_reset(song_t *csf)
 {
--- a/player/csndfile.c
+++ b/player/csndfile.c
@@ -34,21 +34,6 @@
 #include "fmt.h" // for it_decompress8 / it_decompress16
 #include "mem.h"
 
-#ifdef ENABLE_WAVEFORMVIS
-int8_t recent_sample_buffer[(MAX_VOICES + 2) * RECENT_SAMPLE_BUFFER_SIZE];
-static int oldest_recent_output_sample = 0;
-
-int csf_get_oldest_recent_sample_output(void)
-{
-	return oldest_recent_output_sample;
-}
-
-void csf_set_oldest_recent_sample_output(int new_value)
-{
-	oldest_recent_output_sample = new_value;
-}
-#endif
-
 static void _csf_reset(song_t *csf)
 {
 	unsigned int i;
@@ -138,8 +123,14 @@ static void _csf_reset(song_t *csf)
 song_t *csf_allocate(void)
 {
 	song_t *csf = mem_calloc(1, sizeof(song_t));
+
 	_csf_reset(csf);
-	csf->recent_sample_buffer = mem_calloc(MAX_VOICES + 2, NATIVE_SCREEN_WIDTH);
+
+	csf->opl_buffer_data = mem_calloc(OPL_CHANNELS, MIXBUFFERSIZE * 2 * sizeof(int32_t));
+
+	csf->recent_sample_buffer = mem_calloc(MAX_VOICES + 2, RECENT_SAMPLE_BUFFER_SIZE);
+	memset(csf->recent_sample_buffer, 0, (MAX_VOICES + 2) * RECENT_SAMPLE_BUFFER_SIZE);
+
 	return csf;
 }
 
@@ -513,6 +504,9 @@ static void set_current_pos_0(song_t *csf)
 			v->global_volume = 64;
 		}
 	}
+
+	memset(csf->recent_sample_buffer, 0, (MAX_VOICES + 2) * RECENT_SAMPLE_BUFFER_SIZE);
+
 	csf->current_global_volume = csf->initial_global_volume;
 	csf->current_speed = csf->initial_speed;
 	csf->current_tempo = csf->initial_tempo;
--- a/player/csndfile.c
+++ b/player/csndfile.c
@@ -136,6 +136,7 @@ song_t *csf_allocate(void)
 {
 	song_t *csf = mem_calloc(1, sizeof(song_t));
 	_csf_reset(csf);
+	csf->recent_sample_buffer = mem_calloc(MAX_VOICES + 2, NATIVE_SCREEN_WIDTH);
 	return csf;
 }
 
@@ -212,6 +213,8 @@ void csf_destroy(song_t *csf)
 			csf->instruments[i] = NULL;
 		}
 	}
+
+	free(csf->recent_sample_buffer);
 
 	_csf_reset(csf);
 }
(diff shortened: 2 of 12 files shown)


logiclrd:
`K-ACHAET.MTM` is a really fun one to watch with this visualization :-)

I do notice that there is the occasional tearing. One way to eliminate it would be to lock audio during the rendering of every frame. I wonder if it might be possible to do through having multiple "pages" of `recent_sample` buffers, so that the visualization pulls from a page of data that is a little bit behind where the mixer is rendering... Or, maybe it's okay to have the occasional tearing in a sample rendering that's changing 30+ times a second? :-P

include/player/sndfile.h
```diff
@@ -678,6 +681,25 @@ typedef struct song {
...
+#endif
+#ifndef NATIVE_SCREEN_HEIGHT
+#define NATIVE_SCREEN_HEIGHT 400
+#endif
+
+#ifdef ENABLE_WAVEFORMVIS
+#define RECENT_SAMPLE_BUFFER_SIZE NATIVE_SCREEN_WIDTH
+
+/* one for each voice, plus one for each output channel */
+extern int8_t recent_sample_buffer[(MAX_VOICES + 2) * RECENT_SAMPLE_BUFFER_SIZE];
```
mrpapersonic: The recent output buffer for each voice should be in the `song_voice_t` structure, and the two stereo output channels don't need to be there because they're already rendered out in the audio callback (which is how the oscilloscope visualizer works)

I think a circular buffer similar to events.c would work perfectly here

BTW, don't use globals in the replayer, since it's meant to be reentrable (see disko.c)

BTW BTW, we could use a flag to specify whether the mixer should render out this data, but I don't know if the possible performance increase is worth it

BTW BTW BTW this also has to be implemented for the AdLib stuff

player/mixer.c
```diff
@@ -243,23 +255,35 @@ static inline SCHISM_ALWAYS_INLINE uint32_t avg_u32(uint32_t a, uint32_t b)
...
+# define CONVERT_SAMPLE(s) ((int8_t)(((s) >> 18)) ^ 128);
+# define CAPTURE_RECENT_SAMPLE \
+	if (oldest_recent_sample >= NATIVE_SCREEN_WIDTH) \
+		oldest_recent_sample = 0; \
+	recent[oldest_recent_sample++] = CONVERT_SAMPLE(MERGE_SAMPLE(vol_lx, vol_rx));
+#else
+# define CAPTURE_RECENT_SAMPLE
+#endif
+
+#define SNDMIX_STOREVUMETER_AND_RECENT_SAMPLE \
```
mrpapersonic, line 259: These should be separate

schism/page_info.c
```diff
@@ -729,10 +735,133 @@ static void info_draw_note_dots(int base, int height, int active, int first_chan
...
+		uint8_t *recent_samples;
+		int idx;
+
+		for (int y = y1 | 1; y <= y2; y += 2)
+			vgamem_ovl_drawpixel(&window->overlay, x2, y, WAVEFORM_BOX_OUTLINE_COLOUR);
+		for (int x = x1 | 1; x <= x2; x += 2)
+			vgamem_ovl_drawpixel(&window->overlay, x, y2, WAVEFORM_BOX_OUTLINE_COLOUR);
+
+		if ((chan >= MAX_VOICES) || current_song->voices[chan].current_sample_data) {
+			recent_samples = RECENT_SAMPLE_BUFFER(chan);
```
mrpapersonic: what's going on here? why not use the vgamem sample drawing functions?

mrpapersonic:
we can have a *lot* of channels, maybe we should inherit the "channel number" thing in the top left from FT2?

include/player/sndfile.h
logiclrd: Fun fun fun :-D

So... is it okay to balloon `song_voice_t` by more than half a kilobyte?? I notice at least some parts of it seem to be really carefully laid out for cache line optimization and such. :-P

What if the current approach were used, but the buffer were in the owner of the `song_voice_t` (the `song_t`) rather than being static?

I also think a flag to specify whether the mixer should render out the data would likely be a performance _decrease_, because now it has to check that flag all the time. That's why I went with the approach of, "If you can't handle this, then we'll just rip it out entirely and it'll be like it was never there." :-)

logiclrd: The existing output for the oscilloscope visualizer isn't guaranteed to be 8-bit signed. So, either it will need to be bent to this format, or the rendering code will need to be duplicated...

player/mixer.c
logiclrd: Fair :-) In its earliest iteration, it shared the data. Then I came to realize that, actually, it's not the same data. Those graphs looked really weird :-)

schism/page_info.c
logiclrd: Good point :-)

include/player/sndfile.h
logiclrd: Oh, I just remembered another reason I'm not storing it in `song_voice_t`. Every time you stop playback, it does a `memset(voices, 0, sizeof(voices))`. :-P That's in `_csf_reset`.

Do you think it makes more sense to still store it in `song_voice_t` and just make sure it doesn't get obliterated, or to store in a parallel structure?

logiclrd:
> we can have a _lot_ of channels, maybe we should inherit the "channel number" thing in the top left from FT2?

Fair :-) Would be good to be able to toggle it, though...

include/player/sndfile.h
mrpapersonic: > So... is it okay to balloon song_voice_t by more than half a kilobyte?? I notice at least some parts of it seem to be really carefully laid out for cache line optimization and such. :-P

you can just put it directly out of the existing data, it's fine. That note was probably written over 20 years ago. If you're very hard pressed on keeping it optimized, put it directly before the `// Information not used in the mixer` comment.

mrpapersonic: > The existing output for the oscilloscope visualizer isn't guaranteed to be 8-bit signed. So, either it will need to be bent to this format, or the rendering code will need to be duplicated...

The already-existing visualizers just bend the data to the format they need, and even then you can convert 16-bit signed to 8-bit signed with just a simple `rshift_signed(x, 8)`

mrpapersonic: > Oh, I just remembered another reason I'm not storing it in `song_voice_t`. Every time you stop playback, it does a `memset(voices, 0, sizeof(voices))`. :-P That's in `_csf_reset`.
> 
> Do you think it makes more sense to still store it in `song_voice_t` and just make sure it doesn't get obliterated, or to store in a parallel structure?

shouldn't the visualization data get wiped when the song is stopped anyway?

schism/page_info.c
logiclrd: Hmm, the existing sample drawing functions can't draw to a sub-rectangle within an overlay, and overlays are allocated at character boundaries, not pixel boundaries. So, to use the existing functions, it would either be necessary to support having a separate overlay per channel, whose boundaries usually overlapped on a character, or to rewrite the sample rendering to allow a pixel (x1, y1)-(x2, y2) to be supplied by the caller.

What about an intermediate point, for starters, rehoming the per-pixel code into vgamem.c, so it may not be sharing the function but at least its in a more logical place??

logiclrd: Also, both my code and the existing code assume you want a particular colour for your sample data, and they're not the same. :-)

mrpapersonic: If anything, the code should just be changed to allow for any arbitrary coordinates (which means adding x,y,width,height params)

You can also just take in the color as another param

logiclrd:
- [x] Don't use globals
- [x] AdLib
- [x] Move `CAPTURE_RECENT_SAMPLE` out of `SNDMIX_STOREVUMETER`
- [x] Channel number labels (can be toggled with Alt-N, documented in helptext)
- [x] Better selection semantics regarding channels vs. voices
- [x] Find some way to home the sample drawing in `vgamem` and/or reuse

include/player/sndfile.h
logiclrd: > I think a circular buffer similar to events.c would work perfectly here

It already is a circular buffer, just a small one. :-) Each voice gets `NATIVE_SCREEN_WIDTH` recent samples (which I chose because that is the widest you'd need with a one-to-one-mapping even if the visualization spanned the entire screen), and each voice stores an `oldest_recent_sample`. When new data is poured in, `oldest_recent_sample` is advanced, and when rendering, the buffer interpretation starts at `oldest_recent_sample` and wraps as needed.

logiclrd:
Poking through the FM code, I see that it mixes the FM samples directly into the common mix buffer and _then_ scales the mix buffer for the FM-specific volume adjustment (which is something like 227,400% :-P). Have you tried songs combining FM samples and PCM samples since that optimization? :-D

ETA: Oh, haha, that change was just yesterday :-D

I think I might revert that change, because the previous form was actually pretty conducive to `recent_samples`.


https://github.com/user-attachments/assets/39b097b2-e728-4519-820f-ca9655b0d705

mrpapersonic:
> Poking through the FM code, I see that it mixes the FM samples directly into the common mix buffer and _then_ scales the mix buffer for the FM-specific volume adjustment (which is something like 227,400% :-P). Have you tried songs combining FM samples and PCM samples since that optimization? :-D

Nope, because I don't have any. ;)

Fixed with 6722ae63

mrpapersonic:
> I think I might revert that change, because the previous form was actually pretty conducive to `recent_samples`.

The primary motivation behind it was to remove the reliance on variable-length arrays (which are terrible and suck, and we need code to emulate them for compilers that don't implement them)

We could also use a heap-allocated buffer within the `song_t`, and reallocate it if we really need to.

logiclrd:
Okay, that took some finagling but I think I finally got it figured out :-)

Firstly, the OPL2 and OPL3 chips have a major difference: OPL2 has 9 channels and OPL3 legitimately has 18 channels. The source code couldn't seem to decide whether to process 9 or 18 channels, including some loops saying `for (i = 0; i < 9 /* 18 */; i++)`. I resolved this with a commit that adds a macro `FM_CHANNELS` determined based on whether `--enable-opl2` is supplied. If OPL2 is in use, it's 9, if OPL3, 18, and it's used to size buffers and bound loops.

Secondly, when I tried to build the OPL2 code, I found existing code that didn't compile. Has OPL2 not been exercised in a while? :slightly_smiling_face: It was an easy fix, it was just a function that takes no parameters that needed a `(void)` argument list.

Thirdly, this sequence of commits, which unfortunately were done independently and at the same time as your 6722ae6, fixes the issue of the FM volume scale being applied to sample data. As part of this, I discovered additional confusion about what buffers were associated with what channel (to do with 9 vs. 18 above). With these changes, the right data is captured for the right channels.

Fourthly, this introduces gathering of `recent_sample` data from FM channels.

Fifthly, updates `page_info` to recognize channels playing FM data. It also factors in the remapping of voices via `voice_mix`, which previously meant that an active voice wouldn't see its data because the data was going to a different entry in the `voices` table.

With these changes:


https://github.com/user-attachments/assets/433b0308-3a96-4a31-97f4-88be6c985751

logiclrd:
Okay, I believe I've rebased this properly so that 6722ae6 is reverted as part of ce848140.

So at this point, then, the outstanding thing is to rejigger it so that it doesn't need to do big stack allocations inside `Fmdrv_Mix`, but in a way that works with gathering `recent_sample` data.

mrpapersonic:
i'm not sure if it's completely "safe" to allow the 18 extra channels (because then channels that previously cut out no longer will), but maybe i'm just paranoid

mrpapersonic:
> Okay, I believe I've rebased this properly so that [6722ae6](https://github.com/schismtracker/schismtracker/commit/6722ae6358ec26848cffb1e86ded13da31648ee9) is reverted as part of [ce84814](https://github.com/schismtracker/schismtracker/commit/ce8481407672c591095652f098ab69469f218b73).
> 
> So at this point, then, the outstanding thing is to rejigger it so that it doesn't need to do big stack allocations inside `Fmdrv_Mix`, but in a way that works with gathering `recent_sample` data.

just do something like
```
within _csf_reset:
    csf->fm_buf = NULL;
    csf->fm_buflen = 0;

within Fmdrv_Mix:
    int32_t *buffers[FM_CHANNELS];

    if (count > csf->fm_buflen) {
        /* reallocate if necessary */
        csf->fm_buf = mem_realloc(csf->fm_buf, count * FM_CHANNELS * sizeof(int32_t));
        csf->fm_buflen = count;
    }

    memset(csf->fm_buf, 0, count * FM_CHANNELS * sizeof(int32_t));

    for (i = 0; i < FM_CHANNELS; i++)
        buffers[i] = csf->fm_buf + (count);

    OPLUpdateMulti(..., buffers, ...);

    /* now we have buffers we can mix, and shove into the circular buffer */
```

logiclrd:
> just do something like

Yep, that's pretty close to what I wrote :-) I looked at all the call sites and saw that they bound the `count` argument to `MIXBUFFERSIZE`. In practice, in debug sessions, I saw that it was in fact immediately this maximum value anyway, so the code I wrote just uses that as a fixed allocation size. If a caller asks for more than that of `Fmdrv_Mix`, that generates a runtime assertion. Do you think it's likely that a caller ever _would_ ask for more than `MIXBUFFERSIZE`?

logiclrd:
> i'm not sure if it's completely "safe" to allow the 18 extra channels (because then channels that previously cut out no longer will), but maybe i'm just paranoid

What do you mean by cut out? Can we test this?

ETA: Oh, you mean, someone might have written music that assumes 9-channel OPL2, requests more than 9 notes, and some of them don't play because of the channel limitation, and that's okay because that's what they planned for, and now suddenly extra notes are playing that they didn't mean to have playing??

If that's the case, then that corner case can be handled with `./configure --enable-opl2`, I think :grin:

mrpapersonic:
> > i'm not sure if it's completely "safe" to allow the 18 extra channels (because then channels that previously cut out no longer will), but maybe i'm just paranoid
> 
> What do you mean by cut out? Can we test this?

Since the FM channels cap out at 9 channels, new notes will cut off existing notes if they're not already cut.

> Yep, that's pretty close to what I wrote :-) I looked at all the call sites and saw that they bound the count argument to MIXBUFFERSIZE. In practice, in debug sessions, I saw that it was in fact immediately this maximum value anyway, so the code I wrote just uses that as a fixed allocation size. If a caller asks for more than that of Fmdrv_Mix, that generates a runtime assertion. Do you think it's likely that a caller ever would ask for more than MIXBUFFERSIZE?

definitely not, since MIXBUFFERSIZE is the most we can mix at a time anyway.

logiclrd:
Sorry, I edited my comment immediately because I had a subsequent thought, but you were already replying to the original comment :-P

logiclrd:
Okay that's a bit more polished up now. :-)

* All sample drawing is funnelled through draw_sample_data_ex_[8,16,32].
* Repeated code is generated with variant macros.
* Oscilloscope tested with 8-bit, 16-bit and 32-bit output.

I also noticed that CPU usage goes through the roof on the sample view if you have an extremely large sample, and the reason is that it processes _every sample point on every frame_. Even though the sample's waveform display is only, what, 176 pixels wide, every one of those pixels factors in every sample covered by that pixel. If your sample is really long, that's a lot of data. I modified the code to limit how many samples it actually inspects per pixel. (If limited, then it just skips through the sample more quickly.) The small loss in visual fidelity comes with significantly reduced CPU usage.

ETA: This CPU usage issue isn't new to this rewrite. The old code has exactly the same problem for the same reason. :-)

logiclrd:
Rebased. Was getting tired of it crashing on startup, pulled in the upstream fixes to hooks :-)

mrpapersonic:
sorry, just pushed a commit that'll probably make rebasing a pain :(

logiclrd:
Rebased and better than ever. :-)

logiclrd:
During rebasing, I discovered that FM visualization didn't show up on the main output view, only on the per-voice view. Fixed that as part of massaging the relevant commits for rebase.

logiclrd:
Rebased again now that #749 is in.

schism/vgamem.c
```diff
@@ -768,70 +799,121 @@ void draw_vu_meter(int x, int y, int width, int val, int color, int peak)
...
 
-/* somewhat heavily based on CViewSample::DrawSampleData2 in modplug
- *
- * TODO: vectorized sample drawing. */
-#define DRAW_SAMPLE_DATA_VARIANT(bits, doublebits) \
-	static void _draw_sample_data_##bits(struct vgamem_overlay *r, \
-		int##bits##_t *data, uint32_t length, unsigned int inputchans, unsigned int outputchans) \
+#define DRAW_SAMPLE_DATA_EX_VARIANT(BITS, SCALE, SIGNARG_DEF, SIGN_FLIP_DEF, SIGN_FLIP_DO) \
+	void draw_sample_data_ex_##BITS(struct vgamem_overlay *r, int x1, int y1, int x2, int y2, \
+		int##BITS##_t *data, int offset, int stride, int advance, int length, SIGNARG_DEF \
```
mrpapersonic: can't `advance` be calculated with all the info already provided?

p.s. `offset`, `stride`, and `length` should all be `int32_t` or `uint32_t`

schism/page_info.c
```diff
@@ -729,10 +740,147 @@ static void info_draw_note_dots(int base, int height, int active, int first_chan
...
+		for (int x = x1 | 1; x <= x2; x += 2)
+			vgamem_ovl_drawpixel(&window->overlay, x, y2, WAVEFORM_BOX_OUTLINE_COLOUR);
+
+		if (chan < MAX_VOICES)
+			voice = &current_song->voices[chan];
+
+		if ((chan >= MAX_VOICES)
+			? current_song->mix_stat /* main mixer output -- is anything playing? */
+			: ((voice->current_sample_data && (voice->left_volume || voice->right_volume)) /* PCM */
+		 || (current_song->opl_from_chan[chan] >= 0))) /* FM */ {
```
mrpapersonic: `voice->flags & CHN_ADLIB` ?

player/csndfile.c
```diff
@@ -34,6 +34,20 @@
 #include "fmt.h" // for it_decompress8 / it_decompress16
 #include "mem.h"
 
+#ifdef ENABLE_WAVEFORMVIS
+int8_t recent_sample_buffer[(MAX_VOICES + 2) * RECENT_SAMPLE_BUFFER_SIZE];
```
mrpapersonic: unused externs

player/snd_fm.c
```diff
@@ -257,43 +254,102 @@ void Fmdrv_Mix(song_t *csf, uint32_t count)
...
 			if (opl_v < 0 || opl_v >= MAX_VOICES /* this is a bug */)
 				continue;
 
 			buffers[i] = csf->multi_write[opl_v].buffer;
 		}
+	}
+	else {
+		if (!csf->opl_buffer_data) {
+			SCHISM_RUNTIME_ASSERT(sz <= MIXBUFFERSIZE * 2, "Fmdrv_Mix can only process up to MIXBUFFERSIZE samples at a time, caller requested more");
+			csf->opl_buffer_data = mem_calloc(OPL_CHANNELS, MIXBUFFERSIZE * 2 * sizeof(int32_t));
```
mrpapersonic: since this is constant size, it can be moved into the csf init function

player/mixer.c
mrpapersonic: should be renamed back too

mrpapersonic (changes requested):

schism/vgamem.c
logiclrd: `advance` is calculated just before the call to `draw_sample_data_ex`:
```
		int advance = _muldiv(sample->length * chans, 256, r->width * stride);
```
I don't think that information is available in the current parameter set. But, it's also a forced-inline function, and in the current implementation, it's only calculated in one place.

logiclrd: Oh, also, `draw_data_sample_rect_BITS` also uses `draw_sample_data_ex_BITS` and uses a different `advance` calculation.

schism/page_info.c
logiclrd: Mm, I'm not entirely convinced there isn't a 1-to-1 relationship between `CHN_ADLIB` being set and `opl_from_chan[chan]` being mapped, but maybe it doesn't get cleared or something. Okeydoke :-)

player/csndfile.c
logiclrd: :+1:

PR #728, opened 2025-07-31
*_run_hook: only if exists

logiclrd:
```
Dadu@ALTERNATOR MINGW64 /c/code/schismtracker
$ ./schismtracker.exe
'startup-hook.bat' is not recognized as an internal or external command,
operable program or batch file.
'exit-hook.bat' is not recognized as an internal or external command,
operable program or batch file.

Dadu@ALTERNATOR MINGW64 /c/code/schismtracker
$
```

On Windows, the `.bat` file is not itself directly executable. Instead, you run `CMD.EXE /C filename.bat`. `CMD.EXE` does exist and does run even if `filename.bat` doesn't exist.

### DIFF
--- a/sys/posix/osdefs.c
+++ b/sys/posix/osdefs.c
@@ -111,6 +111,11 @@ int posix_run_hook(const char *dir, const char *name, const char *maybe_arg)
 	int st;
 	char *bat_name;
 
+	if (access(name, X_OK) != 0) {
+		// file does not exist or cannot be executed, go no further
+		return 0;
+	}
+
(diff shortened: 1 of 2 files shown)


logiclrd:
Sorry for the multiple force pushes :-P I have another change in the queue that builds on this one, and I was splitting up that change into separate logical change sets, and I didn't do a good job figuring out which bits needed to go in which order.

logiclrd:
I see you went and did a completely different implementation. My follow-up change sort of depended on this implementation. :slightly_frowning_face: Oh well, it's only a few lines, I can rewrite it.

PR #729, opened 2025-07-31
win32_run_hook: support .cmd files

logiclrd:
Some history that you might already know but I'm infodumping here on the offchance that you don't already understand the context for this change: `.BAT` files were created back in the DOS days. The program that ran them was, of course, `COMMAND.COM`.

On Windows 95, `COMMAND.COM` was still the same beast: a 16-bit DOS executable. It had evolved from _actually_ being a `.COM` file, which is just a raw memory dump that gets loaded into a segment as-is. It now had an `MZ` header and internal sections. But it was still called `COMMAND.COM`, and it ran as a 16-bit DOS process, even under the Windows 9x kernel. (There was a good reason for this: In Windows 9x, the GUI and its Windows kernel is actually just a DOS program, and you can exit back to DOS or indeed not even launch it at startup. If you do this, you are running _just_ DOS and `COMMAND.COM` still needs to be able to run.)

When Microsoft created Windows NT, they suddenly had a much more sophisticated environment, with a Unicode API and a registry for settings and such. Most importantly, though, Windows NT _can't run DOS EXEs_. It only supports `PE` executables. So, they made a new interpreter, `CMD.EXE`. It's pretty backwards-compatible, but it adds sophisticated new features related to variable expansion, local scopes for whether features are enabled, and so on. `CMD` can process `.BAT` files, but its native file extension is `.cmd`.

If, on Win32, we're primarily running hooks on Windows NT variants using CMD.EXE, then we should support hooks named `.cmd`.

Note that a recent commit changed the default for if the `COMSPEC` environment variable isn't set from `command.com` to `cmd`. `command.com` doesn't exist on Windows NT variants, but `cmd.exe` doesn't exist on Windows 9x. So, in the unlikely event that `COMSPEC` isn't set, the old version wouldn't work on modern Windows versions, and the new version won't work on Windows 9x. If we want it to work without `COMSPEC` on both systems, then a dynamic fallback is needed. Probably isn't an issue, though, as one _can_ generally assume that `COMSPEC` is set. :-)

My apologies if you already knew all of this.

### DIFF
--- a/sys/win32/osdefs.c
+++ b/sys/win32/osdefs.c
@@ -1655,11 +1655,15 @@ int win32_exec(int *status, const char *dir, const char *name, ...)
 
 int win32_run_hook(const char *dir, const char *name, const char *maybe_arg)
 {
+	/* in order of preference */
+	static const char *extensions[] = {
+		"bat", /* prioritize .bat for legacy */
+		"cmd"  /* because steel is heavier than feathers */
+	};
+	size_t i;
 	int st;
-	char *bat_name;
+	char *bat_name = NULL; /* yes, this is necessary */
 	char *cmd;
-	char *full;
-	struct stat dummy;
 
 	/* obey COMSPEC */
 	SCHISM_ANSI_UNICODE({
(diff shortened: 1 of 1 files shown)


mrpapersonic:
> Forgive me, for I have allowed my IDE to trim trailing whitespace.

Please do this more, trailing whitespace is annoying

PR #732, opened 2025-08-02
midi-ip: rework win32 implementation

logiclrd:
This PR makes a few changes:

(0: On Windows, sockets aren't technically `int` values. It uses something weird under the hood like `UINT_PTR` or something. Anyway, all WinSock functions use a typedef called `SOCKET`. Of course, this doesn't exist on non-Windows systems. I updated the code to use `SOCKET` instead of `int`, and to `typedef int SOCKET` on non-Windows builds.)

1. A new Win32 WinSock2 implementation using `WSAEventSelect` is added. The principle behind this is that Windows has this kernel object called an "event" that is essentially a condition variable. They can be autoreset, in which case each time they're waited upon, the signalled state is "consumed" automatically, or manual reset, in which case once they're signalled they stay signalled until you explicitly clear the signal. Either way, WinSock2 can link sockets to these events, so that whenever a kind of activity that you'd otherwise detect using `select()` happens, it sets a specified event. This makes it really easy to wake a thread that is waiting: since it's just waiting on an event object, you can just set the event object yourself. So, this is what it does. I've tested it and it works. :-)

2. This can only be done with WinSock2, which isn't available in baseline Windows 95 installations. Since that seems to be one of the target platforms of Schism Tracker presently, I made a fallback implementation. It basically does what the POSIX implementation does. The catch is that on POSIX, sockets are just a kind of fd, and so are pipes, so we can just make a pipe and add it to the list of fds we're `select`ing on, but on Windows, sockets are their own unique kind of entity. There is no concept of a pipe fitting into the same universe that you could `select` upon. But, you _can_ just use (drumroll) sockets! :-) I wrote some code to create a loopback connection between two sockets, and that fits into the same model as the `pipe` fds. Writing has to be done with `send` and reading has to be done with `recv`, but it's basically the same. It detects whether it needs to go this route using a new `./configure` test. If it _can_ use `WSAEventSelect`, it does. If it can't, it tries the loopback socket connection. I've tested it and it works. :-)

4. This same code needs to compile & run on Linux as well. I've tested it and it works. :-)

5. If it can't establish the sockets, it falls back on the existing (Win32) implementation, with the 1 second poll. I .. haven't tested this specifically.

6. I found that with the ipMIDI thread actually running, Schism Tracker would hang on shutdown. I tracked this down to the fact that it sets `midi_provider.cancelled` to 1 and then just waits, assuming that the very act of setting it to `1` will cause the provider to shut down. Since with the previous changes, we now have an effective means of waking up the thread, I added some plumbing to expose this, so that before waiting, it wakes the thread. Only the ipMIDI provider actually provides a `wake` implementation presently, but if there are other providers that have a persistent thread that needs to shut down, they could potentially tie into this as well.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -443,12 +443,16 @@ cygwin*|mingw*)
 	WIN32_LIBS="-lwinmm"
 
 	AC_CHECK_HEADERS([winsock.h winsock2.h])
-	if test "x$ac_cv_header_winsock_h" = "xyes"; then
-		AC_SUBST([NETWORK_LIBS], [-lwsock32])
-		use_network=yes
-	elif test "x$ac_cv_header_winsock2_h" = "xyes"; then
+	AC_CHECK_LIB(ws2_32, WSAEventSelect, [use_network_ws2=yes], [use_network_ws2=no])
+	if test "x$use_network_ws2" = "xyes"; then
 		AC_SUBST([NETWORK_LIBS], [-lws2_32])
 		use_network=yes
+		AC_DEFINE([USE_NETWORK], [1], [Networking])
+		AC_DEFINE([USE_NETWORK_WS2], [1], [Win32 WinSock2 networking])
+	elif test "x$ac_cv_header_winsock_h" = "xyes"; then
+		AC_SUBST([NETWORK_LIBS], [-lwsock32])
+		use_network=yes
+		AC_DEFINE([USE_NETWORK], [1], [Networking])
 	else
 		AC_MSG_ERROR([*** Couldn't find winsock (how)])
 	fi
@@ -627,6 +631,7 @@ AM_CONDITIONAL([USE_MACOS], [test "x$use_macos" = "xyes"])
 AM_CONDITIONAL([USE_OS2], [test "x$use_os2" = "xyes"])
 AM_CONDITIONAL([am__fastdepOBJC], [test "x$use_macosx" = "xyes"])
 AM_CONDITIONAL([USE_NETWORK], [test "x$use_network" = "xyes"])
+AM_CONDITIONAL([USE_NETWORK_WS2], [test "x$use_network_ws2" = "xyes"])
 AM_CONDITIONAL([USE_XBOX], [test "x$use_xbox" = "xyes"])
 AM_CONDITIONAL([USE_ASIO], [test "x$use_asio" = "xyes"])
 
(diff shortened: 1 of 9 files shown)


mrpapersonic:
> I found that with the ipMIDI thread actually running, Schism Tracker would hang on shutdown.

That tracks, I have literally no idea how to even test ip-midi so it's kind of just been sitting there untouched ever since the original port to SDL 2.

> I tracked this down to the fact that it sets midi_provider.cancelled to 1 and then just waits, assuming that the very act of setting it to 1 will cause the provider to shut down. Since with the previous changes, we now have an effective means of waking up the thread, I added some plumbing to expose this, so that before waiting, it wakes the thread. Only the ipMIDI provider actually provides a wake implementation presently, but if there are other providers that have a persistent thread that needs to shut down, they could potentially tie into this as well.

For the other midi implementations (alsa, jack) it just goes the lazy route (check for not cancelled in a while loop, and sleep for 10msec after processing midi data), but that sucks for low-latency input. I haven't yet had anyone complain about it though..

logiclrd:
> For the other midi implementations (alsa, jack) it just goes the lazy route (check for not cancelled in a while loop, and sleep for 10msec after processing midi data), but that sucks for low-latency input. I haven't yet had anyone complain about it though..

This opens up the door to a follow-up change that allows them to expose a wake function and avoid busy polling. :-)

I tested this by means of a nifty program I found called VMPK - Virtual MIDI Piano Keyboard. It runs on all platforms. With it, I was able to actually observe notes appearing in Schism Tracker when I played them on the piano. :smile:

mrpapersonic:
right, I don't think we actually need full threads for all our midi. And really, not using full threads opens up the possibility for threads to be optional ;)

logiclrd:
That's fair :slightly_smiling_face: This particular PR, though, makes the Win32 implementation of ipMIDI work properly. :grin:

mrpapersonic:
some of these changes were implemented in https://github.com/schismtracker/schismtracker/commit/9359fae232ea14049cee5cea4d3d1681e661a6cc

There is no longer any waiting (or new threads for that matter), the highlevel midi.c just calls into midi-ip.c which calls `select` to check for available data every so often (its less maintenance burden on an interface virtually no one uses and fewer implement). In fact this could be better with nonblocking sockets, where it could just `recv` and check for an errno value. I'm not sure if a proper threaded implementation would be faster, but this way works whether `USE_THREADS` is enabled or not (again less maintenance burden).

PR #734, opened 2025-08-04
page_orderpan: unhook key handler from current sample

logiclrd:
My theory is that the key handler was created by copy/pasting the `page_patedit` key handler for a starting point, and the keys for adjusting the current sample were never removed from that initial copy/pasted version. I checked, though, and Impulse Tracker does not change the current sample when pressing Ctrl-Up/Down or '/; on the Order List page.

For what it's worth, this bug goes back to the very first commit of Schism Tracker :-)

### DIFF
--- a/schism/page_orderpan.c
+++ b/schism/page_orderpan.c
@@ -519,14 +519,6 @@ static int orderlist_handle_key_on_list(struct key_event * k)
 			new_order++;
 		break;
 	case SCHISM_KEYSYM_UP:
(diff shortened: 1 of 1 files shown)


logiclrd:
Hmm, I just found code in page.c explicitly synchronizing the instrument list page to the currently-selected sample if the previous page was one of the order list pages. So... not a copy/paste error after all??

Is it because the order list page includes functionality to convert a pattern -- by selecting it in the order list -- to a sample, and thus it lets you pick which sample slot you're going to drop it into??

PR #735, opened 2025-08-05
page_log: timestamp the start of each new log section

logiclrd:
I was writing some music, and found that I couldn't tell whether I had saved recently by looking at the log. It said it was saved, but _when_?


logiclrd:
Force-push, force-push...

image

mrpapersonic:
Ended up implementing this myself cause I liked the idea, but the implementation had stuff like `gmtime_r` being introduced even though the result of that call wasn't even being used (plus it was a lot of code for what is a relatively non-intrusive change). I just decided to leverage `printf` because it can do right padding already :)

also apparently I screwed up the co-authored-by crap, oops :')

logiclrd:
Well, the result of `gmtime_r` _was_ being used, in order to compute the offset to get a local time. But, sounds like there was a better way to get a local time? :-) Anyway, cool beans

mrpapersonic:
I probably just passed over it when looking at the code ;)

Getting a local time is really as simple as `time` + `localtime_r`. (but of course my own buggy code caused localtime to actually return GMT time on windows).

logiclrd:
Ahh :slightly_smiling_face: Yeah, I wrote the code on a Windows machine, using the polyfill code. The docs said `localtime_r` would return the local time but in practice it didn't. I didn't question `localtime_r`, and instead just figured, I guess I have to compute and add an offset :stuck_out_tongue: I achieved _that_ by calling `gmtime_r`. I dunno, looking back on it I don't really understand why it worked, because shouldn't `gmtime_r(EPOCH)` still be `EPOCH`, just broken into fields? But in any case, totally unnecessary as it turns out! :grin:

PR #736, opened 2025-08-06
Add networking capabilities (multi-player tracking)

mrpapersonic:
This (will) work on the same basic protocol that IT uses. The main difference being is that we're not in the late 90s and the internet isn't slow as molasses anymore. So we should be able to take in any hostname and (at least try to) connect to it, unlike IT which only takes MAC addresses.

Maybe we could even create a "peer list" of sorts. I'm thinking something like the Minecraft multi-player menu:
image

where we could list peers, their "username", and maybe even a MOTD. I'm open to ideas :)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -470,6 +470,10 @@ endif
 endif
 
 if USE_NETWORK
+files_network = \
+	schism/midi-ip.c \
+	schism/network-tcp.c \
+	schism/network.c
 cflags_network=
 libs_network=$(NETWORK_LIBS)
 endif
@@ -589,7 +593,6 @@ schismtracker_SOURCES = \
 	schism/mem.c           \
 	schism/menu.c			\
 	schism/midi-core.c		\
-	schism/midi-ip.c		\
 	schism/mplink.c			\
 	schism/page.c			\
 	schism/page_about.c		\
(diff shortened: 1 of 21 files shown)


logiclrd:
Instances over the Internet would likely not be adequately synchronized to sound good combined. But, instances over the Internet would be far enough apart that a desynchronization of a hundred milliseconds or two would be irrelevant :-P

What would the use be of sending partial patterns? Incremental updates during an editing session? If so, then I think instead it would make sense for the master instance to have a structure for each peer that includes a copy of the song _as it currently exists on the peer_, and then to use a delta encoding algorithm to generate efficiently-encoded updates. I suppose an argument could be made that one way to efficiently encode an update is to make repeated calls to `Network_SendPartialPattern`...

The reason that Impulse Tracker only works at the Ethernet level, incidentally, is almost certainly simply that it didn't have a TCP stack to work with. On a LAN, TCP's latency is going to be more than good enough, and the ease of using a stream protocol would likely vastly outweigh any benefit from going a level deeper.

One possibility to consider is multicast -- if you had a bunch of instances on a LAN, they could be kept up-to-date with multicast packets that only have to be sent once to update all of them. Would the idea be that different instances play different channels??

mrpapersonic:
> What would the use be of sending partial patterns?

I have no idea. Likely to reduce bandwidth, because when you do anything to a pattern in the pattern editor you're most definitely not going to be editing the entire pattern (only a channel, or a few channels in shift selection)

Note that this is just reimplementing the original protocol IT used. I didn't really ask any questions because I assumed it would work fine (i.e. if it ain't broke, don't fix it)

I have a basic TCP driver using POSIX sockets thats 99% finished, I just need to hook all of the guts together to get it to actually push/pull TCP packets (so that I could test it), and limiting the receiver to only pull the remaining bytes of the message in progress. (eventual TODO is also IPv6 support)

mrpapersonic:
image

TCP networking seems to be working now, after some hacking up of argv to get it to start the server

PR #738, opened 2025-08-08
config-parser: unit tests

logiclrd:
Add unit tests based on the commented-out test code at the end of config-parser.c.
Add a facility for temporary files associated with tests, with autogenerated names and automatic clean-up at the end of the test run. Files are added to a log in a separate file, and that log is later read in to effect clean-up. In this way, a test that creates a temporary file and then crashes does not "forget" about the file it created.


logiclrd:
Hmm, weird. It appears that `test_config_file_null_default_with_value_set_defined_key` consistently crashes on some target platforms, including OS X with Xcode 15.4 and newer and Windows 64-bit. But, it doesn't crash on Windows 64-bit on my end.

logiclrd:
I don't have the tools to debug this. I've done some poking at running OS X in a VM, but that is naturally next to impossible. I haven't figured out how to make it happen. :-(

logiclrd:
Okay, very weird. I fiddled with it some more, and I think I'm reproducing the crash. But, the system it's on doesn't have debugging tools, so I started in on `printf` debugging -- and as soon as I add the first `printf`, it stops crashing. :-/
<hr />
Test:

	...
	char buf[64];
	
	const char *semaphore = "didn't change";
	
	strcpy(buf, semaphore);
	
	// Act
	cfg_get_string(&cfg, "ducks", "weight", buf, 64, NULL);
	
	// Assert
	ASSERT(strcmp(buf, semaphore) != 0);

Output:
```
Jonathan Gilbert@RESONATOR MINGW64 /b/code/schismtracker
# ./schismtrackertest.exe test_config_file_null_default_with_value_set_defined_key
TEST: test_config_file_null_default_with_value_set_defined_key Segmentation fault

Jonathan Gilbert@RESONATOR MINGW64 /b/code/schismtracker
#
```
<hr />

Test:

	...
	char buf[64];

	const char *semaphore = "didn't change";

	strcpy(buf, semaphore);

	// Act
	printf("\nabout to call cfg_get_string\n"); // NEW LINES
	fflush(stdout);                             //

	cfg_get_string(&cfg, "ducks", "weight", buf, 64, NULL);

	// Assert
	ASSERT(strcmp(buf, semaphore) != 0);

Output:
```
Jonathan Gilbert@RESONATOR MINGW64 /b/code/schismtracker
# ./schismtrackertest.exe test_config_file_null_default_with_value_set_defined_key
TEST: test_config_file_null_default_with_value_set_defined_key 
about to call cfg_get_string
... PASS (6 ms)

Jonathan Gilbert@RESONATOR MINGW64 /b/code/schismtracker
#
```
<hr />

logiclrd:
Ahh, I figured it out. The `cfg_get_string` function writes past the end of the supplied buffer, per the interpretation of `len` being the length of the buffer. Or, to put it another way, `len` needs to be the maximum number of characters that can be stored _without including the null-terminator_. All existing call sites pass one less than the actual buffer size. But, the commented-out test code that I translated into test cases passes 64 for `char buf[64]`. As a result, `cfg_get_string` writes `'\0'` to the byte after the end of the buffer. If that byte matters, then the process crashes. :stuck_out_tongue:  Easy fix.

mrpapersonic:
> Files are added to a log in a separate file, and that log is later read in to effect clean-up. In this way, a test that creates a temporary file and then crashes does not "forget" about the file it created.

I thought maybe we could use the lesser-known `tmpfile()` standard library function can do this for us. It deletes the file automatically on `fclose()` or when the program exits "normally"... though the standard says:

> If the program terminates abnormally, whether an open temporary file is removed is implementation-defined.

Windows doesn't include crashes. Mac OS X immediately unlinks the file after its created (so theoretically once the process ends the file descriptor gets closed and the file gets deleted). Linux says "The file will be automatically deleted when it is closed or the program terminates." which I think also includes crashes.

So Windows seems to be the odd one out here... ugh

mrpapersonic:
> Ahh, I figured it out. The `cfg_get_string` function writes past the end of the supplied buffer, per the interpretation of `len` being the length of the buffer. Or, to put it another way, `len` needs to be the maximum number of characters that can be stored _without including the null-terminator_. All existing call sites pass one less than the actual buffer size. But, the commented-out test code that I translated into test cases passes 64 for `char buf[64]`. As a result, `cfg_get_string` writes `'\0'` to the byte after the end of the buffer. If that byte matters, then the process crashes. :stuck_out_tongue:  Easy fix.

Thats a terrible design. Nothing should EVER take the "maximum number of characters without NUL terminator". (source: a similar bug was also in the variables screen.) This same dangerous design flaw is in the standard `strncpy` as well as Microsoft's `_vsnprintf`, and both have caused too many security bugs to count, because programmers assume that the string that comes out will be safe. in fact `_vsnprintf` is even worse, because `vsnprintf` always puts a NUL terminator in literally every other implementation that matters.

In general we should steer away from making these kinds of API decisions, and slowly convert the APIs which do, to take in the full buffer size. Otherwise we are just asking for hard to trace bugs.

logiclrd:
> In general we should steer away from making these kinds of API decisions, and slowly convert the APIs which do, to take in the full buffer size. Otherwise we are just asking for hard to trace bugs.

I'm on board with this :-) There aren't that many calls to `cfg_get_string`, we could have a follow-up change that just reinterprets it to be the size of the buffer. Places that allocate 65 bytes and pass in 64 just get changed to pass in 65, etc.

logiclrd:
> I thought maybe we could use the lesser-known `tmpfile()` standard library function can do this for us. It deletes the file automatically on `fclose()` or when the program exits "normally"... though the standard says:

> Windows doesn't include crashes. Mac OS X immediately unlinks the file after its created (so theoretically once the process ends the file descriptor gets closed and the file gets deleted). Linux says "The file will be automatically deleted when it is closed or the program terminates." which I think also includes crashes.

The purpose of the file here, in any case, is to act as a low-tech pipe between the child and parent process. As such, `tmpfile` is unsuitable in that the file is deleted before the parent would have a chance to read it, and doubly-so on OS X where the file is unlinked before the child has even written anything to it. :-P In this implementation, these files are created by the child and deleted by the parent, by design.

A more sophisticated design could instead capture the output streams of the child process directly, but while that is possible on both Windows and POSIX-y systems, it's not _the same_ and would require considerable abstraction to be built up.

test/tempfile.c
```diff
@@ -0,0 +1,104 @@
...
+
+	if (template != NULL) {
+		if (template_length < 0)
+			template_length = strlen(template);
+
+		const char *buf = template;
+		remaining = template_length;
+
+		while (1) {
+			int num_written = fwrite(buf, 1, remaining, f);
```
mrpapersonic, line 52: i think `fwrite` can only return less than `remaining` if an error occurs (in which case any further writes will be blocked)

test/harness.c
```diff
@@ -63,10 +64,15 @@ static testresult_t run_test_child(const char *self, test_index_entry *entry)
...
 #ifdef HAVE_OS_EXEC
-	int status;
-
-	if (os_exec(&status, NULL, self, entry->name, (char *)NULL))
-		return status;
+	int status, abnormal_exit;
+
+	if (os_exec(&status, &abnormal_exit, NULL, self, entry->name, (char *)NULL))
+	{
+		if (abnormal_exit)
```
mrpapersonic: IMO this should be use the ternary operator

test/cases/config-parser.c
```diff
@@ -0,0 +1,809 @@
+#include "test.h"
+#include "test-tempfile.h"
+#include "test-assertions.h"
+
+#include "charset.h"
+
+#include "config-parser.h"
+
+#define CONFIG_FILE \
```
mrpapersonic: can this just be a static const global?
```
static const char *sample_config =
    "[ducks]\n"
    <...>
```

test/tempfile.c
```diff
@@ -0,0 +1,104 @@
...
+#include "headers.h"
+#include "util.h"
+
+#include "test-tempfile.h"
+
+#ifdef SCHISM_WIN32
+#define unlink _unlink
+#endif
+
+int test_temp_file(char **temp_file, const char *template, int template_length)
```
mrpapersonic: `temp_file` has a maximum character count anyway (16), so we can just `#define` that value as a constant in `test-tempfile.h` and not have to worry about freeing it in the tests

test/cases/config-parser.c
mrpapersonic, line 1: Like I did with baa66af1eed6af175177a49c2f3b3287f8c0cdad, the "common sequence" can be moved into its own function, maybe `static int test_config_file_init(cfg_file_t *cfg)`, and the tests can just call into it.

test/tempfile.c
logiclrd: According the documentation for `fwrite`, there are plenty of transitive reasons why a write might be partial:

* Insufficient disk space (could change at any time)
* Interrupted syscalls
* Device-specific limitations ("Some devices or file systems might impose limitations that lead to partial writes in specific scenarios")

When a partial write happens, `fwrite` doesn't return an error. I'm pretty sure it is canonically correct to treat a partial write as a success and retry with the remaining data. If the error was not transient, then the next call will fail to write any items. But, the way the interface is defined, the next call _could_ succeed and write more items.

test/harness.c
logiclrd: Can do.

test/cases/config-parser.c
logiclrd: I don't see why not. :-)

test/tempfile.c
logiclrd: Fair enough :-)

test/cases/config-parser.c
logiclrd: Done. I used a macro so that it could return directly from its point of invocation.

test/cases/config-parser.c
```diff
@@ -0,0 +1,579 @@
...
+
+	// Act
+	const char *colour = cfg_get_string(&cfg, "ducks", "colour", NULL, 0, "abcd");
+	int count = cfg_get_number(&cfg, "ducks", "count", 1234);
+	int weight = cfg_get_number(&cfg, "ducks", "weight", 1234);
+
+	// Assert
+	ASSERT(strcmp(colour, "brown") == 0);
+	ASSERT(count == 7);
+	ASSERT(weight == 64);
```
mrpapersonic, line 47: `cfg_free` ? :)

logiclrd: Oh, I forgot that was a thing. :grin:

logiclrd: Hmm, in the absence of a scope-exit structure like `try/finally` or C++ scoped objects, I don't think there's a way to guarantee that it always gets freed properly. But, we can at least have it free it on the happy path. A passing test run thus includes known-working `cfg_free`. Since tests are isolated in their own processes, failed tests are going to immediately have their entire memory space torn down anyway. Acceptable?

logiclrd:
Rebased.

mrpapersonic:
Ok, I pushed a commit changing the `cfg_get_string` behavior to be sane, so you can rebase & drop 0d91ddd6cccb54915ed4e04e51dd45bb6fcdb3c2.

test/cases/config-parser.c
```diff
@@ -0,0 +1,617 @@
...
+	ARRANGE_TEST_CONFIG_FILE(cfg);
+
+	// Act
+	// could be separate tests...
+	const char *str_null_section = cfg_get_string(&cfg, NULL, "shouldn't crash", NULL, 0, "ok");
+	const char *str_null_key = cfg_get_string(&cfg, "shouldn't crash", NULL, NULL, 0, "ok");
+	int num_null_section = cfg_get_number(&cfg, NULL, "shouldn't crash", 1);
+	int num_null_key = cfg_get_number(&cfg, "shouldn't crash", NULL, 1);
+	const char *str_null_default_value = cfg_get_string(&cfg, "doesn't", "exist", NULL, 0, NULL);
+
```
mrpapersonic, line 108: as per b11e3dd9ab191f20830062be74f6035af4db78ba we should have a test that supplies a valid buffer but puts in an (obviously bad) 0 for the length

logiclrd:
Rebased. There was in fact already a test for the exact scenario, just the semantics of it changed with the reinterpretation of `len`. So, I didn't add a test, I just reworked the body of `test_config_file_string_boundary_zero`.

mrpapersonic:
Rebased and merged

logiclrd:
I notice you've independently merged a few PRs with amendments of your own and then marked the PR itself "abandoned" a few times :-) I need to check this off, don't I?

image

mrpapersonic:
no, it's just that github really sucks at "detecting" that I've rebased onto master and then merged it in locally, which is my workflow. I wish I could just mark it as "merged" myself but I can't :(

logiclrd:
Ahh :slightly_smiling_face: If I've just pushed a rebase onto `master`, why does it need to be rebased onto `master` again?? If it's freshly-rebased, wouldn't clicking the "Merge" button in GitHub do the exact same thing as the commands you're running?? I guess doing it through GitHub means you can't build/test it before pushing the merge...

I wonder what would happen if I rebased the PR _after_ your merge -- it'd turn into a no-op, but could the no-op be "merged" through GitHub?

PR #739, opened 2025-08-08
test: show overall elapsed time

logiclrd:
Update test/harness.c to measure & display total elapsed time.
Add function `str_from_num_thousands` to format a number with thousands separators. (Is there something built-in for this that I missed??)
Add unit tests for `str_from_num_thousands`. :slightly_smiling_face:


mrpapersonic:
> Add function `str_from_num_thousands` to format a number with thousands separators. (Is there something built-in for this that I missed??)

To my surprise, no, there's nothing in printf that allows for thousands separation (unless you count POSIX extensions)

mrpapersonic:
Okay, I actually went refactoring this code and uncovered a compiler bug in gcc: https://godbolt.org/z/8d14T1c3s

seems it's been fixed in the latest build though. I'll just force off inlining the function in older gcc, since that seems to get it to play nice

EDIT: clang totally fails at it too, it just fails at the absolute value instead of the places. sigh. maybe im tripping over some weird UB but it doesn't feel like it.

EDIT EDIT: okay managed to fix it by shuffling around stuff in the babs32() impl, though I think its weird that doing bitwise operations made it fail

EDIT EDIT EDIT: it's signed integer overflow, and I hate C

mrpapersonic:
rebased onto master at 891b342c41aee856ea5eebccb7eedfc065cc9895

PR #741, opened 2025-08-09
page_preferences: consolidate save operations

logiclrd:
Currently, when you press `Save all Preferences`, the following happens:

* The configuration file is read & parsed.
* MIDI configuration is transferred into the `cfg_file_t`.
* If a change occurred, then the file is rewritten.
* The configuration file is read & parsed.
* Most common settings (the "atexit" bunch) are transferred into the `cfg_file_t`.
* If a change occurred, then the file is rewritten.
* The configuration file is read & parsed.
* Audio driver/device names are transferred into the `cfg_file_t`.
* If a change occurred, then the file is rewritten.

In the best case, the exact same file is read and parsed three times instead of once.

In the worst case, more than one of the change sets has changes, and the file is written multiple times, and the second time it is written obliterates the backup made by the first time it runs.

This PR adds method `cfg_preferences_save`:

* The configuration file is read & parsed.
* MIDI configuration is transferred into the `cfg_file_t`.
* Most common settings (the "atexit" bunch) are transferred into the `cfg_file_t`.
* Audio driver/device names are transferred into the `cfg_file_t`.
* If a change occurred, then the file is rewritten.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -54,6 +54,7 @@ extern char cfg_export_pattern[SCHISM_PATH_MAX + 1];
 void cfg_init_dir(void);
 void cfg_load(void);
 void cfg_save(void);
+void cfg_preferences_save(void);
 void cfg_midipage_save(void);
 void cfg_atexit_save(void); /* this only saves a handful of settings, not everything */
 void cfg_save_output(void);
--- a/schism/config.c
+++ b/schism/config.c
@@ -410,16 +410,9 @@ void cfg_save_output(void)
 	cfg_free(&cfg);
 }
 
-void cfg_atexit_save(void)
+static void cfg_save_world(cfg_file_t *cfg)
 {
-	char *ptr;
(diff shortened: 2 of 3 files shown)


mrpapersonic:
Maybe in the future we should just load the entire config on startup and then never again (i.e. keep it in memory)? That's what I've done in other projects ..

logiclrd:
The rationale for loading on-demand is if multiple processes could be sharing it -- and while that isn't the common case, it's definitely _possible_ to have multiple Schism Tracker instances running. I've done that on occasion. Without a cross-process mutex, there's still a race condition, but it's pretty minimal with load/update/save cycles triggered by UI events. With a persistent configuration, though, I think there would be unexpected effects if the user made changes in one instance and then closed the other, or what have you.

logiclrd:
That said, it's a far cry from any sort of cooperative sharing of the file as-is :-)

PR #744, opened 2025-08-10
dmoz: remove unused macros

logiclrd:
These two macros date right back to the initial commit in the Git repository 20 years ago. I can't find evidence that they have ever been used.

Also, allowed my IDE to trim trailing whitespace in this file.


mrpapersonic:
these could probably be useful for classic mode, so I'm not gonna remove them just yet

logiclrd:
Fair enough :-)

PR #748, opened 2025-08-11
fm_opl2: fixed CALC_CHANNEL

logiclrd:
When the code was refactored into the macro `CALC_CHANNEL` inside `ym3812_update_multi`, some of the references to the loop iterator `j` as the channel number were overlooked. When the macro is used for rhythm mode, `j` will not have the correct value `0`. Instead, it'll be the value 6 left over from the `for` loop that handles the first 6 channels.

Maybe this is actually correct and the rhythm output is supposed to be written to specifically `buffers[6]`. If that's the case, then the code needs to call it out explicitly, because it looks like a bug presently. :-)

### DIFF
--- a/player/fmopl2.c
+++ b/player/fmopl2.c
@@ -2057,9 +2057,10 @@ do { \
 \
 	vu_max[CHN] = MAX(vu_max[CHN], ab); \
 \
-	if (buffers[j]) { \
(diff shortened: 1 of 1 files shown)


mrpapersonic:
no, it's definitely supposed to be put in channel 0, since that's what the equivalent OPL3 code does.

Thanks

PR #749, opened 2025-08-12
OPL3: full channel support

logiclrd:
I noticed that if you ask Schism Tracker to play more than 9 AdLib notes concurrently, it doesn't work, even when built against the OPL3 chip with 18 channel support. I did some digging and uncovered why:

- `struct song` only tracks channel mappings for a hardcoded 9 OPL channels.
- `snd_fm.c` has explicit support for the second bank of channels (`OPL_Byte_RightSide`), but it never actually calls it.

This PR fixes this:

- `fmopl.h` now defines macros that capture the size of an OPL channel bank and how many channels there are (can be more than one bank).
- `struct song` now tracks channel mappings for `OPL_CHANNELS` channels, whether that's 9 or 18.
- `OPLUpdateOne` now only bothers mapping as many `buffers` as there are `OPL_CHANNELS`.
- `Fmdrv_Mix` now processes & mixes `OPL_CHANNELS` channels.
- All of the various OPL control methods use `OPL_Byte_RightSide` for channels outside of the first bank.
- `OPL_Detect` sends the same timer controls to both sides. When built against OPL2, the address for the second bank is simply ignored. I don't know if this is actually necessary or not.

Tested working with OPL2 (limited, of course to 9 channels) and OPL3 (full 18 channels supported). Now we just need a file format that isn't limited to 9 channels :-P Experimentally, I found that if you write FM notes in more than 9 channels with an S3M file, it actually does save all of your notes, but it marks channels that use FM notes past the 9th one as disabled. You can re-enable them when reloading the file.

### DIFF
--- a/include/player/fmopl.h
+++ b/include/player/fmopl.h
@@ -5,6 +5,16 @@
 
 #define logerror(...) /**/
 
+#define OPL_BANK_SIZE 9
+
+#if OPLSOURCE == 2
+# define OPL_CHANNELS 9
+#elif OPLSOURCE == 3
+# define OPL_CHANNELS 18
+#else
+# error "The current value of OPLSOURCE isn't supported! Check build-config.h."
+#endif
+
 typedef int16_t OPLSAMPLE;
 
 typedef void (*OPL_TIMERHANDLER)(void *param,int timer,double period);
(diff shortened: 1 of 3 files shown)


logiclrd:
`./configure --enable-opl3`

https://github.com/user-attachments/assets/86f702f2-0705-4d9e-8b01-414f304b7335

logiclrd:
`./configure --enable-opl2` (with #751 as well)

https://github.com/user-attachments/assets/430b827d-0e70-4483-acca-141e10a285eb

logiclrd:
@mrpapersonic Can we get this merged before doing other FM-related stuff, since this standardizes a count of FM channels and treatment of associated arrays?

logiclrd:
Rebased.

PR #750, opened 2025-08-12
opl: play first note

logiclrd:
The first AdLib note triggered in a channel doesn't play. The reason is that it checks the channel's flags to see if it is marked as an AdLib channel. If it isn't, it doesn't call the OPL handling. _Then_ it updates the channel's flags. This PR updates it to check the sample instead of the channel. (The sample is where the channel gets its status from in the next line anyway.)

### DIFF
--- a/schism/audio_playback.c
+++ b/schism/audio_playback.c
@@ -534,12 +534,13 @@ static int song_keydown_ex(int samp, int ins, int note, int vol, int chan, int e
 	if (i)
 		csf_check_nna(current_song, chan_internal, ins, note, 0);
 	if (s) {
-		if (c->flags & CHN_ADLIB) {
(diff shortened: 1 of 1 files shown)


mrpapersonic:
I think this is still wrong. The note off code everywhere else does
```
if (c->flags & CHN_ADLIB) {
    OPL_NoteOff(csf, chan_internal);
    OPL_Touch(csf, chan_internal, 0);
}
```

so we should follow suit. Maybe also put the call to `OPL_Patch` under a `s->flags & CHN_ADLIB` conditional

logiclrd:
Mm, because the `OPL_NoteOff` is related to what's already happening on the channel, but the `OPL_Touch` call is needed based on what's in the sample, not what's previously in the channel.

PR #751, opened 2025-08-12
fmopl2.c: fix some bitrot

logiclrd:
Evidently, the OPL2 code does not get compiled and tested with any regularity. It hasn't been exercised since before the code was updated to use `-Werror=old-style-definition`.

Perhaps a check could be set up that does the build with `./configure --enable-opl2`, so that if someone breaks it but doesn't have it enabled locally, they have a way to find out about it.

### DIFF
--- a/player/fmopl2.c
+++ b/player/fmopl2.c
@@ -73,6 +73,8 @@ Revision History:
 
 #include "player/fmopl.h"
 
+#include "bits.h"
+
 // XXX why is this here?
 #include "log.h"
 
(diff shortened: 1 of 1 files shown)


logiclrd:
The alternative would be to simply rip out the OPL2 code, if the OPL3 code is a strict superset of it anyway. What does the OPL2 code add that the OPL3 code doesn't cover? I don't know enough about it to know if that's the case. :-P

mrpapersonic:
i do test it every time i update the OPL stuff, but i guess my compiler just sucks and didn't error for some reason

and `safe_abs_32` will compile, since it's just a deprecated macro expanding to `babs32`, and I was too lazy to change all of the references

PR #753, opened 2025-08-12
sndmix: fix initialization of current_tempo in csf_read_note

logiclrd:
I can't think of a scenario where this wouldn't be wrong. :-P


mrpapersonic:
I'd be surprised if `csf->initial_tempo` could ever be 0 to begin with

logiclrd:
That's probably why this hasn't come to anybody's attention until now. :slightly_smiling_face:

PR #754, opened 2025-08-12
effects: fix assignment of volume to midi_vol_tracker

logiclrd:
Looks like an autocomplete error. The `midi_vol_tracker` tracks volumes, not effect types. Just three lines down there's a correct-looking assignment.


mrpapersonic:
This whole function is a load of spaghetti with magic numbers everywhere (e.g. `120` should be `NOTE_LAST`), and tbh I have no idea how much of it is still used, considering `snd_gm.c` SHOULD be doing all of the midi crap now? .............. idk

PR #757, opened 2025-08-14
midi-ip: uninitialized loop iterator fix

logiclrd:


mrpapersonic:
stylistic choice really, but initialization of `q` ought to be next to the iteration, since it relies on the existing value

logiclrd:
Fair enough :-)

mrpapersonic:
I wonder if there's some kind of compiler attribute that says "this absolutely has to be initialized before its passed as a parameter". It would be *really* useful to warn in cases like this ...

PR #764, opened 2025-08-17
fmt/med: implemented reading song title

logiclrd:
image

[Aliens_Par-Tay_Too2.zip](https://github.com/user-attachments/files/21824471/Aliens_Par-Tay_Too2.zip)

[MMD_FileFormat.txt](https://github.com/user-attachments/files/21824375/MMD_FileFormat.txt)

### DIFF
--- a/fmt/med.c
+++ b/fmt/med.c
@@ -25,10 +25,6 @@
 #include "fmt.h"
 #include "mem.h"
 
-// the most characters we will try to read for the song name.
-// we only care about 26 characters of song name, right??
-#define MAX_NAME_LENGTH 26
-
 #define MMD0_ID "MMD0"
 
 struct MMD0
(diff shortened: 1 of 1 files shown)


fmt/med.c
```diff
@@ -25,17 +25,98 @@
 #include "fmt.h"
 #include "mem.h"
 
+// the most characters we will try to read for the song name.
+// we only care about 26 characters of song name, right??
```
mrpapersonic: what I do in the info readers is read as much as allowed and str_dup it into a buffer (or mem_alloc a buffer and read into it, both work but allocing first needs extra stuff to account for read fails)

if we're going to define an upper bounds for this it ought to be in `dmoz.h` where all of the readers can share it

fmt/med.c
```diff
@@ -25,17 +25,98 @@
...
+#define MAX_NAME_LENGTH 26
+
+#define MMD0_ID "MMD0"
+
+struct MMD0
+{
+	/* 0 */  char     id[4];
```
mrpapersonic, line 32: we probably shouldn't be loading a whole bunch of stuff we don't use if we don't have to. Just seek to the expdata pointer, read it in, and seek to it. Same with the name pointer in the exp structure.

(I don't see any major padding issues here, but it's good practice to NOT assume that the way the structure is laid out in memory will be the exact same as the disk format. In the past we used `#pragma pack(1)`, but this isn't very portable to oddball compilers, so I dumped all of that out.)

fmt/med.c
```diff
@@ -25,17 +25,98 @@
...
 
 int fmt_med_read_info(dmoz_file_t *file, slurp_t *fp)
 {
-	unsigned char magic[4];
-	if (slurp_read(fp, magic, sizeof(magic)) != sizeof(magic)
-		|| memcmp(magic, "MMD0", 4))
+	struct MMD0 hdr;
+	struct MMD0exp exp;
+
+	int length = slurp_length(fp);
```
mrpapersonic: `int` -> `uint64_t`

fmt/med.c
```diff
@@ -25,17 +25,98 @@
...
+
+	if (sizeof(hdr) > length)
+		return 0;
+
+	if (slurp_read(fp, &hdr, sizeof(hdr)) != sizeof(hdr)
+	 || memcmp(hdr.id, MMD0_ID, 4))
+		return 0;
+
+	int exp_struct_ptr = bswapBE32(hdr.expdata_ptr);
```
mrpapersonic: `int` -> `uint32_t`

fmt/med.c
```diff
@@ -25,17 +25,98 @@
...
+
+	if ((exp_struct_ptr + sizeof(struct MMD0exp) >= length)
+	 || slurp_seek(fp, exp_struct_ptr, SEEK_SET)
+	 || (slurp_read(fp, &exp, sizeof(exp)) != sizeof(exp)))
+		return 0;
+
+	int name_ptr = bswapBE32(exp.songname_ptr);
+	int name_len = bswapBE32(exp.songnamelen);
```
mrpapersonic: `int` -> `uint32_t`

mrpapersonic (changes requested):

fmt/med.c
logiclrd: That's fair. Better to have the generic `fmt_read_info` function not apply any conventions; they're out of scope.

fmt/med.c
logiclrd: Okay, that's fair. The flipside is that there's the outside chance that a full OctaMed loader might be implemented, and having the full header types could be useful. Maybe I'll leave the struct definition there but only read the fields needed at this point.

mrpapersonic: yep, that's fine. That's what we do in many of the other loaders anyway (loading in each element one by one)

mrpapersonic:
thanks for this!

btw, there's no plans to ever implement a med loader, because there's just too many different variations of the format, all with their own [weirdness](https://github.com/OpenMPT/openmpt/blob/master/soundlib/Load_med.cpp). Having to read 2 different pointers to get the song name is already weird :)

logiclrd:
Fair enough :grin:

PR #765, opened 2025-08-17
song_get_pattern_offset: fix wrap point for pattern rows when MODE_PATTERN_LOOP

logiclrd:
When I reworked this code, I preserved the logic. The old code did `row = (row + offset) % song_get_rows_in_pattern`, and the rework I did was because `song_get_rows_in_pattern` returned, unintuitively, the maximum row number, _not_ the actual number of rows in the pattern but one less than it. As part of f39fe3434, I eliminated a division by 0 error when the pattern's length is 1, because `song_get_rows_in_pattern` would then return 0. But, I think the correct fix is just to take the modulus with the _actual length of the pattern_, not one less than it. Am I wrong? :-P

Technically speaking, this code could now once again have a division by 0, but only if `*n` is an out-of-range number for which the return value becomes -1. This function is actually only called from one place, and the value stored in `n` before the call is `current_pattern`. I think if `current_pattern >= MAX_PATTERNS` then we have bigger problems :-)


mrpapersonic:
this thing really needs tests, though it's actually only used in one place (for incoming midi events). unfortunately all of my midi gear is back at home, and I'm at college for the time being, so it'll be a while until I can test this branch properly.

> Technically speaking, this code could now once again have a division by 0, but only if *n is an out-of-range number for which the return value becomes -1.

we'd be screwed anyway. I think it would probably be fine to just assign to `tot` and have a runtime assertion that the return value is valid. At least then we would be able to go into the assertion branch. also these variables desperately need renaming, `n` for pattern number is stupid

note to self: the assertion function really needs to attempt to save the song like the crash handler already does

logiclrd:
Rebased to bring in `song_get_max_row_number_in_pattern` change upstream.

test/cases/mplink.c
```diff
@@ -0,0 +1,237 @@
...
+		2, 16, // starting from
+		49,    // advance by
+		4, 0); // expect to be at
+}
+
+testresult_t test_song_get_pattern_offset_song_LAST(void)
+{
+	return test_song_get_pattern(
+		2, 15, // starting from
+	  113,    // advance by
```
mrpapersonic: Tabs? :)

logiclrd: Oh ship! :ship:

PR #772, opened 2025-08-26
Fix str_from_num

logiclrd:
This PR fixes global economic collapse.

(Specifically, `str_from_num` was switched to `snprintf` but the buffer length was accidentally copy/pasted as 3, instead of the 11 it needs to be in this function, and that function is called during initialization to set the SDL buffer size hint. As a result, instead of 1024-byte buffer, SDL was initialized with a 10-byte buffer. This caused famine in Africa and further entrenched the authoritarian regime in China.)

### DIFF
--- a/schism/str.c
+++ b/schism/str.c
@@ -152,7 +152,7 @@ char *str_from_num(int digits, uint32_t n, char buf[11])
 		snprintf(buf, digits + 1, fmt, n);
 		buf[digits] = 0;
 	} else {
-		snprintf(buf, 3, "%" PRIu32, n);
+		snprintf(buf, 11, "%" PRIu32, n);
 	}
 	return buf;
 }


mrpapersonic:
oops

logiclrd:
This was a really obscure one! `git bisect` said that it was a bunch of text manipulation that caused these weird sample artefacts and audio output failures. How in the heck? But, sure enough, switched back and forth between the two commits, and the `snprintf` commit reliably failed. So then, having done a `git bisect` to narrow it down to the commit, I now had a commit that touched, what, some 30 different files. I used PowerShell to hack together a quick bisect-within-the-commit function, trying the first of 15 those files but not the second 15, and then whichever half had the problem-causing line, split that into two sets and try them, and so on, and thereby narrowed it down to that particular line in only 5 assays. The question remained, though: How on earth does that change to `str_from_num`, _even if it is buggy_, cause this behaviour so deep in the sound? I mean, it was sounding _weird_, I thought for sure something had gone horribly wrong computing loop offsets or something. Put a breakpoint on it, fired it up, and it immediately hit it from the SDL code. :-P

PR #774, opened 2025-08-28
Test cases and semantic assertions

logiclrd:
This PR augments the test system in two ways:

* Assertions are now semantic.

    Instead of all assertions being simply `ASSERT(some condition)`, you can now specify the meaning behind the condition.
    ```
    ASSERT_EQUAL_D(result, 37);
    ASSERT_EQUAL_D_NAMED(result, OPEN_SUCCESS);
    ASSERT_EQUAL_S(data, "monkey");
    ASSERT_TRUE(slurp_eof(fd));
    // etc.
    ```
    The resulting assertion messages are then much more meaningful:
    ```
    TEST: test_str_from_num_thousands[22] ............................ FAIL (0 ms)
    
    TEST LOG: test_str_from_num_thousands[22] (n == -5834) (test/cases/str.c:80): assertion failed: result
     expected: "-5,834"
     actual: "-5,835"
    
    
    TEST: test_slurp_stdio ........................................... FAIL (0 ms)
    
    TEST LOG: test_slurp_stdio (test/cases/slurp.c:202): assertion failed: slurp_stdio(&fp, stdfp)
     expected: SLURP_OPEN_SUCCESS (1)
     actual: 0
    ```
    (actual output created by making the subject methods return a spuriously incorrect result)

* Tests can now be declared to have an array of test cases.

    The harness runs each one in turn with a separate call to the test method (i.e., a separate child process, for isolation).

    This is done by:
    - Defining the test method using the macro `TEST_CASES` for the signature line:
        ````
        TEST_CASES(test_str_concat, 2)
        {
        	static const struct {
        		const char *args[5];
        		const char *expected;
        	} cs[] = {
        		{{"123", "456", "789", "012"}, "123456789012"},
        		{{"123", "ok", "67"}, "123ok67"},
        	}, *c;
        
        	c = &cs[test_index];
        	size_t i;
        	char *r;
        
        	test_set_name("%s(\"%s\", \"%s\", \"%s\", \"%s\", \"%s\")",
        		test_get_name(),
        		c->args[0] ? c->args[0] : "NULL",
        		c->args[1] ? c->args[1] : "NULL",
        		c->args[2] ? c->args[2] : "NULL",
        		c->args[3] ? c->args[3] : "NULL",
        		c->args[4] ? c->args[4] : "NULL");
        
        	r = str_concat(c->args[0], c->args[1], c->args[2], c->args[3], c->args[4], (char *)NULL);
        
        	ASSERT_NOT_NULL(r);
        	ASSERT_EQUAL_S(r, c->expected);
        
        	free(r);
        
        	RETURN_PASS;
        }

    (for parity, there is also a `TEST(x)` macro that generates the signature line of a simple test)
    - Adding a matching declaration of the test to `test-funcs.h`, using `TEST_CASE_FUNC` instead of `TEST_FUNC`:
        ```
        TEST_CASE_FUNC(test_str_concat, 2)
        ```

    If the counts fall out of sync, a compile-time error is generated.

    One added feature, to support tracking the test indices, is the concept of the "ambient test name". The harness sets this to `test_functor` or `test_functor[N]`, but the test can override it. You can see an example of this in the failure message in the first bullet point above for `test_str_from_num_thousands[22]` -- the test appends ` (n == %d)` to the end of the test name, and then when the assertion fails, the value of n is part of the output.

All existing tests are updated (in a separate commit) to follow these conventions.


include/test-assertions.h
```diff
@@ -30,27 +30,89 @@
 
 void test_assert(const char *file, long line, const char *cond, const char *msg, const char *fmt, ...);
 
-/* TODO prefix these with TEST_ */
-#define REQUIRE_PRINTF(cond, fmt, ...) \
+#define _TEST_ASSERTION(cond, actual, return_value, msg, fmt, ...) \
```
mrpapersonic: don't prefix anything with underscores (reserved by the implementation and it can cause weird conflicts)

include/test-assertions.h
```diff
@@ -30,27 +30,89 @@
...
-			test_assert(__FILE__, __LINE__, #cond, "test initialization failed", fmt, __VA_ARGS__); \
-			return SCHISM_TESTRESULT_INCONCLUSIVE; \
+			test_assert(__FILE__, __LINE__, #actual, msg, fmt, __VA_ARGS__); \
+			return return_value; \
 		} \
 	} while (0)
 
+#define _TEST_INITIALIZATION_ASSERTION(cond, actual, fmt, ...) \
+	_TEST_ASSERTION(cond, actual, SCHISM_TESTRESULT_INCONCLUSIVE, "test initialization failed", fmt, __VA_ARGS__)
+#define _TEST_ASSERTION_ASSERTION(cond, actual, fmt, ...) \
```
mrpapersonic: weird, i wasn't able to get recursive variadic macros working. Maybe I was just doing something totally wrong

include/test-assertions.h
```diff
@@ -30,27 +30,89 @@
...
+#define ASSERT_TRUE(actual) _TEST_ASSERTION_ASSERTION(actual, actual, " expected: true (non-zero)\n actual: false (0)", NULL)
+#define REQUIRE_FALSE(actual) _TEST_INITIALIZATION_ASSERTION(!actual, actual, " expected: false (0)\n actual: true (non-zero)", NULL)
+#define ASSERT_FALSE(actual) _TEST_ASSERTION_ASSERTION(!actual, actual, " expected: false (0)\n actual: true (non-zero)", NULL)
+
+#define REQUIRE_EQUAL_SD(actual, expected) do { int16_t a = actual, e = expected; _TEST_INITIALIZATION_ASSERTION(a == e, actual, " expected: %" PRId16 "\n actual: %" PRId16, e, a); } while (0)
+#define ASSERT_EQUAL_SD(actual, expected) do { int16_t a = actual, e = expected; _TEST_ASSERTION_ASSERTION(a == e, actual, " expected: %" PRId16 "\n actual: %" PRId16, e, a); } while (0)
+#define REQUIRE_EQUAL_D(actual, expected) do { int32_t a = actual, e = expected; _TEST_INITIALIZATION_ASSERTION(a == e, actual, " expected: %" PRId32 "\n actual: %" PRId32, e, a); } while (0)
+#define ASSERT_EQUAL_D(actual, expected) do { int32_t a = actual, e = expected; _TEST_ASSERTION_ASSERTION(a == e, actual, " expected: %" PRId32 "\n actual: %" PRId32, e, a); } while (0)
```
mrpapersonic: can these all be collapsed into one thing without repeating all of this? maybe with variants for the "real" C types (`char`, `signed char`, `unsigned char` ...)

also IMO we these should be `I16X`/`S16X` and `U16X` instead of `SX`. the one for pointers is fine.

mrpapersonic (changes requested):

include/test-assertions.h
logiclrd: Huh, okay. I mean, seems unlikely in this specific case, but fair, good principle & rule to follow.

mrpapersonic:
also, with the array-based stuff, we could also use a macro template, e.g. `test_str_concat` would go from this:
```c
testresult_t test_str_concat(void)
{
	static const struct {
		const char *args[5];
		const char *expected;
	} c[] = {
		{{"123", "456", "789", "012"}, "123456789012"},
		{{"123", "ok", "67"}, "123ok67"},
	};
	size_t i;
	char *r;

	for (i = 0; i < ARRAY_SIZE(c); i++) {
		r = str_concat(c[i].args[0], c[i].args[1], c[i].args[2], c[i].args[3], c[i].args[4], (char *)NULL);
		REQUIRE(r);

		ASSERT_PRINTF(!strcmp(c[i].expected, r), "result \"%s\" with args "
			"(\"%s\", \"%s\", \"%s\", \"%s\", \"%s\") was not \"%s\" as expected",
			r,
			c[i].args[0] ? c[i].args[0] : "NULL",
			c[i].args[1] ? c[i].args[1] : "NULL",
			c[i].args[2] ? c[i].args[2] : "NULL",
			c[i].args[3] ? c[i].args[3] : "NULL",
			c[i].args[4] ? c[i].args[4] : "NULL",
			c[i].expected);

		free(r);
	}

	RETURN_PASS;
}
```
to this:
```c
#define TEST_STR_CONCAT_TEMPLATE(name, arg1, arg2, arg3, arg4, arg5, expected) \
	testresult_t test_str_concat_##name(void) \
	{ \
		char *r; \
	\
		r = str_concat(arg1, arg2, arg3, arg4, arg5, (char *)NULL); \
		REQUIRE(r); \
	\
		ASSERT_PRINTF(!strcmp(expected, r), "result \"%s\" with args " \
			"(\"%s\", \"%s\", \"%s\", \"%s\", \"%s\") was not \"%s\" as expected", \
			r, \
			(arg1) ? (arg1) : "NULL", \
			(arg2) ? (arg2) : "NULL", \
			(arg3) ? (arg3) : "NULL", \
			(arg4) ? (arg4) : "NULL", \
			(arg5) ? (arg5) : "NULL", \
			expected); \
	\
		free(r); \
	\
		RETURN_PASS; \
	}

TEST_STR_CONCAT_TEMPLATE(123456789012, "123", "456", "789", "012", NULL, "123456789012")
TEST_STR_CONCAT_TEMPLATE(123ok67, "123", "ok", "67", NULL, NULL, "123ok67")

#undef TEST_STR_CONCAT_TEMPLATE
```
we could even move the guts of the functions to a separate one to save space, like we already do with `test_str_from_num_thousands` and friends

include/test-assertions.h
logiclrd: Mm, okay. I had named them after the `printf` specifiers, with the addition of `%sd` for a 16-bit `%d`.

As for collapsing them, well, they're all different in different ways, such that I think separating out all the ways they can differ would likely not _hugely_ reduce the length of the lines while making maintenance worse. I'll see what it looks like, though.

logiclrd:
> also, with the array-based stuff, we could also use a macro template, e.g. test_str_concat would go from this:

I don't believe this is an improvement. You can't trace through a macro expansion in a debugger, at least not the ones I've used. You can put a breakpoint on `r = str_concat(...);`, but not on `r = str_concat(...); \` inside a macro.

mrpapersonic:
my favorite debugger, printf, works fine in macros

Anyway, we could shove everything into a separate function (like `str_from_num_thousands`), which would make breakpoints work, and just have the test cases be stubs that call into it. We could also use a macro for reducing the amount of lines they take up (around 4 lines to only 1 line)

include/test-assertions.h
logiclrd: Is it that much of an improvement?
```
#define GENERIC_EQUALITY_ASSERTION(actual, expected, T, M, F) do { T a = actual, e = expected; M(a == e, actual, " expected: %" F "\n actual: %" F, e, a); } while (0)

#define REQUIRE_EQUAL_SD(actual, expected) GENERIC_EQUALITY_ASSERTION(actual, expected, int16_t, SCHISM__TEST_INITIALIZATION_ASSERTION, PRId16)
#define ASSERT_EQUAL_SD(actual, expected) GENERIC_EQUALITY_ASSERTION(actual, expected, int16_t, SCHISM__TEST_ASSERTION_ASSERTION, PRId16)
#define REQUIRE_EQUAL_D(actual, expected) GENERIC_EQUALITY_ASSERTION(actual, expected, int32_t, SCHISM__TEST_INITIALIZATION_ASSERTION, PRId32)
#define ASSERT_EQUAL_D(actual, expected) GENERIC_EQUALITY_ASSERTION(actual, expected, int32_t, SCHISM__TEST_ASSERTION_ASSERTION, PRId32)
```

logiclrd: And that _only_ factors in the basic set. `_NAMED` either requires additional complexity in `GENERIC_EQUALITY_ASSERTION` or a different macro. I suppose different macros isn't as bad as the complexity...

logiclrd: I have come to terms with it, after the realization that the common macro can be multiline :-)

logiclrd:
I'm not sure I quite follow -- you're saying that instead of the arrays of test cases, as you've introduced recently and I took up in this PR, you'd prefer to see the test cases split out into different exported symbols as with my initial `str_from_num_thousands` approach??

mrpapersonic:
well, if we're going to split them up, that's really the best way to do it. Other methods introduce unnecessary complexity and hax...

mrpapersonic:
sort of like this
```c
#define TEST_CASE_STUB(name, tocall, ...) \
    testresult_t test_##name(void) \
    { \
        return tocall(__VA_ARGS__); \
    }

#define STR_CONCAT_CASE(name, expect, arg1, arg2, arg3, arg4, arg5) \
    TEST_CASE_STUB(str_concat_##name, str_concat_common, expect, arg1, arg2, arg3, arg4, arg5)

STR_CONCAT_CASE(123ok67, "123ok67", "123", "ok", "67", NULL, NULL)
...
```
and all the logic for test_str_concat be moved into a str_concat_common

logiclrd:
So, ditch the test case infrastructure then?? :-P

logiclrd:
Hmm, how to get from that to the main index of tests? The only way I can see is to have the `CASE` macro invocations be in a separate file from the actual test method. That feels messy... :-/

mrpapersonic:
f78ef5e5a150eba7895de86f933064a30dbf4988 this is what I had in mind :)

logiclrd:
Mm, the problem with that is that there are two disconnected definitions of the test entrypoints. They can (and will) fall out of sync. This exact problem existed as recently as an earlier version of this very test suite; there were test functions defined and not being called because they weren't in the index.

In my latest commit, I've come up with a way to address this problem. A macro is used to generate test thunks that present as a regular test method with the `(void)` signature and call an underlying test implementation with arguments defining a test case. This same macro can be redefined to create just declarations, and to generate the index structure. Thus there is a single source of truth.

logiclrd:
Question: do all platforms Schism Tracker is built against use glibc?

logiclrd:
If this approach is acceptable, can you back out f78ef5e so that I don't have to do a messy rebase? :-P

logiclrd:
Oh, I guess I could just do a rebase and insert a revert up front. :-)

logiclrd:
```
../test/name.c:70:2: error: call to undeclared function 'register_printf_specifier'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        register_printf_specifier('Q', quote_string, quote_string_arg_counter);
        ^
```
Okay well that answers the glibc question :-P Guess I can't just casually go around doing `register_printf_prefix`. Shucks.

logiclrd:
I'm not sure if you're aware of this, but after participating in discussion on this PR, providing feedback that solicited updates, what you did on the mainline chopped the legs out from under this PR, and it was a bit of a faux pas. I put like 6 hours of work in, produced a 2700-line diff, updated it based on your feedback, and then you went and pushed 6 commits that are _entirely_ incompatible with pretty much everything in this PR. The message, intentional or not, is that you actually don't see this PR as important or desirable at all. Normally, one refrains as much as possible from making changes that make a PR unmergeable. You could, for instance, have pushed commits demonstrating your approach to this PR's branch instead of the mainline. Unless you were, in fact, deliberately trying to send the message that you don't care for the changes in this PR.

I articulated the reasoning behind this approach. To reiterate, the approach you have right now has multiple sources of truth, and they can fall out of sync. It's entirely possible to, say, create new test cases for a test, and then completely forget to link them into the test index. The changes in this PR make that impossible: The act of adding the new test cases also adds them to the test index.

mrpapersonic:
> I'm not sure if you're aware of this, but after participating in discussion on this PR, providing feedback that solicited updates, what you did on the mainline chopped the legs out from under this PR, and it was a bit of a faux pas. I put like 6 hours of work in, produced a 2700-line diff, updated it based on your feedback, and then you went and pushed 6 commits that are entirely incompatible with pretty much everything in this PR. The message, intentional or not, is that you actually don't see this PR as important or desirable at all. Normally, one refrains as much as possible from making changes that make a PR unmergeable. You could, for instance, have pushed commits demonstrating your approach to this PR's branch instead of the mainline. Unless you were, in fact, deliberately trying to send the message that you don't care for the changes in this PR.

i think the assertions for equal value are fine, since it'll just reduce clutter. That part in itself should be relatively easy to put into the master branch, and I might do it myself too.

> I articulated the reasoning behind this approach. To reiterate, the approach you have right now has multiple sources of truth, and they can fall out of sync. It's entirely possible to, say, create new test cases for a test, and then completely forget to link them into the test index. The changes in this PR make that impossible: The act of adding the new test cases also adds them to the test index.

This part I don't really care much about. This happens everywhere! I guess we could combine the declaration and definition, but can't it just be caught by e.g. `-Wmissing-declarations` ?

logiclrd:
I don't think `-Wmissing-declarations` helps in this case. The problem isn't that something, somewhere, calls a function without it having been declared. It's the exact opposite: something exports a function but nothing calls it. Is there a warning for that?

logiclrd:
And, it doesn't happen everywhere, because most places, when you add functionality, it's because you _actually need that functionality_. You're not just going to forget to call it, because the feature won't function until it's called. Tests, though, add nothing directly to the main application. You could shut the entire test suite off and not notice in main application development. That's why it's more important there to safeguard against oversights.

mrpapersonic:
> I don't think `-Wmissing-declarations` helps in this case. The problem isn't that something, somewhere, calls a function without it having been declared. It's the exact opposite: something exports a function but nothing calls it. Is there a warning for that?

That's exaxtly what `-Wmissing-declaration` does, it warns if there is an exported function without a declaration before it. Hell, we could force-enable this warning as an error specifically for the test suite.

we should probably force-enable this as an error everywhere though...

logiclrd:
Hmm, okay, well, it's _not_ exactly what it does. There's a difference between having the linker warn if a symbol is exported but never consumed and having the compiler warn that you exported a symbol without declaring it. But, if one file generates both the usage (the reference from the test index) and the declaration, and then the files where the tests are defined _don't_ declare them themselves, then I guess it adds up to the same thing: If you don't put it into the index, then you get an error.. Okay, could work :-P

logiclrd:
Rebased and squashed down all the iterative changes.

This version:

- Enables `-Werror=missing-declarations`.
- Declares all tests when `test.h` is included (by means of `test-funcs.h`).
- Defines all test cases using explicit thunks, à les tests `str`.
- Has the latest iteration of the assertion macros, factored out to central definitions to avoid repetition.
- Updates all of the tests to use semantic assertions.

I also explicitly verified that if I add a test to a `test/*.c` file and then "forget" to add it to `test-funcs.h`, a compile-time error is generated.

logiclrd:
Okay well that was curious. I accidentally pushed my test undeclared test up, because I had stopped getting build errors and thought I had removed it. Then the automated tests fired up and a dozen of them passed before the Linux one failed. It looks like `-Werror=missing-declarations` isn't entirely reliable... I guess as long as one of the automated builds catches it, though... But if a PR isn't used and the code is built in a setting where `-Werror=missing-declarations` isn't doing its job, then tests that aren't actually getting called could sneak in...

mrpapersonic:
some of them use old versions of gcc (the OS/2 build doesn't even use gcc!) and many of the CI runs do not run the test suite, since it's run independently from regular builds on everything but linux

OS X uses clang and i dont know whether it properly implements the warning even though it ought to

logiclrd:
Ahh, that explains it then. Cool cool

PR #784, opened 2025-09-18
midi/win32: identify duplex midi ports and treat them accordingly

mrpapersonic:
> hexadecimator: wonder if there's a way to tell midi input from output without trial and error, cause both input and output have the same name on my list xd

100% untested, but probably works.

*note to self*: CoreMIDI has the same issue, but it uses an opaque pointer `MIDIEndpointRef`, which MAY be used to identify duplex ports, but I'm not sure

### DIFF
--- a/sys/win32/midi-win32mm.c
+++ b/sys/win32/midi-win32mm.c
@@ -51,7 +51,8 @@ union win32mm_ocp {
 };
 
 struct win32mm_midi {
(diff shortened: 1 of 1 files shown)


mrpapersonic:
ok, fixed it up, seems to work now, so applied to master: af279945b3b18b7b634f9e594cd938564a0353a1

Issue #788, opened 2025-10-19
Mouse cursor very rarely changes to envelope cursor; need to click back into an envelope to restore it


mrpapersonic:
An easy way to trigger this is by editing an envelope with the mouse, and then switching to another page.

mrpapersonic:
Fixed with #790

PR #790, opened 2025-10-22
Properly handle cursor shape switching

hotgar:
#788 
Switch back to arrow mouse cursor shape if another page or menu are opened while editing an envelope.


mrpapersonic:
Thanks!!

PR #824, opened 2026-05-22
headers: restore mingw-w64 _mingw_off_t.h workaround for unistd.h

esaruoho:
Restore the mingw-w64 `<_mingw_off_t.h>` workaround on the `keybinds` branch. Master already carries this; the branch lost it during the wider `headers.h` simplification.

## Symptom

Without this fix, mingw64 and ucrt64 CI fail on every translation unit that includes `headers.h`:

```
D:/a/_temp/msys64/mingw64/include/unistd.h:59:20: error:
  unknown type name 'off_t'; did you mean '_off_t'?
```

[Upstream mingw-w64 bug](https://sourceforge.net/p/mingw-w64/bugs/1014/) landed November 2025: `unistd.h` references `off_t` without pulling in its definition. Newer msys2 toolchains then fail to compile anything that goes through this header.

## Why this branch and not master

Master already has the workaround block in `include/headers.h` (cherry-picked when the upstream bug landed). The `keybinds` branch's last green Windows CI run is from 2025-07-30 — before the mingw bug existed — so the regression wasn't visible until something else re-triggered CI this week.

## What the commit does

One commit, +21/-4 lines in `include/headers.h`:

1. Move `#include <build-config.h>` ahead of the (new) mingw block, because `SCHISM_WIN32` is defined in `build-config.h` and is what gates the workaround. (On master this ordering was already correct.)
2. Insert the mingw workaround block — `undef NO_OLDNAMES; include <_mingw.h>; conditionally include <_mingw_off_t.h>` — between `build-config.h` and the existing `NO_OLDNAMES` defines. The block is gated on `defined(SCHISM_WIN32) && defined(__MINGW32__)` so it's a no-op everywhere else.

Verified: builds clean on macOS arm64 with SDL2/SDL3/SDL12 locally, no behavioural change anywhere except mingw.

## Relationship to #823

Carving this out so it can land on its own — it's pure regression fix with no design content. I'll drop the broken first attempt at this fix from #823 once this merges and rebase.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/headers.h
+++ b/include/headers.h
@@ -34,6 +34,27 @@
 #define _GNU_SOURCE /* need this for some stupid gnu crap */
 #endif
 
+/* build-config.h must be included before the mingw workaround below
+ * because that's where SCHISM_WIN32 is defined. */
+#ifdef HAVE_CONFIG_H
+# include <build-config.h>
+#endif
+
+#if defined(SCHISM_WIN32) && defined(__MINGW32__)
+/* Work around a mingw-w64 bug introduced in nov 2025 where unistd.h
+ * references off_t without pulling in its definition. Including
+ * _mingw_off_t.h here (while NO_OLDNAMES is not yet defined) gets
+ * off_t typedef'd before anything else touches unistd.h. The defines
+ * just below this block then suppress the rest of the OLDNAMES
+ * pollution.
+ * upstream bug: https://sourceforge.net/p/mingw-w64/bugs/1014/ */
+# undef NO_OLDNAMES
+# include <_mingw.h>
+# ifdef __MINGW64_VERSION_MAJOR
+#  include <_mingw_off_t.h>
+# endif
+#endif
+
 #ifndef NO_OLDNAMES
 /* Mingw-w64 */
 # define NO_OLDNAMES
@@ -41,10 +62,6 @@
 #ifndef _NO_OLDNAMES
 /* Mingw.org */
 # define _NO_OLDNAMES
-#endif
-
-#ifdef HAVE_CONFIG_H
-# include <build-config.h>
 #endif
 
 /* ------------------------------------------------------------------------ */
--- a/sys/sdl2/init.h
+++ b/sys/sdl2/init.h
@@ -27,7 +27,21 @@
 // We have to include this **before** headers.h because
 // otherwise tgmath.h doesn't work quite right combined
 // with intrin.h under Win32. What the hell.
-#include <build-config.h>
+#ifdef HAVE_CONFIG_H
+# include <build-config.h>
+#endif
+
+/* Pull in just the mingw-w64 off_t workaround here, *before* the
+ * SDL include below transitively drags in <unistd.h>. We can't
+ * include the whole of headers.h yet because its <tgmath.h> conflicts
+ * with SDL's <intrin.h> on mingw — see the comment above.
+ * upstream bug: https://sourceforge.net/p/mingw-w64/bugs/1014/ */
+#if defined(SCHISM_WIN32) && defined(__MINGW32__)
+# include <_mingw.h>
+# ifdef __MINGW64_VERSION_MAJOR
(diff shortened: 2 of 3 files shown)


esaruoho:
CI update — the off_t fix is now confirmed working on all platforms. Took three pushes to land it because the bug had two layers:

1. **First push** (`d9656332`): restored the `<_mingw.h>` + `<_mingw_off_t.h>` workaround in `include/headers.h`, moved `<build-config.h>` ahead of it (master ordering). Necessary but not sufficient on this branch.

2. **Second push** (`50b75341`, force-replaced): moved `#include "headers.h"` ahead of `<SDL.h>` in `sys/sdl[23]/init.h`. This fixed mingw64/ucrt64 (`off_t` got defined before SDL transitively pulled in `<unistd.h>`) but **regressed mingw32-x86_64** — the comment I had dismissed as stale was actually accurate: `<tgmath.h>` (in headers.h) and `<intrin.h>` (in SDL) genuinely conflict on Win32. Master sidesteps this because `<tgmath.h>` is commented out in its `headers.h`; this branch re-enabled it.

3. **Third push** (`c69e8349`, current): surgical fix. Keep `<SDL.h>` before `headers.h` (preserving the tgmath/intrin ordering this branch needs), but pull in **just** `<_mingw.h>` + `<_mingw_off_t.h>` at the top of `init.h` so `off_t` exists before SDL touches `<unistd.h>`. Doesn't drag tgmath.h into scope early. Mirrors the conditional in `headers.h` so the workaround stays no-op everywhere except mingw.

Build state with this commit:
- macos / ubuntu / wii / wiiu / xbox / os2-i386 / mingw32-{armv7,aarch64,x86_64} / mingw64 / ucrt64 — **all build successfully** ✓

**Remaining red on `test (mingw64)` and `test (ucrt64)`:** these run `./schismtrackertest.exe` after the build. The binary links fine but segfaults at startup (`exit code 139`). This appears to be unrelated to the mingw header work — happens after build completion, only on the two msys2 test-suite jobs, doesn't reproduce on the macOS or other-platform test jobs. The branch's wider test-harness rewrite (commit `40841786` and `191514bc`) is the more likely cause; the `test/cases/*` deletions I flagged in #823 may be related. Not chasing that here since it's out of scope for the mingw fix.

Should I open a third PR investigating the test SEGV separately, or leave it for you?

esaruoho:
Tried merging `origin/master` into the branch locally as a quick check for whether the remaining test segfault is something master already fixed. Reporting findings rather than pushing — the merge has 5 conflicts and 2 of them need your call:

| File | Conflict | Notes |
|---|---|---|
| `include/util.h` | `crc32b` (this branch) vs `mem_xor` + `minmax_8/16/32` (master) | mechanical — both keepable |
| `schism/util.c` | `crc32b` impl vs SIMD-enhanced `mem_xor` family | mechanical — both keepable |
| `include/osdefs.h` | small `log_appendf` fragment | mechanical |
| `schism/main.c` | this branch added `keybinds_init()` at the exact spot where master refactored `did_classic` → `BITARRAY_ISSET(startup_flags, SF_DID_CLASSIC)` | **your call** — both changes are real, need to combine |
| `sys/sdl3/events.c` | this branch simplified the SDL3 keydown path; master added `SCHISM_SDL3_USE_COMPOSITION`-gated IME handling | **your call** — IME behaviour |

Aborted the merge — didn't seem right to do it on your behalf, especially since there's no actual evidence a master-merge would resolve the `schismtrackertest.exe` SIGSEGV (the test crashes inside `mt_init`/`timer_init` paths, and the branch's `schism/mt.c` + `sys/win32/mt.c` rewrites are independent of any conflict files above).

Leaving the test issue for you to triage. Build is solid across the matrix — that part is done.

mrpapersonic:
Just rebased the keybinds branch on top of master, so this is outdated anyway.

PR #830, opened 2026-05-31
Sinc256 interpolation

reales:
# Add 256-tap windowed-sinc interpolation mode

Adds a new `SRCMODE_SINC256` interpolation mode: a runtime 256-tap windowed sinc (Hann window, dynamic `1/speed` anti-alias cutoff), ported from HighLife. Exposed in the preferences page as **"256-Tap Sinc"**.

## Implementation

- New mixer function-table block (`MIXNDX_SINC256SRC`) covering all mono/stereo × 8/16-bit × ramp × filter variants.
- The kernel reads the real sample buffer with tap indices clamped to `[0, length)`, so out-of-range taps contribute zero.
- It does **not** depend on the loop lookahead buffers.

## Known tradeoffs

- **Loop seams:** because the lookahead buffers only guarantee a small amount of context, taps near a loop point read clamped/adjacent data rather than properly wrapped samples. Reads never go out of bounds, but the seam is softened rather than fully band-limited. Extending the lookahead infrastructure for >4-tap kernels would make this exact.
- **Cost:** computes sin/cos per tap (no LUT), single-threaded, ~64× the taps of the 8-tap FIR. Intended for offline/mastering renders; heavy for realtime with many voices. A LUT or SIMD pass would be a sensible follow-up.

## Also included

A small fix so `--headless --diskwrite` honors the configured interpolation mode. It previously always rendered as nearest-neighbour because the headless path skips `audio_init()`.

### DIFF
--- a/player/mixer.c
+++ b/player/mixer.c
@@ -99,6 +99,43 @@
 #include "player/precomp_lut.h"
 
 // ----------------------------------------------------------------------------
+// HighLife-style 256-tap windowed sinc (runtime, dynamic anti-alias cutoff)
+// ----------------------------------------------------------------------------
+
+#define SINC256_TAPS 256
+
+// One interpolated output point. `base`/`len`/`chans` describe the real sample
+// buffer so taps falling outside [0,len) read zero (clamped, loop-seam softens).
+// `cutoff` is the anti-alias cutoff (1/speed, clamped to 1) shared by both channels.
+#define DEFINE_SINC256_INTERP(bits) \
+	static inline SCHISM_ALWAYS_INLINE \
+	int32_t sinc256_interp##bits(const int##bits##_t *base, int32_t len, int chans, int ch, \
+		int32_t poshi, double frac, double cutoff) \
+	{ \
+		if (frac == 0.0) \
+			frac = 1.0e-25; \
+	\
+		double mix = 0.0; \
+		for (int s = -SINC256_TAPS; s <= SINC256_TAPS; s++) { \
+			int32_t idx = poshi + s; \
+			if (idx < 0 || idx >= len) \
+				continue; \
+	\
+			double x = (double)s - frac; \
+			double k_window = 0.5 + 0.5 * cos(x / (double)SINC256_TAPS * M_zPI); \
+			double k_sinc = sin(cutoff * M_zPI * x) / (M_zPI * x); \
+			mix += (double)base[idx * chans + ch] * k_sinc * k_window; \
+		} \
+	\
+		return (int32_t)mix; \
+	}
+
+DEFINE_SINC256_INTERP(16)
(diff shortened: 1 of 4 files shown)


mrpapersonic:
Does this actually fit on the preferences page? Last I checked there wasn't any space for it.

player/mixer.c
```diff
@@ -98,6 +98,60 @@
...
+// HighLife-style 256-tap windowed sinc (runtime, dynamic anti-alias cutoff)
+// ----------------------------------------------------------------------------
+
+#define SINC256_TAPS 256
+
+// One interpolated output point. `base`/`len`/`chans` describe the real sample
+// buffer so taps falling outside [0,len) read zero (clamped, loop-seam softens).
+// `cutoff` is the anti-alias cutoff (1/speed, clamped to 1) shared by both channels.
+static inline SCHISM_ALWAYS_INLINE
+int32_t sinc256_interp16(const int16_t *base, int32_t len, int chans, int ch,
```
mrpapersonic: you can collapse these two function definitions into a macro since the actual implementation is the exact same.

player/mixer.c
```diff
@@ -138,6 +192,14 @@
 	int32_t poshi  = csf_smp_pos_get_whole(position); \
 	int32_t poslo  = csf_smp_pos_get_frac(position) >> 16;
 
+#define SNDMIX_GETSINC256POS \
+	int32_t poshi   = csf_smp_pos_get_whole(position); \
+	double  sinc_frac = (double)csf_smp_pos_get_frac(position) / 4294967296.0; \
+	double  sinc_cutoff = fabs((double)csf_smp_pos_get_full(chan->increment) / 4294967296.0); \
+	if (sinc_cutoff <= 0.0) sinc_cutoff = 1.0e-25; \
+	sinc_cutoff = 1.0 / sinc_cutoff; \
+	if (sinc_cutoff > 1.0) sinc_cutoff = 1.0;
```
mrpapersonic:
```c
sinc_cutoff = MIN(sinc_cutoff, 1.0);
```

mrpapersonic (changes requested):
minor nitpicks

player/mixer.c
reales: Done. Combined into a DEFINE_SINC256_INTERP(bits) macro.

player/mixer.c
reales: Done

reales:
> Does this actually fit on the preferences page? Last I checked there wasn't any space for it.

Good catch, it didn't fit with a 5th mode. I replaced the per-mode toggle buttons with a single fixed-height listbox (like the Audio Devices/Drivers boxes), so the layout no longer depends on the number of modes. Adding new interpolation modes is now trivial, just append to the interpolation_modes[] array.

Preferences

PR #831, opened 2026-06-01
Add check to sdl2_video_refresh_rate

drradio:
Add checking for null pointer and fullscreen mode. This solves crashing after turning monitor off on my system. 
According to [SDL Wiki](https://wiki.libsdl.org/SDL2/SDL_GetWindowDisplayMode), SDL_GetWindowDisplayMode is useful only in fullscreen modes.

System specs:
- Debian GNU/Linux 13 (trixie);
- CPU: AMD Ryzen 7 5700X
- RAM: 32 GB;
- GPU: Nvidia GeForce RTX 3060;
- Display: BenQ GW3290QT on DisplayPort;
- WM: Openbox 3.6.1-12+b2 + Picom 13.

### DIFF
--- a/sys/sdl2/video.c
+++ b/sys/sdl2/video.c
@@ -883,6 +883,9 @@ static void sdl2_video_show_cursor(int enabled)
 /* negative if unknown */
 static int sdl2_video_refresh_rate(float *r)
 {
+	if (r == NULL || !video.fullscreen)
+		return -1;
+
 	/* This can get nasty if the user has the window
 	 * in two, three, four monitors */
 	SDL_DisplayMode m;


sys/sdl2/video.c
```diff
@@ -883,6 +883,9 @@ static void sdl2_video_show_cursor(int enabled)
 /* negative if unknown */
 static int sdl2_video_refresh_rate(float *r)
 {
+	if (r == NULL || !video.fullscreen)
```
vcaputo, line 886: the `r == NULL` is unnecessary, if we want to be defensive about callers passing NULL for this it should be in the schism/video.c video_refresh_rate() wrapper.

did you identify a call path that was passing NULL for r?

mrpapersonic:
> According to [SDL Wiki](https://wiki.libsdl.org/SDL2/SDL_GetWindowDisplayMode), SDL_GetWindowDisplayMode is useful only in fullscreen modes.

It never says anything about the window needing to be fullscreen already:

> Query the display mode to use when a window is visible at fullscreen.

In fact the wording implies that you *should* use this when the window is *not* fullscreen to know which display mode to use upon switching to fullscreen. (though admittedly it is worded pretty poorly)

If this function is what's crashing when you turn off your monitor, it's a bug in sdl2-compat or SDL3.

sys/sdl2/video.c
drradio: Nope, I added null-check only as safeguard. The essential logic is check for fullscreen mode.

drradio:
Ok, that's fast dirty hack. But do we need some workaround here?

PR #832, opened 2026-06-29
Add GStreamer sample loading support

logiclrd:
The core Schism Tracker supports a specific set of file formats that does not include the more sophisticated compressed formats such as MP3, AC3, Opus, etc. On Windows, MP3 files can still be loaded as samples, because Schism Tracker uses Media Foundation for generic media format support. On OS X, MP3 files can be loaded because Schism Tracker uses the CoreAudio API. However, on Linux, if you select an MP3 file, or any other unsupported format, Schism Tracker simply loads the bytes as raw audio.

This PR plugs this gap, adding a layer that uses the GStreamer library on Linux to the same end.

- `fmt_gstreamer_read_info` uses the GStreamer Discoverer component to probe flies and determine what kind of media they contain, if it is supported by GStreamer.

- `fmt_gstreamer_load_sample` sets up a pipeline that decodes the raw audio data, coerces it to s16le format, and then copies the samples  received from GStreamer to the sample buffer, translating them to u8 if the underlying audio data uses 8 bits (or fewer) per sample.

- The use of the GStreamer follows the same pattern as the Media Foundation code, allowing it to be statically-linked or probed at runtime with `dlopen` et al. By default, it is dynamically loaded/resolved, but if `GSTREAMER_LINKING` is `yes` (or `check`, since Linux has `loadso`), then it is statically-linked instead.

    - One minor complication with this is that the GStreamer header files supply `static inline` functions that redirect a handful of `_unref` functions to a common `gst_mini_object_unref`, which means that including the headers will generate a statically-linked dependency on `gst_mini_object_unref`. `gstreamer.c` gets around that by redirecting `gst_mini_object_unref` to a trampoline function using the preprocessor, and then the trampoline calls the real `gst_mini_object_unref` using the appropriate strategy for the build.

https://github.com/user-attachments/assets/fc08b54c-48db-449a-944f-5a8097c6efac

### DIFF
--- a/fmt/gstreamer.c
+++ b/fmt/gstreamer.c
@@ -344,11 +344,7 @@ int fmt_gstreamer_read_info(dmoz_file_t *file, slurp_t *fp)
 							file->smp_speed = sample_rate;
 							file->smp_length = duration * sample_rate / GST_SECOND;
 
-							if (!display_name) {
-								display_name = file->base;
-							}
-
-							file->title = str_dup(display_name);
+							file->title = str_dup(display_name ? display_name : file->base);
 							file->description = description;
 
 							success = 1;
--- a/fmt/gstreamer.c
+++ b/fmt/gstreamer.c
@@ -217,6 +217,10 @@ static gst_buffer_unmap_spec GST_DYNAMIC_gst_buffer_unmap = NULL;
 	? GST_DYNAMIC_g_free_sized(mem, __builtin_object_size((mem), 0)) \
 	: GST_DYNAMIC_g_free(mem)
 
+#else
+
+#define g_free GST_DYNAMIC_g_free
+
 #endif /* G_GNUC_CHECK_VERSION (4, 1) && GLIB_VERSION_MAX_ALLOWED >= GLIB_VERSION_2_78 && defined(G_HAVE_FREE_SIZED) */
 
 #endif /* GSTREAMER_DYNAMIC_LOAD && !LINK_TO_GSTREAMER */
--- a/fmt/gstreamer.c
+++ b/fmt/gstreamer.c
@@ -1,0 +1,859 @@
+/*
+ * Schism Tracker - a cross-platform Impulse Tracker clone
+ * copyright (c) 2003-2005 Storlek <storlek@rigelseven.com>
+ * copyright (c) 2005-2008 Mrs. Brisby <mrs.brisby@nimh.org>
+ * copyright (c) 2009 Storlek & Mrs. Brisby
+ * copyright (c) 2010-2012 Storlek
+ * URL: http://schismtracker.org/
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include "headers.h"
+
+#include "loadso.h"
+#include "dmoz.h"
+#include "mem.h"
+
+/* The GStreamer header files emit static inline function definitions that call gst_mini_object_unref. If we're
+ * trying to load everything dynamically, this defeats the straightforward attempt, because the call occurs
+ * before we have a chance to use macros to redirect. We can't just define all our trampoline functions
+ * before including the headers, because none of the types have been declared yet.
+ *
+ * So, we redirect to a trampoline that will then call either the statically-linked gst_mini_object_unref or
+ * call through the dynamically-resolved function pointer.
+ */
+struct _GstMiniObject;
+
+typedef struct _GstMiniObject GstMiniObject;
+
+static void GST_TRAMPOLINE_gst_mini_object_unref(struct _GstMiniObject *mini_object);
+
+#define gst_mini_object_unref GST_TRAMPOLINE_gst_mini_object_unref
+
+#include <gst/gst.h>
+#include <gst/pbutils/gstdiscoverer.h>
+#include <gst/pbutils/pbutils.h>
+#include <gst/app/gstappsrc.h>
+#include <gst/audio/audio.h>
+
+#undef gst_mini_object_unref
+
+#if GSTREAMER_DYNAMIC_LOAD
+
+/* ----------------------------------------------------------- */
+
+typedef gpointer (*g_malloc_spec)(gsize n_bytes);
+typedef void (*g_free_spec)(gpointer mem);
+typedef void (*g_free_sized_spec)(gpointer mem, size_t size);
+typedef const gchar *(*g_intern_string_spec)(const gchar *string);
+typedef gboolean (*g_strv_contains_spec)(const gchar *const *strv, const gchar *str);
+typedef void (*g_object_set_spec)(gpointer object, const gchar *first_property_name, ...);
+typedef void (*g_object_unref_spec)(gpointer object);
+typedef gulong (*g_signal_connect_data_spec)(gpointer instance, const gchar *detailed_signal, GCallback c_handler, gpointer data, GClosureNotify destroy_data, GConnectFlags connect_flags);
+typedef void (*g_signal_emit_by_name_spec)(gpointer instance, const gchar *detailed_signal, ...);
+typedef gboolean (*gst_init_check_spec)(int *argc, char **argv[], GError **error);
+typedef void (*gst_object_unref_spec)(gpointer object);
+typedef void (*gst_mini_object_unref_spec)(GstMiniObject *mini_object);
+typedef GstStructure *(*gst_structure_new_spec)(const gchar *name, const gchar *firstfield, ...);
+typedef const gchar *(*gst_structure_get_name_spec)(const GstStructure *structure);
+typedef gchar *(*gst_filename_to_uri_spec)(const gchar *filename, GError **error);
+typedef GstDiscoverer *(*gst_discoverer_new_spec)(GstClockTime timeout, GError **err);
+typedef GstDiscovererInfo *(*gst_discoverer_discover_uri_spec)(GstDiscoverer *discoverer, const gchar *uri, GError **err);
+typedef GstDiscovererResult (*gst_discoverer_info_get_result_spec)(const GstDiscovererInfo *info);
+typedef GstClockTime (*gst_discoverer_info_get_duration_spec)(const GstDiscovererInfo *info);
+typedef GList *(*gst_discoverer_info_get_audio_streams_spec)(GstDiscovererInfo *info);
+typedef gchar *(*gst_discoverer_stream_info_get_stream_type_nick_spec)(GstDiscovererStreamInfo *info);
+typedef GstCaps *(*gst_discoverer_stream_info_get_caps_spec)(GstDiscovererStreamInfo *info);
+typedef void (*gst_discoverer_stream_info_list_free_spec)(GList *infos);
+typedef guint (*gst_discoverer_audio_info_get_sample_rate_spec)(const GstDiscovererAudioInfo *info);
+typedef guint (*gst_discoverer_audio_info_get_channels_spec)(const GstDiscovererAudioInfo *info);
+typedef guint (*gst_discoverer_audio_info_get_depth_spec)(const GstDiscovererAudioInfo *info);
+typedef gchar *(*gst_pb_utils_get_decoder_description_spec)(const GstCaps *caps);
+typedef GstElement *(*gst_pipeline_new_spec)(const gchar *name);
+typedef GstElement *(*gst_element_factory_make_spec)(const gchar *factoryname, const gchar *name);
+typedef GstStateChangeReturn (*gst_element_set_state_spec)(GstElement *element, GstState state);
+typedef GstPad *(*gst_element_get_static_pad_spec)(GstElement *element, const gchar *name);
+typedef gboolean (*gst_element_link_spec)(GstElement *src, GstElement *dest);
+typedef GstBus *(*gst_element_get_bus_spec)(GstElement *element);
+typedef GstMessage *(*gst_bus_timed_pop_filtered_spec)(GstBus *bus, GstClockTime timeout, GstMessageType types);
+typedef void (*gst_bin_add_many_spec)(GstBin *bin, GstElement *element_1, ...);
+typedef GstCaps *(*gst_pad_get_current_caps_spec)(GstPad *pad);
+typedef gboolean (*gst_pad_is_linked_spec)(GstPad *pad);
+typedef GstPadLinkReturn (*gst_pad_link_spec)(GstPad *srcpad, GstPad *sinkpad);
+typedef GstCaps *(*gst_caps_new_empty_spec)(void);
+typedef GstCaps *(*gst_caps_new_empty_simple_spec)(const char *media_type);
+typedef guint (*gst_caps_get_size_spec)(const GstCaps *caps);
+typedef GstStructure *(*gst_caps_get_structure_spec)(const GstCaps *caps, guint index);
+typedef void (*gst_caps_append_structure_spec)(GstCaps *caps, GstStructure  *structure);
+typedef gboolean (*gst_caps_can_intersect_spec)(const GstCaps *caps1, const GstCaps *caps2);
+typedef gboolean (*gst_audio_info_from_caps_spec)(GstAudioInfo *info, const GstCaps *caps);
+typedef GstBuffer *(*gst_sample_get_buffer_spec)(GstSample *sample);
+typedef GstCaps *(*gst_sample_get_caps_spec)(GstSample *sample);
+typedef GstBuffer *(*gst_buffer_new_wrapped_spec)(gpointer data, gsize size);
+typedef gboolean (*gst_buffer_map_spec)(GstBuffer *buffer, GstMapInfo *info, GstMapFlags flags);
+typedef void (*gst_buffer_unmap_spec)(GstBuffer *buffer, GstMapInfo *info);
+
+static g_malloc_spec GST_DYNAMIC_g_malloc = NULL;
+static g_free_spec GST_DYNAMIC_g_free = NULL;
+static g_free_sized_spec GST_DYNAMIC_g_free_sized = NULL;
+static g_intern_string_spec GST_DYNAMIC_g_intern_string = NULL;
+static g_strv_contains_spec GST_DYNAMIC_g_strv_contains = NULL;
+static g_object_set_spec GST_DYNAMIC_g_object_set = NULL;
+static g_object_unref_spec GST_DYNAMIC_g_object_unref = NULL;
+static g_signal_connect_data_spec GST_DYNAMIC_g_signal_connect_data = NULL;
+static g_signal_emit_by_name_spec GST_DYNAMIC_g_signal_emit_by_name = NULL;
+static gst_init_check_spec GST_DYNAMIC_gst_init_check = NULL;
+static gst_object_unref_spec GST_DYNAMIC_gst_object_unref = NULL;
+static gst_mini_object_unref_spec GST_DYNAMIC_gst_mini_object_unref = NULL;
+static gst_structure_new_spec GST_DYNAMIC_gst_structure_new = NULL;
+static gst_structure_get_name_spec GST_DYNAMIC_gst_structure_get_name = NULL;
+static gst_filename_to_uri_spec GST_DYNAMIC_gst_filename_to_uri = NULL;
+static gst_discoverer_new_spec GST_DYNAMIC_gst_discoverer_new = NULL;
+static gst_discoverer_discover_uri_spec GST_DYNAMIC_gst_discoverer_discover_uri = NULL;
+static gst_discoverer_info_get_result_spec GST_DYNAMIC_gst_discoverer_info_get_result = NULL;
+static gst_discoverer_info_get_duration_spec GST_DYNAMIC_gst_discoverer_info_get_duration = NULL;
+static gst_discoverer_info_get_audio_streams_spec GST_DYNAMIC_gst_discoverer_info_get_audio_streams = NULL;
+static gst_discoverer_stream_info_get_stream_type_nick_spec GST_DYNAMIC_gst_discoverer_stream_info_get_stream_type_nick = NULL;
+static gst_discoverer_stream_info_get_caps_spec GST_DYNAMIC_gst_discoverer_stream_info_get_caps = NULL;
+static gst_discoverer_stream_info_list_free_spec GST_DYNAMIC_gst_discoverer_stream_info_list_free = NULL;
+static gst_discoverer_audio_info_get_sample_rate_spec GST_DYNAMIC_gst_discoverer_audio_info_get_sample_rate = NULL;
+static gst_discoverer_audio_info_get_channels_spec GST_DYNAMIC_gst_discoverer_audio_info_get_channels = NULL;
+static gst_discoverer_audio_info_get_depth_spec GST_DYNAMIC_gst_discoverer_audio_info_get_depth = NULL;
+static gst_pb_utils_get_decoder_description_spec GST_DYNAMIC_gst_pb_utils_get_decoder_description = NULL;
+static gst_pipeline_new_spec GST_DYNAMIC_gst_pipeline_new = NULL;
+static gst_element_factory_make_spec GST_DYNAMIC_gst_element_factory_make = NULL;
+static gst_element_set_state_spec GST_DYNAMIC_gst_element_set_state = NULL;
+static gst_element_get_static_pad_spec GST_DYNAMIC_gst_element_get_static_pad = NULL;
+static gst_element_link_spec GST_DYNAMIC_gst_element_link = NULL;
+static gst_element_get_bus_spec GST_DYNAMIC_gst_element_get_bus = NULL;
+static gst_bus_timed_pop_filtered_spec GST_DYNAMIC_gst_bus_timed_pop_filtered = NULL;
+static gst_bin_add_many_spec GST_DYNAMIC_gst_bin_add_many = NULL;
+static gst_pad_get_current_caps_spec GST_DYNAMIC_gst_pad_get_current_caps = NULL;
+static gst_pad_is_linked_spec GST_DYNAMIC_gst_pad_is_linked = NULL;
+static gst_pad_link_spec GST_DYNAMIC_gst_pad_link = NULL;
+static gst_caps_new_empty_spec GST_DYNAMIC_gst_caps_new_empty = NULL;
+static gst_caps_new_empty_simple_spec GST_DYNAMIC_gst_caps_new_empty_simple = NULL;
+static gst_caps_get_size_spec GST_DYNAMIC_gst_caps_get_size = NULL;
+static gst_caps_get_structure_spec GST_DYNAMIC_gst_caps_get_structure = NULL;
+static gst_caps_append_structure_spec GST_DYNAMIC_gst_caps_append_structure = NULL;
+static gst_caps_can_intersect_spec GST_DYNAMIC_gst_caps_can_intersect = NULL;
+static gst_audio_info_from_caps_spec GST_DYNAMIC_gst_audio_info_from_caps = NULL;
+static gst_sample_get_buffer_spec GST_DYNAMIC_gst_sample_get_buffer = NULL;
+static gst_sample_get_caps_spec GST_DYNAMIC_gst_sample_get_caps = NULL;
+static gst_buffer_new_wrapped_spec GST_DYNAMIC_gst_buffer_new_wrapped = NULL;
+static gst_buffer_map_spec GST_DYNAMIC_gst_buffer_map = NULL;
+static gst_buffer_unmap_spec GST_DYNAMIC_gst_buffer_unmap = NULL;
+
+#define g_malloc GST_DYNAMIC_g_malloc
+#define g_object_set GST_DYNAMIC_g_object_set
+#define g_object_unref GST_DYNAMIC_g_object_unref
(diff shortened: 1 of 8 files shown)


logiclrd:
```
checking for gstreamer-1.0 gstreamer-pbutils-1.0 gstreamer-audio-1.0... yes
```

...on X-Box?? That's gotta be a false positive, right?

logiclrd:
I notice the Ubuntu build agents aren't finding GStreamer. The clanker tells me that a GitHub action `blinemedical/setup-gstreamer` is the ideal way to get GStreamer installed.

logiclrd:
The clanker is not smart. :-) Let's try this...

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+	discoverer = gst_discoverer_new(5 * GST_SECOND, NULL);
+
+	if (!discoverer)
+		return 0;
+
+	/* Translate path to URI */
+	uri = gst_filename_to_uri(file->path, NULL);
+
+	if (!uri)
+	{
```
mrpapersonic: Please use
```c
if (!uri) {
        ...
}
```
The basic style for Schism is to never put a brace on a newline *unless* it's for a function definition (or if it's a really large if statement where it makes sense).

For example:
```c
struct x {
        int a;
        int b;
};

int z(struct x *q)
{
        if (q->a && q->b) return 5;

        if (q->a) {
                return 4;
        } else if (q->b) {
                return 3;
        } else {
                return 1;
        }
}
```

The [Linux kernel style guide](https://docs.kernel.org/process/coding-style.html) is pretty close to what Schism does. Schism does a few things differently, but they're not that important.

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+
+							file->smp_flags = 0;
+
+							if (channels > 1)
+								file->smp_flags |= CHN_STEREO;
+							if (bits > 8)
+								file->smp_flags |= CHN_16BIT;
+
+							file->smp_speed = sample_rate;
+							file->smp_length = (duration / (double)GST_SECOND) * sample_rate;
```
mrpapersonic: UGH! Annoyingly, GStreamer made the (incorrect) assumption that all streams in a container are the same length.

...anyway, this could be changed to
```c
file->smp_length = duration * sample_rate / GST_SECOND;
```
which keeps everything in integer-land rather than floating-point, maybe with some checks for overflow (although highly unlikely).

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+}
+
+/* ---------------------------------------------------------------- */
+
+#define BUFFER_SIZE 4096
+
+static void source__need_data(GstElement *source, guint unused_size, gpointer data)
+{
+	gpointer buffer;
+	int num_read;
```
mrpapersonic: use `size_t` here (I believe `gsize` is just a synonym for `size_t` but I may be wrong)

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+	GstFlowReturn ret;
+
+	slurp_t *fp = (slurp_t *)data;
+
+	buffer = g_malloc(BUFFER_SIZE);
+
+	num_read = slurp_read(fp, buffer, BUFFER_SIZE);
+
+	if (num_read <= 0)
+		g_signal_emit_by_name(source, "end-of-stream", &ret);
```
mrpapersonic, line 400: please add braces to the `if` statement here

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+int fmt_gstreamer_read_info(dmoz_file_t *file, slurp_t *fp)
+{
+	GstDiscoverer *discoverer;
+	GstDiscovererInfo *info;
+	GstDiscovererResult result;
+	gchar *uri;
+	int success = 0;
+
+	/* Initialize GStreamer -- do not call gst_deinit because it isn't subsequently possible to reinitialize in the same process */
+	if (!gstreamer_init())
```
mrpapersonic: This stuff should be initialized at startup like everything else, see `fmt/avformat.c`, `fmt/zlib.c`, etc.

Though tbh that stuff ought to be in some `fmt_init()` function somewhere...

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+}
+
+static int gstreamer_load(void)
+{
+	int success = 1;
+
+#define LOAD_LIBRARY(lib, filename)     \
+	do if (success)                       \
+	{                                     \
+		lib = loadso_object_load(filename); \
```
mrpapersonic: Use the `library_load` function for this:
```c
#define LOAD_LIBRARY(lib, name) \
    do { \
         if (!success) break; \
    \
         lib = library_load(name, 0, 0); \
         if (!lib) \
             success = 0; \
    } while (0)

LOAD_LIBRARY(lib_glib, "glib-2.0");
LOAD_LIBRARY(lib_gobject, "gobject-2.0");
LOAD_LIBRARY(lib_gstreamer, "gstreamer-1.0");
LOAD_LIBRARY(lib_gstpbutils, "gstpbutils-1.0");
LOAD_LIBRARY(lib_gstaudio, "gstaudio-1.0");
```
You should also use `goto` for cleanup instead of repeated `if (success)` checks.

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+/* ----------------------------------------------------------- */
+
+static int gstreamer_initialized = 0;
+
+static int gstreamer_init(void)
+{
+	if (gstreamer_initialized)
+		return 1;
+
+	#if GSTREAMER_DYNAMIC_LOAD
```
mrpapersonic: please don't indent preprocessor directives

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+#include <gst/app/gstappsrc.h>
+#include <gst/audio/audio.h>
+
+#undef gst_mini_object_unref
+
+#if GSTREAMER_DYNAMIC_LOAD
+
+/* ----------------------------------------------------------- */
+
+typedef gpointer (*g_malloc_spec)(gsize n_bytes);
```
mrpapersonic, line 61: Can these all be collapsed into a .h file? (this is something I've wanted to do for all the other dynload stuff too)

e.g.
```c
#ifdef GSTREAMER_DYNAMIC_LOAD
/* define library globals */
# define LIB(var, name) static void *lib_##var;
#define FUNC(lib, ret, name, params) static ret (*GST_##name) params;
# include "gstreamer-funcs.h"

static int gstreamer_init(void)
{
        /* TODO goto cleanup when any of these fail */
# define LIB(var, name) lib_##var = library_load(name, 0, 0);
# define FUNC(lib, ret, name, params) GST_##name = (ret (*) params)loadso_function_load(lib_##var, #name);
# include "gstreamer-funcs.h"

        return 0;

cleanup:
        /* TODO unload stuff here; could set everything to NULL beforehand and check here? */
        return -1;
}

static void gstreamer_quit(void)
{
# define LIB(var, name) loadso_object_unload(lib_##var); lib_##var = NULL;
# define FUNC(lib, ret, name, params) GST_##name = NULL;
# include "gstreamer-funcs.h"
}
#else
/* could also define static inline functions that just call the gst_ functions */
# define FUNC(lib, ret, name, params) static ret (*const GST_##name) params = name;

static int gstreamer_init(void)
{
        return 0;
}

static void gstreamer_quit(void)
{
}
#endif
```
As hinted above, I'd also prefer if *our* code never touches anything `gst_*` more than we have to, to prevent namespace collisions. That is, prefix all of these with `GST_`.

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+#define gst_sample_get_caps GST_DYNAMIC_gst_sample_get_caps
+#define gst_buffer_new_wrapped GST_DYNAMIC_gst_buffer_new_wrapped
+#define gst_buffer_map GST_DYNAMIC_gst_buffer_map
+#define gst_buffer_unmap GST_DYNAMIC_gst_buffer_unmap
+
+/* g_free is a macro that calls g_free or g_free_sized */
+
+#undef g_free
+#define g_free(mem)                                                \
+  (__builtin_object_size((mem), 0) != ((size_t)-1))                \
```
mrpapersonic, line 221: pls add outer parenthesis and a `SCHISM_GNUC_HAS_BUILTIN` check

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+
+							if (channels > 1)
+								file->smp_flags |= CHN_STEREO;
+							if (bits > 8)
+								file->smp_flags |= CHN_16BIT;
+
+							file->smp_speed = sample_rate;
+							file->smp_length = (duration / (double)GST_SECOND) * sample_rate;
+
+							if (!display_name)
```
mrpapersonic: just do
```c
file->title = str_dup(display_name ? display_name : file->base);
```

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+#define BUFFER_SIZE 4096
+
+static void source__need_data(GstElement *source, guint unused_size, gpointer data)
+{
+	gpointer buffer;
+	int num_read;
+	GstFlowReturn ret;
+
+	slurp_t *fp = (slurp_t *)data;
+
```
mrpapersonic, line 394: You can change everything to use `unused_size`, and use `BUFFER_SIZE` as a minimum:
```c
/* https://gstreamer.freedesktop.org/documentation/app/appsrc.html#appsrc::need-data 
 * "length is just a hint and when it is set to -1, any number of bytes can be pushed into appsrc." */
unused_size = (unused_size != (guint)-1) ? unused_size : BUFFER_SIZE;

buffer = g_malloc(unused_size);
```

Since slurp is always seekable, maybe we should implement `seek-data` as well?

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+		GstMapInfo map;
+
+		if (gst_buffer_map(buffer, &map, GST_MAP_READ))
+		{
+			sample_buffer_chain_t *buffer_chain = (sample_buffer_chain_t *)data;
+
+			sample_buffer_chain_node_t *new_node = (sample_buffer_chain_node_t *)malloc(sizeof(sample_buffer_chain_node_t));
+
+			new_node->next = NULL;
+			new_node->sample_count = map.size / 2;
```
mrpapersonic: What's this magic divide by two?

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+			new_node->sample_data = malloc(new_node->sample_count * buffer_chain->sample_bits / 8);
+
+			if (!new_node->sample_data)
+				free(new_node);
+			else
+			{
+				if (buffer_chain->sample_bits == 16)
+					memcpy(new_node->sample_data, map.data, new_node->sample_count * buffer_chain->sample_bits / 8);
```
mrpapersonic: This will break under big endian.

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+				else
+				{
+					/* Convert from S16 to U8 */
```
mrpapersonic: Could we just make it output unsigned 8-bit if `sample_bits` is 8?

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+ * trying to load everything dynamically, this defeats the straightforward attempt, because the call occurs
+ * before we have a chance to use macros to redirect. We can't just define all our trampoline functions
+ * before including the headers, because none of the types have been declared yet.
+ *
+ * So, we redirect to a trampoline that will then call either the statically-linked gst_mini_object_unref or
+ * call through the dynamically-resolved function pointer.
+ */
+struct _GstMiniObject;
+
+typedef struct _GstMiniObject GstMiniObject;
```
mrpapersonic: this typedef is unnecessary and will break pre-C11 compilers; duplicate typedefs weren't allowed until C11

fmt/gstreamer.c
```diff
@@ -0,0 +1,859 @@
...
+typedef struct sample_buffer_chain sample_buffer_chain_t;
+typedef struct sample_buffer_chain_node sample_buffer_chain_node_t;
+
+struct pipeline_data
+{
+	GstElement *convert;
+	sample_buffer_chain_t *buffer_chain;
+};
+
+struct sample_buffer_chain
```
mrpapersonic: is there any reason in specific we can't use a disko membuf?

This feels like a lot of extra complexity for no real gain.

mrpapersonic (changes requested):

fmt/gstreamer.c
logiclrd: Oh, right, I've been deeply immersed in other codebases and forgot. 😅

fmt/gstreamer.c
logiclrd: Mm, that makes sense. What is the call site, though? The Media Foundation support was initially the template I was working from, and its call to `win32mf_init()` is in `sys/win32/osdefs.c`, but unlike MF, GStreamer is not itself OS-specific. The implementation here is implicitly targeting Linux specifically at this point, but there isn't a `sys/linux/osdefs.c`. You mention `avformat` and `zlib`, and I see calls to their initialization directly from `schism_main` so I guess I'll toss it in there.

mrpapersonic: Yep, i meant in `schism_main`.

fmt/gstreamer.c
logiclrd: Mm, okay, fair enough. I wrote it this way because I just intuitively wanted to avoid any possibility of integer overflow, but I ran the numbers and at 48 kHz, a sample would have to be over 12 hours long for this to be an issue. 😁

fmt/gstreamer.c
logiclrd: Mm, okay. The code feels cleaner this way, but the `goto` paradigm is already in use in other places, and consistency is even more valuable. 🙂 

The use of `loadso_object_load` was copied from `win32mf.c`. I'll switch it over.

fmt/gstreamer.c
mrpapersonic: It would have to be a little bit over 2 days. Nevertheless, a 64-bit overflow checker is simple to write (albeit going through some hoops):
```c
static inline uint64_t u64_mul_overflow(uint64_t x, uint64_t y, int *poverflow)
{
        uint64_t z = x * y;
        *poverflow = (z / x != y);
        return z;
}

static inline int64_t i64_mul_overflow(int64_t x, int64_t y, int *poverflow)
{
        /* doing this is necessary, as signed integer overflow is undefined behavior in C
         * fortunately for us, unsigned overflow just wraps around,
         * so we can (ab)use that behavior to get something that works. */
        uint64_t z = u64_mul_overflow(*(uint64_t *)&x, *(uint64_t *)&y, poverflow);
        return *(int64_t *)&z;
}
```
GCC also has the `__builtin_mul_overflow` function, but it optimizes this code down to the same thing anyway.

Something like this ought to be shoved in `bits.h`, too, so it's not duplicated everywhere...

fmt/gstreamer.c
logiclrd: I didn't think there was anything with a `gst_` prefix in this codebase. Do you mean the `gstreamer_` things? Because to the best of my knowledge, no part of GStreamer ever actually spells out `gstreamer`, it's very consistently a `gst_` prefix.

The other change you've described here feels like a refactoring to allow a common implementation, which means that it would be best done to e.g. the MF code at the same time. As such, it feels like it would be better as a separate refactoring PR.

fmt/gstreamer.c
logiclrd: This is copy/pasted directly from the GLib header files verbatim, the only change being the addition of the `GST_DYNAMIC_` prefix so that it calls through the function pointer when loading dynamically. If you can `#include <glib.h>`, then you can do this macro. Do we want to define `g_free` differently than GLib does??

fmt/gstreamer.c
logiclrd: I'm not sure if `seek-data` is ever called by a pipeline that only does a linear read. The guides I read didn't bother to implement it, but maybe they were just being lazy and there are some formats where it would fail?

The clanker thinks that if nothing downstream ever seeks the stream, then the pipeline won't ever generate `seek-data`. The clanker has been known to be wrong, though...

logiclrd: I'm also not entirely sure if it's actually better to only push `length` bytes, when `length` is not -1. If you push more than `length` bytes, then it simply buffers the additional data you send, and that can potentially reduce the frequency with which it has to fire the signal in the first place. From what I've just read, it might even be preferable to bump `BUFFER_SIZE` to a significantly larger value, greatly reducing the number of times the signal has to be emitted. Then it just becomes a balancing act of how much memory is comfortable to allocate in chunks like this. I mean, if `BUFFER_SIZE` were ridiculous, like 8MB or something, then the signal would in almost all cases only fire once. But, the majority of samples would only use a fraction of that buffer size. My gut thinks that 128KB or 256KB wouldn't be terrible values.

fmt/gstreamer.c
logiclrd: Commented. :-)

fmt/gstreamer.c
logiclrd: Ah, the native format is S16LE on little-endian and S16BE on big-endian platforms?

fmt/gstreamer.c
logiclrd: It is possible but it's more complexity at the boundary. I chose this architecture because it compartmentalizes the complexity. It does mean that the intermediate buffers are twice as large as they need to be for 8-bit samples, but they're short-lived.

fmt/gstreamer.c
mrpapersonic: I meant all of those `#define`s to change the `gst_` functions to pointers.

That aside, this can wait for now until I put together something functional, then I'll worry about getting everything together.

fmt/gstreamer.c
mrpapersonic: 256KiB is probably fine. We could also allocate that buffer as part of setting up the callback, so it doesn't perform a malloc, although the benefits would probably be moot compared to the cost of doing a read from disk.

fmt/gstreamer.c
logiclrd: Ohh, so, instead of redirecting the real `gst_` functions, have our own independent names and then define those to either the real `gst_` functions or function pointers?

fmt/gstreamer.c
mrpapersonic: glib also uses a GCC version check:

https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gmem.h?ref_type=heads#L169

Just separate it with
```c
#undef g_free
#if SCHISM_GNUC_HAS_BUILTIN(__builtin_object_size, 4, 1, 0)
# define g_free(mem) /* ... */
#endif
```

fmt/gstreamer.c
logiclrd: Well, there's a problem here, because GStreamer wasn't designed to be dynamically loaded, and, as an implementation detail, it defines a number of the `_unref` functions as inline functions that call `gst_mini_object_unref`. This means that when you are including the GStreamer header files, before any of your own code, it's _already_ emitting hard references to GStreamer symbols.

One that we need to call, for example:

```
static inline void
gst_caps_unref (GstCaps * caps)
{
  gst_mini_object_unref (GST_MINI_OBJECT_CAST (caps));
}
```

We _could_ just directly call `gst_mini_object_unref` instead of `gst_caps_unref`, but that's technically an internal implementation detail.

For a first attempt, you might try defining a trampoline function that just takes `void *`, because any `GstMiniObject *` will of course just implicitly cast down to `void *` -- but it _also_ has, in the header file, calls that pass `gst_mini_object_unref` into other functions as a function pointer. These calls will then fail, because the function pointer _does_ have to take type `GstMiniObject *`. You can't pass a `void(*)(void *)` into a `void(*)(GstMiniObject *)` parameter.

Perhaps we could add a check to `configure` so that if the type redefinition _is_ an error, then that's a platform where we simply don't (can't) support the GStreamer support?

I think one thing that _can_ be done is to only define the trampoline when dynamically-linking. If statically-linking, then it's fine to just let the header's inline functions directly call `gst_mini_object_unref`. Then, even if a platform for whatever reason _can_ use GStreamer but _can't_ handle the type redeclaration, it can still use GStreamer through linking statically.

fmt/gstreamer.c
mrpapersonic: Yep, check `sys/alsa/midi-alsa.c` for an example of what I mean. It also has inline functions that call real functions too, so you have a basis on what to do there too :)

fmt/gstreamer.c
mrpapersonic: I'm just saying that the typedef isn't necessary. The function definition already uses the `struct _GstMiniObject *` which is perfectly fine in C99.

The problem is just with the typedef, which I don't see anything obvious breaking if it's removed.

fmt/gstreamer.c
logiclrd: Mm, indeed, especially considering that a given run might not actually _use_ the GStreamer loader at all.

fmt/gstreamer.c
logiclrd: This came about because a) I initially wrote the code in a stand-alone project to prove the concept, and b) I don't have an existing familiarity with disko membuf 🙂 I will look into it but suspect it will translate pretty straightforwardly.

fmt/gstreamer.c
logiclrd: Does it? I think that's a GLib version check, no?

If the `#if` test fails, then I guess don't bother with `g_free_sized` and just always call `g_free`?

I did a bit of reading, and from what I read, GLib uses allocation pools for different small buffer sizes. For the sizes N that have dedicated pools, `g_malloc` allocates from the pools, and then `g_free` needs to figure out which pool the pointer is in. But, if the compiler knows that the pointer's target has `sizeof == N`, then it can go directly to the pool in question. From what I read, it's always safe, just potentially slightly less efficient, to call the actual underlying `g_free` for any pointer instead of directly forwarding to `g_free_sized` for the correct pool size.

So I guess, yeah, that `#if` has an `#else` that doesn't try to be fancy and just calls `g_free`.

But the version guard in the header file implies that if you're compiling for a GLib version that supports `g_free_sized`, then you are _by definition_ using a GCC version that supports the intrinsic. But maybe this code supports older GLib versions? I have no idea what minimum versions the GStreamer bits I've cobbled together have for the underlying libraries. GLib 2.76 was released in March 2023. I don't know what the matrix of Schism Tracker version support looks like but this project probably intends to build and run on Linux versions that came out before then, I guess? 🙂

Worst comes to worst, I guess they can always just `./configure --without-gstreamer` to stay unblocked at least...

mrpapersonic: `G_GNUC_CHECK_VERSION` just does a version check on `__GNUC__`, which is the GCC version. The builtin was added in GCC 4.1.

logiclrd: Oh, interesting, the anchor didn't work when I first clicked your link, so I was looking at the wrong line. I just clicked on it again, because I didn't remember the word `GNUC`, and this time it properly took me to line 169. Right-o :-) Yes, absolutely makes sense to be duplicating the same check so that the `g_free` macro is only redefined if it was defined in the first place.

I'm of the mind that copy/pasting their syntax is actually preferred -- keeping it as close to the original as possible. Specifically because of what it is. Anywhere else, the local paradigm is of course `SCHISM_GNUC_HAS_BUILTIN`, but this kind-of-sort-of isn't really Schism source code per se.

fmt/gstreamer.c
logiclrd: Oh, wait, when I calculated the 12 hours, I took 2,147,483,647 and divided it by 48000. Somehow, I divided it by 2 again, not sure why, but anyway that's how I ended up at 44,739 seconds. But I didn't factor in `GST_SECOND`. So this is actually the maximum possible raw value of `duration`, not the maximum number of _seconds_. But I was operating under the misapprehension that `duration` was 32-bit. If it were, it would in fact _immediately_ be a problem for any sample longer than 2 samples, as `GST_SECOND`, I've just learned, is in fact 1,000,000,000. 😛 

So, yes, as a 64-bit value, the maximum number of seconds is 2**63 / (sample_rate * GST_SECOND), or 192,152.

Such a sample would require over 8.5 GB of RAM, 17 GB if stereo, and that's for 8-bit samples. 😂 And, the current loading strategy would have to buffer the entire thing first, before consolidating it into contiguous buffer.

logiclrd: Another option would be to add a 64-bit `muldiv` function, something like this perhaps?

```
#include <stdint.h>

uint64_t muldiv64_unsigned(uint64_t a, uint64_t b, uint64_t c)
{
  if (c == 0) {
    return 0; // TODO: Some other way to handle division by zero?
  }

  // Split 'a' into 32-bit halves
  uint64_t a_high = a >> 32;
  uint64_t a_low = a & 0xFFFFFFFF;

  // Split 'b' into 32-bit halves
  uint64_t b_high = b >> 32;
  uint64_t b_low = b & 0xFFFFFFFF;

  // Cross-multiplication to form the 128-bit product (prod_high : prod_low)
  uint64_t p0 = a_low * b_low;
  uint64_t p1 = a_low * b_high;
  uint64_t p2 = a_high * b_low;
  uint64_t p3 = a_high * b_high;

  // Combine intermediate products
  uint64_t p0_high = p0 >> 32;

  uint64_t mid = p1 + p2;
  uint64_t mid_carry = (mid < p2);

  mid += p0_high;
  mid_carry += (mid < p0_high);

  uint64_t prod_low = (mid << 32) | (p0 & 0xFFFFFFFF);
  uint64_t prod_high = p3 + (mid_carry << 32) + (mid >> 32);

  // Perform long division of 128-bit product by 64-bit 'c'
  uint64_t rem = prod_high % c;
  uint64_t res_high = prod_high / c;

  for (int i = 0; i < 64; i++) {
    rem = (rem << 1) | ((prod_low >> (63 - i)) & 1);
    res_high = (res_high << 1) | (rem / c);
    rem %= c;
  }

  return res_high;
}
```

Then you get close to 50 more bits of range before any potential overflow 😛

fmt/gstreamer.c
logiclrd: Shouldn't I use the exact prototype the GStreamer documentation specifies??

mrpapersonic: I meant for `num_read`, but github includes a whole bunch of junk alongside it. `slurp_read` returns size_t.

fmt/gstreamer.c
logiclrd: This feels like it would fit alongside the other denizens of `bits.h`.

mrpapersonic: > This feels like it would fit alongside the other denizens of `bits.h`.

That's fine with me, maybe name it `muldivu64` or `muldiv_u64`.

It might be useful to detect native 128-bit types (such as the ones in GCC) and do
```c
return (__uint128_t)a * b / c;
```
when they are available.

logiclrd: One other thought occurs to me, in the other direction: This is not a hot path. This calculation is done exactly once across the entire import process. For the cost of a floating-point operation to be even detectable it'd have to running on a ZX Spectrum or something. There's an argument for just using the floating-point version, because it's very easy to reason about and has basically zero maintenance cost because it doesn't add any new code to the codebase.

fmt/gstreamer.c
logiclrd: Ohh. Righto. It's strictly bounded by `BUFFER_SIZE`, but should probably use the proper type anyway. :-)

fmt/gstreamer.c
logiclrd: Huh. At one point, it wasn't building without it. But whatever it was has changed and it builds fine without it now. :shrug:

fmt/gstreamer.c
logiclrd: Hmm, so, a disko membuf isn't a buffer chain, it's a single buffer which is going to be reallocated and copied over and over as the size increases.

logiclrd: Also, if I'm not mistaken, the growth is linear; if extension is needed, then it extends the array by just the number of bytes provided. So, it's going to be copied and reallocated _every single time_ there's a write past the current allocation. We can estimate the number of bytes GStreamer is going to give us, but I'm pretty sure we only know for certain when we've actually received all of the `new-sample` signals on the appsink.

mrpapersonic: disko memory buffers allocate in relation to the fibonacci sequence.

logiclrd: Ahh, yes, I did overlook that further down in the `_dw_bufcheck` function.

```
	ds->length = MAX(ds->length, newlen);

	if (ds->length >= ds->allocated) {
		size_t newsize = next_fib_size(ds->length);
		uint8_t *new = realloc(ds->data, newsize);
```

I saw lines 1, 3 and 5, and my eye skipped over 4. I feel silly. Right-o then.

fmt/gstreamer.c
mrpapersonic: I was concerned about inaccuracies, but 64-bit floating point is already pretty good, and I think the sample would have to be pretty damn big for any difference to be noticeable. (also I think the other files use doubles too lol)

fmt/gstreamer.c
mrpapersonic: Just a thought: if the gstreamer API design allows it, we could use `slurp_nonseek` in conjunction with `csf_read_sample` and not have to do anything crazy.

logiclrd: Done and tested.

fmt/gstreamer.c
logiclrd: Okeydoke, I'll leave it as-is then :-)

logiclrd:
I had somehow gotten the idea in my head that 8-bit samples are supposed to be unsigned, while 16-bit samples should be signed. (S3M files maybe??) Anyway, I realized that I hadn't effectively tested the 8-bit path, and flailed around a bit before finding a way to effectively test it -- I had to temporary break the normal WAV file load path so that when I selected a WAV file, it'd pass it off to the GStreamer code. So, that's about when I discovered I was wrong about the 8-bit data being unsigned :-P That's now fixed, and the 8-bit path has been observed working. I think that last commit also fixes an index calculation mistake and I forgot to mention it in the commit message.

Just need to check about using `load_library` instead of `loadso_object_load` and I think that's all the feedback to this point addressed. Did I miss any?

logiclrd:
Merged in upstream fix to MacOS Classic build. Hopefully green lights across the board now. 🙂

PR #833, opened 2026-06-29
Fix MacOS build issue

logiclrd:
I noticed the MacOS build failing in my other PR's C.I. builds, and when I looked into it, it appears that the primary mirror for `MPW-GM.img.bin` has gone offline. It now reports HTTP 404. But, Curl doesn't treat that as an error, and so the HTTP 404 error text gets written to `MPG-GM.img.bin` and it moves on instead of trying one of the backup mirrors.

This PR updates every invocation of `curl` in `macos.yml` to specify the `-f` flag so that any HTTP failure will translate to a non-zero exit code, which should allow the fallback logic to work.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -54,10 +54,10 @@ jobs:
         if: steps.cache-universal-interfaces.outputs.cache-hit != 'true'
         run: |
           # make sure we have plenty of mirrors so we don't get bogus build fails
-          curl "https://staticky.com/mirrors/ftp.apple.com/developer/Tool_Chest/Core_Mac_OS_Tools/MPW_etc./MPW-GM_Images/MPW-GM.img.bin" > "$RUNNER_TEMP/MPW-GM.img.bin" \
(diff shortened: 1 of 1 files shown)


logiclrd:
Well, it got past that problem anyway...

mrpapersonic:
made obsolete with https://github.com/schismtracker/schismtracker/commit/2f64fa94a6bbf37644b2f06142aa2d5125a45a14, https://github.com/schismtracker/schismtracker/commit/bd0a897a7726b24843e0e345734498e525798b33, https://github.com/schismtracker/schismtracker/commit/8562dcc0d01978decc80528f7f5e209a525b4979, https://github.com/schismtracker/schismtracker/commit/09797153e1f690073cb8dc9019f20233894013ee

There were a few other problems as a result of retro68 upgrading gcc (which are generally good things and I found [a thing I needed to patch in libflac](https://github.com/schismtracker/schismtracker/commit/8562dcc0d01978decc80528f7f5e209a525b4979) because of it). other things such as `int32_t` being `long int` rather than `int` even though they are the same size (iirc) caused weird pointer problems from e.g. flac and utf8proc assuming that `int` and `int32_t` are synonyms. Instead of patching it to do it correctly I just told gcc to not error, because there are [a lot of places where it needs patching](https://github.com/schismtracker/schismtracker/actions/runs/28645668121/job/84951552883#step:20:491) and it's in a test library we don't even touch anyway.

PR #834, opened 2026-07-01
Fix W64 sample loading when the files ends right at the end of the data chunk

logiclrd:
In testing and debugging the GStreamer support with 8-bit files, I was trying to find a file format that would supply true 8-bit data (I discovered my previous testing, while an 8-bit file on disk, erroneously reports as 16-bit data from GStreamer), and I created a W64 file and found that it wouldn't load with the built-in W64 support. I debugged it and discovered that the `w64_chunk_peek` function returns false for its `data` chunk even though the chunk is properly-formed and fully there because it contains code that tests for at least 1 byte _after_ the chunk in the file, and the file ends with the chunk.

This PR updates `w64_chunk_peek` to accept chunks that abut the end of the file. With this change, the built-in W64 handling works with my test file:

[M1F1-uint8-AFsp.w64.zip](https://github.com/user-attachments/files/29563541/M1F1-uint8-AFsp.w64.zip)

### DIFF
--- a/fmt/w64.c
+++ b/fmt/w64.c
@@ -42,6 +42,7 @@ struct w64_chunk {
 static int w64_chunk_peek(struct w64_chunk *chunk, slurp_t *fp)
 {
 	int64_t pos;
+	uint64_t chunk_data_size;
 
 	if (slurp_read(fp, chunk->id, sizeof(chunk->id)) != sizeof(chunk->id))
 		return 0;
@@ -62,13 +63,16 @@ static int w64_chunk_peek(struct w64_chunk *chunk, slurp_t *fp)
 		return 0;
 
 	/* w64 sizes are aligned to 64-bit boundaries */
-	slurp_seek(fp, (chunk->size + 7) & ~7, SEEK_CUR);
+	chunk_data_size = (chunk->size + 7) & ~7;
 
-	pos = slurp_tell(fp);
-	if (pos < 0)
+	if (slurp_available(fp, chunk_data_size, SEEK_CUR)) {
+		/* skip chunk data, return successfully peeked chunk */
+		slurp_seek(fp, chunk_data_size, SEEK_CUR);
+		return 1;
+	} else {
+		/* stop reading here, this chunk's data is incomplete */
 		return 0;
-
-	return slurp_available(fp, 1, SEEK_CUR);
+	}
 }
 
 static int w64_chunk_empty(struct w64_chunk *chunk)


mrpapersonic:
It might be enough to just fail if `slurp_seek` fails.

This problem probably also exists within the regular IFF handler `fmt/iff.c`, since the code is extremely similar

logiclrd:
Mm, okay, I see `slurp_seek` uses `slurp_available` to test the offset that's passed in, so `slurp_seek` explicitly does _not_ allow seeking past the end of the file. But it _does_ allow seeking _to_ the end of the file? I see `slurp_available` just _always_ returns 0 if `SEEK_END` is used for `whence`, but otherwise, it's okay for `pos + x` to be equal to the length of the file.

Hmm... so then why is this bug happening?

Maybe the `slurp_t` in question is specifying a custom `available` that doesn't properly implement this edge case?? :thinking: Nope, looks like custom `available` is only used for non-seekable `slurp_t`s.

On my Linux machine, the debugger isn't currently working, not sure why. But, maybe I can trace through this on a Windows machine in Visual Studio, see what exactly is going on...

(Always returning 0 from `slurp_available` when `whence == SEEK_END` seems like a consistency bug to me, as an aside. It should probably return `(x == 0)`, because a seek to `SEEK_END` with `offset == 0` should work, right? Or am I misunderstanding what `slurp_available` should do? I'm interpreting it as testing whether a seek with the same parameters would be valid.)

mrpapersonic:
> Hmm... so then why is this bug happening?

because it was using `slurp_available(fp, 1, SEEK_CUR);` as the return value.

You can tell `slurp_available` was bad API design because even I don't really know what it's supposed to mean, and I *wrote* it. It's not even a good design for an I/O library because whether something is available should be determined as the result of the read operation, not before it.

FWIW, this is what `fmt/iff.c` does:

```c
        return slurp_available(fp, 0, SEEK_CUR); /* I have NO idea if this is right */
```

Note the comment from my past self who is also confused on what `slurp_available` is supposed to do.

jogged my memory: it was intended to be a replacement for `slurp_tell(fp) + (size) <= slurp_length(fp)` and `(size) <= slurp_length(fp)` for `SEEK_CUR` and `SEEK_SET` respectively. which also isn't good. ERRRRR. The reasoning being that calculating the length of a nonseekable stream requires reading the entire thing.

logiclrd:
Hmm... So maybe it should be renamed `slurp_can_seek` or something. I'm pretty sure `slurp_available(.., 0, SEEK_CUR)` should _always_ return 1, unless someone has done a seek past the end of the file out from underneath slurp, no? 😀 Alternately, to be really explicit, it could be `slurp_is_valid_file_pointer` -- same arguments, same semantics (though with `SEEK_END` fixed -- and maybe `x` signed instead of unsigned??)

mrpapersonic:
well, either way, grabbing `pos` is unnecessary now. I just tested a quick fix that just amounts to returning whether `slurp_seek` succeeded and it seems to work just fine for what this is supposed to return (1 if the chunk overflows EOF and 0 if it doesn't)

PR #835, opened 2026-07-02
`slurp_available` refactor

logiclrd:
This PR eliminates uncertainty and ambiguity around what exactly `slurp_available` does and how it is intended to be called.

- `slurp_available` is renamed to `slurp_is_valid_file_pointer`
- A use of `slurp_available` that depended on a non-obvious side-effect was updated to call a function that wraps `slurp_available` to provide explicitly the semantics needed.
- Comments have been added explicitly clarifying that `slurp_seek`'s return value is like that of `fseek` (0 on success, -1 on failure).
- A TODO comment was removed because the thing it referred to appears to exist.

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -165,7 +165,7 @@ int slurp_mmap(slurp_t *useme, const char *filename, uint64_t st);
 #endif
 
 /* stdio-style file processing */
-int slurp_seek(slurp_t *t, int64_t offset, int whence); /* whence => SEEK_SET, SEEK_CUR, SEEK_END */
+int slurp_seek(slurp_t *t, int64_t offset, int whence); /* whence => SEEK_SET, SEEK_CUR, SEEK_END, return => 0 on success, -1 on error */
 int64_t slurp_tell(slurp_t *t);
 #define slurp_rewind(t) slurp_seek((t), 0, SEEK_SET)
 
@@ -221,7 +221,9 @@ int slurp_xz(slurp_t *src);
 int slurp_zstd(slurp_t *src);
 #endif
 
-int slurp_available(slurp_t *fp, size_t x, int whence);
+int slurp_is_valid_file_pointer(slurp_t *fp, ssize_t x, int whence);
+
+void slurp_fill_nonseek_buffer(slurp_t *t, size_t count);
 
 /* ------------------------------------------------------------------------ */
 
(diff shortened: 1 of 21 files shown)


schism/slurp.c
```diff
@@ -804,8 +805,7 @@ int slurp_receive(slurp_t *t, int (*callback)(const void *, size_t, void *), siz
 	}
 }
 
-/* TODO actually test this function within slurp crap */
-int slurp_available(slurp_t *fp, size_t x, int whence)
+int slurp_is_valid_file_pointer(slurp_t *fp, ssize_t x, int whence)
```
mrpapersonic: use `int64_t` instead of `ssize_t` for consistency with other APIs

schism/slurp.c
```diff
@@ -831,6 +831,19 @@ int slurp_available(slurp_t *fp, size_t x, int whence)
 	}
 }
 
+void slurp_fill_nonseek_buffer(slurp_t *t, size_t count)
```
mrpapersonic: make this `static`; nothing else uses it (and really they shouldn't)

include/slurp.h
```diff
@@ -221,7 +221,9 @@ int slurp_xz(slurp_t *src);
 int slurp_zstd(slurp_t *src);
 #endif
 
-int slurp_available(slurp_t *fp, size_t x, int whence);
+int slurp_is_valid_file_pointer(slurp_t *fp, ssize_t x, int whence);
```
mrpapersonic: Maybe `slurp_seek_test` could also be acceptable...? I don't know. `slurp_is_valid_file_pointer` is about as unhelpful as `slurp_available` as a function name, it sounds more like it's testing the equivalent of `feof() || ferror()`.

schism/slurp.c
logiclrd: Fair enough.

include/slurp.h
logiclrd: Well, except with the context of the parameters being passed in, it's a bit more obvious what it's doing. The problem with `slurp_seek_test` is that it isn't clear that it doesn't actually modify the current file pointer. Perhaps `slurp_could_seek`??

mrpapersonic: `slurp_could_seek` is a much better name, thanks ;)

PR #837, opened 2026-07-08
csf_read_sample: testing and fixes

logiclrd:

This code is the result of discovering a bug in `csf_read_sample` during a debug session. The bug itself was relatively small, but fixing it directly would only contribute to the inconsistency and repetition in the implementation of `csf_read_sample`. The code was screaming for a refactor that would centralize the computation of key quantities and range tests.

There is a problem with doing broad changes to `csf_read_sample`, and that is that it is a very large function that does a lot of different things with intricate inconsistencies and whose functioning is core to the entire program. So, if a change is made that introduces a subtle bug, it's going to cause problems, probably in more than one place.

This PR addresses this problem by adding comprehensive automated testing of the formats supported by `csf_read_sample`. It does this using a code generator, which I have written in C# targetting .NET 10, because it provides far more robust string and collection handling than C, is not a part of the main codebase proper, and is fully supported on Windows and Linux (meaning it could trivially be run on a GitHub agent if needed).

The bug that instigated this is as follows: Each of the paths in `csf_read_sample` includes a `slurp_could_seek` probe to make sure that the necessary data is present. However, the 16-bit stereo sample formats do not calculate the total number of bytes correctly.

The 8-bit stereo paths look like this:

```
	len = sample->length * 2;

	if (!slurp_could_seek(fp, len, SEEK_CUR))
		break;
```

So, `len` is set to the total number of samples across both channels.

The 16-bit mono paths look like this:

```
	len = sample->length;

	if (!slurp_could_seek(fp, len * 2, SEEK_CUR))
		break;
```

Here, `len` is set to the number of _frames_, and then the frames are multiplied out into channels inside the `slurp_could_seek` call argument.

The 16-bit stereo paths look like this:

```
	len = sample->length * 2;

	if (!slurp_could_seek(fp, len, SEEK_CUR))
		break;
```

It isn't clear whether the multiplication here is setting `len` to the number of frames, or to the number of bytes in one channel. Either way, the value passed into `slurp_could_seek` is off by a factor of 2.

Every path has its own `slurp_could_seek` call, and there isn't good consistency in how they compute their parameters.

This PR moves the `slurp_could_seek` out of the `switch` block, and adds code to compute the byte counts up front. Instead of `len` and `mem`, the code now computes:

* `num_frames`: The number of unique points in time where there are samples.
* `num_samples`: The total number of samples across all channels.
* `num_bytes_in`: The total number of bytes being read. For compressed formats, this gets overwritten later by the actual number of bytes read, and either way becomes the return value.
* `num_bytes_out`: The total number of bytes being stored in `sample->data`.

It is now really clear when reading the code what the meaning is of the value returned by `csf_read_sample`. Also, the automated testing picked up a bug in this area: The MDL decompression functions return the number of bytes they processed but fail to include the 4-byte length prefix that is also consumed.

It was fun figuring out how to do Impulse Tracker sample compression. 🙂


logiclrd:
It's not immediately evident to me why that one build profile on Ubuntu is encountering that error. Every other platform passes, and also Ubuntu passes on the `test` target. Weird. I'll give it a go on my Linux machine later.

logiclrd:
Ahh I see -- it makes the source tarball and then verifies that it can build from it. But this PR adds the two `.inc` files and they're not included. I'm not super familiar with this, but I see a list of headers being assigned to `noinst_HEADERS` and the new `readsample.c` test case is in the `schismtrackertest_SOURCES` list. I'm not super familiar with this, but let's see what happens if I add the `.inc` files to those lists.

logiclrd:
:rocket:

PR #838, opened 2026-07-09
disko: export pattern to sample volume

logiclrd:
I recently thought the Export Pattern function was completely non-functional, because with my settings and ambient sound levels, I literally couldn't hear the sample that it generated. I assumed that playing the sample back would sound the same as playing the song, but the mixer output is considerably quieter than the input even with all the volumes at maximum.

After some debugging, which uncovered another bug and prompted the creation of #837, I figured out that the key factor is the `mixing_volume` for the song, which is by default clamped to the range 0..128 -- but, experimentally, in order for sample values to pass through as-is, a value of 512 is needed!

This PR updates the `disko_writeout_sample` and `disko_multiwrite_samples` functions to force the `mixing_volume` value to 512 in the shadow copy of the `song_t`. Experimentally, this seems to produce the right sound, and logically, it makes sense, because when that sample is subsequently used, the real `mixing_volume` will be applied at that time. Without this change, `mixing_volume` is applied _twice_.

Without this change:

https://github.com/user-attachments/assets/967470da-92e2-410b-bd6b-23531b0accd0

With it:

https://github.com/user-attachments/assets/2d4e74b5-e711-44f5-b920-3474eb54d3b5

### DIFF
--- a/schism/disko.c
+++ b/schism/disko.c
@@ -662,6 +662,10 @@ int disko_writeout_sample(int smpnum, int pattern, int dobind)
 	dwsong.repeat_count = -1; // FIXME do this right
 	csf_loop_pattern(&dwsong, pattern, 0);
 
+	/* emit samples at a level where if the resulting sample is inserted
+	 * back into the pattern, it will be the same volume */
+	dwsong.mixing_volume = 0x200;
+
 	do {
 		disko_write(&ds, buf, csf_read(&dwsong, buf, sizeof(buf)) * bps);
 		if (ds.length >= (size_t) (MAX_SAMPLE_LENGTH * bps)) {
@@ -738,6 +742,10 @@ int disko_multiwrite_samples(int firstsmp, int pattern)
 		errno = err;
 		return DW_ERROR;
 	}
+
+	/* emit samples at a level where if the resulting sample is inserted
+	 * back into the pattern, it will be the same volume */
+	dwsong.mixing_volume = 0x200;
 
 	for (n = 0; n < MAX_CHANNELS; n++) {
 		dwsong.multi_write[n].data = &ds[n];


logiclrd:
I do note that the sample appears _extremely_ saturated in the waveform visualizer, but it sounds exactly like it's supposed to. 😛

mrpapersonic:
I think this has the possibility of breakage if someone were to open a file that has pattern -> sample sync enabled.

logiclrd:
So an existing file would be expecting the sample with the 0x30 volume and suddenly get it with 0x200? Mm, because `_pattern_update_magic` leverages `disko_writeout_sample`. But, this also implies that if someone is using this feature and then they press F12 and edit the Mixing Volume, then the next time the pattern gets updated, the volume of the associated sample will abruptly change, right? I'll see if I can verify this one way or the other.

logiclrd:
Okay, confirmed on both counts: Yes, this change interferes with pattern binding, and the fix looks to be pretty straightforward: don't pin the mixing volume at 0x200 if `dobind`. But also, the user making changes to Mixing Volume _does_ result in unexpected sudden changes to the resulting samples. Also, just from some brief experimentation, the change doesn't always immediately trigger the pattern being written out. If you play back the song, then you edit the Mixing Volume, then you play it again, everything is uniformly louder or quieter, but then if you go to the Pattern Editor and switch to a pattern using a bound sample, at that point it will get rewritten and suddenly the sample will have the wrong volume -- basically baking in Mixing Volume at two different levels, so if Mixing Volume got louder, the sample gets louder twice, and if it got quieter, the sample gets quieter twice.

I think the technically correct thing to do would be:

```
	if (dobind) {
		/* use the default mixing volume for consistent bound samples */
		dwsong.mixing_volume = 0x30;
	} else {
		/* emit samples at a level where if the resulting sample is inserted
		 * back into the pattern, it will be the same volume */
		dwsong.mixing_volume = 0x200;
	}
```

There's an intractable problem, though: existing songs that have had their mixing volume updated and _then_ been composed with patterns bound to samples. They're going to expect those patterns to keep getting written out at their non-default mixing volume. :-/

logiclrd:
It is interesting that if you set Mixing Volume to the highest value the UI supports, 0x80, and then you export a sample, the waveform in the exported sample is only 25% the magnitude of the original samples. Is that a deliberate thing or did it just fall out of the way the mixing pipeline was designed??

logiclrd:
It's probably more trouble than it's worth, but one way to fix it would be to introduce a separate field for the mixing volume to use for bound patterns, tied of course to a new version of the file format. When an old file is loaded, that field is populated initially with the current output mixing volume, and can then diverge going forward. Either way, now the UI has two different sliders, and editing the output mixing volume doesn't affect bound patterns. If the user elects to save a file using the old version, then if they have bound patterns, they get a compatibility warning.

PR #839, opened 2026-07-09
fmt/iff: fix iff_chunk_peek_ex return value and add regression test

logiclrd:

The latest builds can no longer load WAV files. The reason appears to be that a recent change to `iff_chunk_peek_ex` was switched from calling `slurp_available` (now `slurp_could_seek`) to using the return value of an existing `slurp_seek` call -- but it gets the polarity of the return value wrong. `slurp_seek` returns 0 on success, like `fseek`, but `iff_chunk_peek_ex` is looking for non-zero.

This PR adds regression tests that initially fail because of this bug, and then corrects the bug, making the tests pass.

### DIFF
--- a/test/cases/iff.c
+++ b/test/cases/iff.c
@@ -1,0 +1,128 @@
+/*
+ * Schism Tracker - a cross-platform Impulse Tracker clone
+ * copyright (c) 2003-2005 Storlek <storlek@rigelseven.com>
+ * copyright (c) 2005-2008 Mrs. Brisby <mrs.brisby@nimh.org>
+ * copyright (c) 2009 Storlek & Mrs. Brisby
+ * copyright (c) 2010-2012 Storlek
+ * URL: http://schismtracker.org/
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include "test.h"
+#include "test-assertions.h"
+
+#include "slurp.h"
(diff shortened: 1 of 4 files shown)


test/cases/iff.c
```diff
@@ -0,0 +1,128 @@
...
+		0x70, 0x6F, 0x02, 0x00, // average byte rate (samples x 2 bytes per sample x 2 channels = 0x26F70 == 159,600)
+		0x04, 0x00, // frame alignment (4 bytes)
+		0x10, 0x00, // bits per sample (16)
+		'd', 'a', 't', 'a', // Lead in to the following chunk
+		0x80, 0x61, 0x01, 0x00, // etc.
+	};
+
+	int fmt_start = 4;
+	int fmt_content_start = fmt_start + 8;
+	int fmt_length = bswapLE32(((int *)&input_data[fmt_start])[1]);
```
mrpapersonic: use `uint32_t` and `memcpy` instead -- casting a `char *` to an `int *` and dereferencing it is a strict aliasing violation. It's always a good idea to use exact-width integer types in code dealing with raw data. (its really a good idea to use them regardless as it makes things the same size everywhere)

test/cases/iff.c
```diff
@@ -0,0 +1,128 @@
...
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include "test.h"
+#include "test-assertions.h"
+
+#include "slurp.h"
+#include "fmt.h"
+
+#include <stdio.h>
```
mrpapersonic:
```suggestion
#include "headers.h"
```

test/cases/iff.c
```diff
@@ -0,0 +1,128 @@
...
+
+#define IFFID_fmt  UINT32_C(0x666D7420)
+
+testresult_t test_iff_chunk_peek_ex_middle(void)
+{
+	// Arrange
+
+	char input_data[] = {
```
mrpapersonic:
```suggestion
	static const char input_data[] = {
```
additionally it would be nice to just make this data global and put a lower size into the EOF function's `slurp_read` call

mrpapersonic (changes requested):
thx for fixing this, I just have a few nitpicks

test/cases/iff.c
logiclrd: Mm, my baseline is always to keep tests strictly independent, unless it results in outright absurdity. A bit of repetition in the main codebase is a smell, in test code, it's a bit of reuse that's a smell. If, in the future, there's ever a need to alter the input data to one of the tests, now the other one also needs to be factored in and possibly decoupled. I can make that change if you really prefer it but it is my opinion that it makes the code worse, not better.

PR #846, opened 2026-08-16
slurp: add a simple built-in buffer

mrpapersonic:

Adds a buffer to slurp; it defaults to 4KiB but can be expanded or shrunk according to the `st_blksize` returned by `stat`.

I tried to make it as simple as I could, similar to the already existing `slurp_decompress`.

It can be improved by not immediately throwing out the already existing buffer on `seek()` if the seeked position is within the range of the buffer. For the first few bytes of the file this is pretty important as we will be re-reading the magic bytes a lot.


mrpapersonic:
rebased and merged

